PR #2, opened 2026-01-05
added conversation, chat_message and tool_calls tables

MatLBS:
- added 3 tables : conversation, chat_message and tool_calls
- a conversation is linked to a user ; a chat_message is linked to a conversation and a tool_call is linked to a chat_message


apps/backend/migrations/0000_smart_gideon.sql
Standlc, line 1: Should not be deleted, revert the changes here. What happened?

apps/backend/src/db/schema.ts
```diff
@@ -88,9 +94,63 @@ export const verification = sqliteTable(
 	(table) => [index('verification_identifier_idx').on(table.identifier)],
 );
 
+export const conversation = sqliteTable(
+	'conversation',
+	{
+		id: text('id').primaryKey(),
```
Standlc, line 100: Is it not possible to have drizzle handle the generation of the id?

apps/backend/migrations/meta/0001_snapshot.json
Standlc, line 1: add `migrations/meta` to the gitignore, then delete all this folder from the repo (remotely)

apps/backend/src/db/schema.ts
MatLBS: It is with "id: int().primaryKey({ autoIncrement: true })," but I though it would be better to use the a proper id generated by a function like generateId() from AI SDK

PR #3, opened 2026-01-05
Add a Python CLI + bundle the node server + frontend to be server by …

Bl3f:
* Added a Python CLI with 2 commands: init and chat
* Bundled the Chat application to be server by the chat command
* init command get a project name and ask for your OpenAI API KEY (then forward it)

### DIFF
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -1,6 +1,5 @@
 {
 	"name": "frontend",
-	"version": "0.0.1",
 	"private": true,
 	"type": "module",
 	"scripts": {
--- a/cli/build.py
+++ b/cli/build.py
@@ -8,7 +8,6 @@ This script:
 4. Bundles everything into a Python wheel
 """
 
-import os
 import re
 import shutil
 import subprocess
@@ -34,15 +33,6 @@ def run(cmd: list[str], cwd: Path | None = None, env: dict | None = None) -> Non
     if result.returncode != 0:
         print(f"❌ Command failed: {' '.join(cmd)}")
         sys.exit(1)
-
-
-def get_env_with_bun() -> dict:
-    """Get environment with bun in PATH."""
-    env = os.environ.copy()
-    bun_path = Path.home() / ".bun" / "bin"
-    if bun_path.exists():
-        env["PATH"] = f"{bun_path}:{env.get('PATH', '')}"
-    return env
 
 
 def parse_version(version: str) -> tuple[int, int, int]:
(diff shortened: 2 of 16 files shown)


apps/frontend/package.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "frontend",
+	"version": "0.0.1",
```
Standlc: why?

apps/frontend/tsconfig.json
```diff
@@ -1,19 +1,28 @@
 {
-	"include": ["**/*.ts", "**/*.tsx", "eslint.config.js", "vite.config.js"],
-
+	"include": [
```
Standlc, line 2: I think your IDE formatter is not setup correctly (make sure to have prettier used as the formatter), unless you did that manually

cli/nao_core/commands/sync.py
```diff
@@ -0,0 +1,150 @@
...
+
+from rich.console import Console
+from rich.progress import BarColumn, Progress, SpinnerColumn, TaskProgressColumn, TextColumn
+
+from nao_core.config import NaoConfig
+
+console = Console()
+
+
+def get_table_schema_markdown(conn, dataset: str, table: str) -> str:
```
Standlc: I have no idea what is a `conn` and how to use it :(

cli/build.py
```diff
@@ -0,0 +1,237 @@
...
+
+def run(cmd: list[str], cwd: Path | None = None, env: dict | None = None) -> None:
+    """Run a command and exit on failure."""
+    result = subprocess.run(cmd, cwd=cwd, env=env)
+    if result.returncode != 0:
+        print(f"❌ Command failed: {' '.join(cmd)}")
+        sys.exit(1)
+
+
+def get_env_with_bun() -> dict:
```
Standlc: why?

cli/build.py
```diff
@@ -0,0 +1,237 @@
...
+    backend_public = backend_dir / "public"
+    if backend_public.exists():
+        shutil.rmtree(backend_public)
+    shutil.copytree(frontend_dir / "dist", backend_public)
+
+    # Step 3: Compile backend with Bun
+    print("\n⚡ Compiling backend with Bun...")
+    env = get_env_with_bun()
+    run(
+        [
```
Standlc: just do `npm run build:bun`?

cli/pyproject.toml
```diff
@@ -0,0 +1,60 @@
...
+    "cyclopts>=4.4.4",
+    "pydantic>=2.10.0",
+    "pyyaml>=6.0.0",
+    "rich>=14.0.0",
+    "ibis-framework[bigquery]>=9.0.0",
+    "openai>=1.0.0",
+]
+
+[project.scripts]
+nao = "nao_core.main:app"
```
Standlc, line 32: is this how you bind `nao` to the `main.py` script? Why `:app` suffix?

cli/build.py
Standlc, line 1: I think this build file has nothing to do in the CLI, it builds both python and ts. Why not move it at the root in `scripts/`?

apps/frontend/package.json
Bl3f: ah it's a leftover, at the beggining i tried to instead publish npm package, so it was required to add a version, but drop this idea after and forgot to remove this

cli/nao_core/commands/sync.py
Bl3f: will type (and also I've added ty + ruff which enforce typing mainly when hint given, i think we can also find a rule to make it mandatory in these case for instance)

cli/Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: lint lint-fix format
+
+lint:
```
MatLBS, line 3: I don't know exactly what is the purpose but running 'uv run ty check' through 'make lint' raised some errors like 

```
warning[possibly-missing-attribute]: Attribute `project_name` may be missing on object of type `NaoConfig | None`
  --> nao_core/commands/debug.py:71:70
   |
69 |         sys.exit(1)
70 |
71 |     console.print(f"[bold green]✓[/bold green] Loaded config: [cyan]{config.project_name}[/cyan]\n")
   |                                                                      ^^^^^^^^^^^^^^^^^^^
72 |
73 |     # Test databases
   |
info: rule `possibly-missing-attribute` is enabled by default
```

cli/nao_core/commands/debug.py
```diff
@@ -0,0 +1,144 @@
...
+            tables = conn.list_tables()
+            table_count = len(tables)
+            return True, f"Connected successfully ({table_count} tables found)"
+        else:
+            # If no dataset, list datasets in the project instead
+            datasets = conn.list_databases()
+            dataset_count = len(datasets)
+            return True, f"Connected successfully ({dataset_count} datasets found)"
+    except Exception as e:
+        return False, str(e)
```
MatLBS, line 32: Should you close the connection with 'conn.close()' ?

cli/Makefile
Bl3f: yeah there are a few linting warning i will work on it to remove them

the purpose of the Makefile is to get all commands in one place, i usually don't like Makefile but it's a first good way to get 4 commands in one

cli/nao_core/commands/debug.py
Bl3f: not mandatory as it's a cli command that ends, but will add

cli/build.py
Bl3f: i keeping it there because it needs packages that are in the uv env in cli folder to install

PR #4, opened 2026-01-06
Feature/implementing switch db

MatLBS:
- Implemented switch between pg and sqlite 
- Switch is based on the env variable 'DB_SELECTED'

### DIFF
--- a/apps/backend/src/db/sqlite-schema.ts
+++ b/apps/backend/src/db/sqlite-schema.ts
@@ -7,8 +7,8 @@ export type Conversation = typeof conversation.$inferSelect;
 export type NewConversation = typeof conversation.$inferInsert;
 export type ChatMessage = typeof chat_message.$inferSelect;
 export type NewChatMessage = typeof chat_message.$inferInsert;
-export type ToolCall = typeof tool_calls.$inferSelect;
-export type NewToolCall = typeof tool_calls.$inferInsert;
+export type ToolCall = typeof tool_call.$inferSelect;
+export type NewToolCall = typeof tool_call.$inferInsert;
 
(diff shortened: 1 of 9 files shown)


apps/backend/src/utils.ts
Standlc, line 1: I think you need to import `dotenv` like in other files to load the env vars.

apps/backend/src/db/sqlite-schema.ts
```diff
@@ -88,9 +94,63 @@ export const verification = sqliteTable(
...
+		role: text('role', { enum: ['user', 'assistant', 'system'] }).notNull(),
+		content: text('content').notNull(),
+		createdAt: integer('created_at', { mode: 'timestamp_ms' })
+			.default(sql`(cast(unixepoch('subsecond') * 1000 as integer))`)
+			.notNull(),
+	},
+	(table) => [index('chat_message_conversationId_idx').on(table.conversationId)],
+);
+
+export const tool_calls = sqliteTable(
```
Standlc: `tool_call` without 's'

apps/backend/tests/pg.test.ts
```diff
@@ -0,0 +1,55 @@
...
+		name: 'John',
+		email: 'john@example.com',
+	};
+
+	afterEach(async () => {
+		await db.delete(user).where(eq(user.email, testUser.email));
+	});
+
+	it('should insert a new user', async () => {
+		await db.insert(user).values(testUser);
```
Standlc, line 27: we're gonna have problems when we create rows if we have to use the table objects specific to each db (here `user`).

PR #6, opened 2026-01-09
Feature/implementing agent tools

MatLBS:
Added some tools : 
- read_file
- search_files
- grep_codebase
- list_directory
- execute_sql

Based on current tools in vsnao repo

Created a simple fastAPI server to run the tool execute_sql

### DIFF
--- a/apps/backend/src/agents/tools.ts
+++ b/apps/backend/src/agents/tools.ts
@@ -1,8 +1,11 @@
 import { tool } from 'ai';
-import fs from 'fs/promises';
-import { glob } from 'glob';
-import path from 'path';
 import z from 'zod/v3';
+
+import { execute_sql } from './tools/execute_sql';
+import { grep_codebase } from './tools/grep_codebase';
+import { list_directory } from './tools/list_directory';
+import { read_file } from './tools/read_file';
+import { search_files } from './tools/search_files';
 
 export const tools = {
 	getWeather: tool({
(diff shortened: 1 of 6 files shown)


apps/backend/src/agents/tools.ts
```diff
@@ -23,4 +26,208 @@ export const tools = {
...
+			const response = await fetch(`${process.env.FASTAPI_URL}/execute_sql`, {
+				method: 'POST',
+				headers: {
+					'Content-Type': 'application/json',
+				},
+				// TO DO : replace with only the query and get the project_id and credentials_path from cli config
+				body: JSON.stringify({
+					sql: query,
+					project_id: 'nao-corp',
+					credentials_path: '/Users/mateolebrassancho/Downloads/nao-corp-1693265c8499.json',
```
Standlc: oops

apps/backend/src/agents/tools.ts
```diff
@@ -23,4 +26,208 @@ export const tools = {
...
+		description: 'Read the contents of a file at a given path.',
+		inputSchema: z.object({
+			file_path: z.string(),
+		}),
+		outputSchema: z.object({
+			content: z.string(),
+			numberOfTotalLines: z.number(),
+		}),
+		execute: async ({ file_path }) => {
+			try {
```
Standlc: i'd remove all the try & catch, they add very little value. If an error is thrown, the agent will already know what tool execution caused it.

apps/backend/src/agents/tools.ts
```diff
@@ -23,4 +26,208 @@ export const tools = {
...
+							size,
+						};
+					}),
+				);
+			} catch (error) {
+				throw new Error(`Error listing directory at ${dirPath}: ${error}`);
+			}
+		},
+	}),
+	grep_codebase: tool({
```
Standlc, line 79: Not viable in large codebases and lots of agents using `grep_codebase`, although it's ok for this PR.
We should rather use `ripgrep` like in vscode (check how they spawn `ripgrep` workers for that).

apps/backend/src/agents/tools.ts
Standlc, line 1: Can create one file per tool? This file is bound to grow and is not going to be easy to navigate when we have 20+ tools.

Standlc (changes requested):

PR #12, opened 2026-01-15
Feat/slack integration

MatLBS:
https://github.com/user-attachments/assets/a92efebc-a339-4631-8dc7-35beb6397a67

Implemented slack integration : 

- The user has to be log in on nao-chat with the same email as Slack
- Conversation on Slack happens in a thread
- Each messages in the thread is stored in the same 'chat' table
- Created a middleware for slack

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,8 @@
 		},
 		"apps/backend/node_modules/@types/node": {
 			"version": "25.0.3",
+			"resolved": "https://registry.npmjs.org/@types/node/-/node-25.0.3.tgz",
+			"integrity": "sha512-W609buLVRVmeW693xKfzHeIV6nJGGz98uCPfeXI1ELMLXVeKYZ9m15fAMSaUPBHYLGFsVRcMmSCksQOrZV9BYA==",
 			"dev": true,
 			"license": "MIT",
 			"peer": true,
@@ -66,6 +68,8 @@
 		},
 		"apps/backend/node_modules/@vitest/expect": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/expect/-/expect-4.0.16.tgz",
+			"integrity": "sha512-eshqULT2It7McaJkQGLkPjPjNph+uevROGuIMJdG3V+0BSR2w9u6J9Lwu+E8cK5TETlfou8GRijhafIMhXsimA==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -82,6 +86,8 @@
 		},
 		"apps/backend/node_modules/@vitest/mocker": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/mocker/-/mocker-4.0.16.tgz",
+			"integrity": "sha512-yb6k4AZxJTB+q9ycAvsoxGn+j/po0UaPgajllBgt1PzoMAAmJGYFdDk0uCcRcxb3BrME34I6u8gHZTQlkqSZpg==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -107,6 +113,8 @@
 		},
 		"apps/backend/node_modules/@vitest/pretty-format": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/pretty-format/-/pretty-format-4.0.16.tgz",
+			"integrity": "sha512-eNCYNsSty9xJKi/UdVD8Ou16alu7AYiS2fCPRs0b1OdhJiV89buAXQLpTbe+X8V9L6qrs9CqyvU7OaAopJYPsA==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -118,6 +126,8 @@
 		},
 		"apps/backend/node_modules/@vitest/runner": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/runner/-/runner-4.0.16.tgz",
+			"integrity": "sha512-VWEDm5Wv9xEo80ctjORcTQRJ539EGPB3Pb9ApvVRAY1U/WkHXmmYISqU5E79uCwcW7xYUV38gwZD+RV755fu3Q==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -130,6 +140,8 @@
 		},
 		"apps/backend/node_modules/@vitest/snapshot": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/snapshot/-/snapshot-4.0.16.tgz",
+			"integrity": "sha512-sf6NcrYhYBsSYefxnry+DR8n3UV4xWZwWxYbCJUt2YdvtqzSPR7VfGrY0zsv090DAbjFZsi7ZaMi1KnSRyK1XA==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -143,6 +155,8 @@
 		},
 		"apps/backend/node_modules/@vitest/spy": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/spy/-/spy-4.0.16.tgz",
+			"integrity": "sha512-4jIOWjKP0ZUaEmJm00E0cOBLU+5WE0BpeNr3XN6TEF05ltro6NJqHWxXD0kA8/Zc8Nh23AT8WQxwNG+WeROupw==",
 			"dev": true,
 			"license": "MIT",
 			"funding": {
@@ -151,6 +165,8 @@
 		},
 		"apps/backend/node_modules/@vitest/utils": {
 			"version": "4.0.16",
+			"resolved": "https://registry.npmjs.org/@vitest/utils/-/utils-4.0.16.tgz",
+			"integrity": "sha512-h8z9yYhV3e1LEfaQ3zdypIrnAg/9hguReGZoS7Gl0aBG5xgA410zBqECqmaF/+RkTggRsfnzc1XaAHA6bmUufA==",
 			"dev": true,
 			"license": "MIT",
 			"dependencies": {
@@ -163,6 +179,8 @@
 		},
 		"apps/backend/node_modules/chai": {
 			"version": "6.2.2",
+			"resolved": "https://registry.npmjs.org/chai/-/chai-6.2.2.tgz",
+			"integrity": "sha512-NUPRluOfOiTKBKvWPtSD4PhFvWCqOi0BGStNWs57X9js7XGTprSmFoz5F0tWhR4WPjNeR9jXqdC7/UpSJTnlRg==",
 			"dev": true,
 			"license": "MIT",
 			"engines": {
@@ -171,6 +189,8 @@
 		},
 		"apps/backend/node_modules/tinyexec": {
 			"version": "1.0.2",
+			"resolved": "https://registry.npmjs.org/tinyexec/-/tinyexec-1.0.2.tgz",
+			"integrity": "sha512-W/KYk+NFhkmsYpuHq5JykngiOCnxeVL8v8dFnqxSD8qEEdRfXk1SDM6JzNqcERbcGYj9tMrDQBYV9cjgnunFIg==",
 			"dev": true,
 			"license": "MIT",
 			"engines": {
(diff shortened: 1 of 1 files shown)


apps/backend/fastapi/main.py
```diff
@@ -98,5 +99,11 @@ async def execute_sql(request: ExecuteSQLRequest):
         raise HTTPException(status_code=500, detail=str(e))
 
 
+@app.post("/")
```
Bl3f: do we need this right now?

.env.example
```diff
@@ -14,4 +14,9 @@ BETTER_AUTH_SECRET= #run `openssl rand -base64 32` to generate
...
 
 # LLM providers
-OPENAI_API_KEY=sk-123456789
\ No newline at end of file
+OPENAI_API_KEY=sk-123456789
+
+# Slack App
+SLACK_BOT_TOKEN=
+SLACK_POST_MESSAGE_URL=https://slack.com/api/chat.postMessage
+SLACK_SIGNING_SECRET=
```
Bl3f: missing newline

cli/uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
Bl3f: not sure it should change

apps/backend/src/app.ts
```diff
@@ -22,6 +24,13 @@ export type App = typeof app;
 app.setValidatorCompiler(validatorCompiler);
 app.setSerializerCompiler(serializerCompiler);
 
+// Register raw body plugin for Slack signature verification
+app.register(fastifyRawBody, {
```
Bl3f, line 28: is there any side effect of this?

apps/backend/src/routes/slack.ts
```diff
@@ -0,0 +1,36 @@
+import type { App } from '../app';
+import { slackAuthMiddleware } from '../middleware/auth';
+import { getSlackUser, handleSlackWorkFlow, sendSlackRequestAcknowledgement } from '../utils/slack';
+
+export const slackRoutes = async (app: App) => {
+	app.addHook('preHandler', slackAuthMiddleware);
+
+	app.post('/app_mention', { config: { rawBody: true } }, async (request, reply) => {
```
Bl3f, line 11: is it required by slack to be name `/app_mention`?

apps/backend/src/routes/slack.ts
```diff
@@ -0,0 +1,36 @@
...
+		try {
+			/* eslint-disable @typescript-eslint/no-explicit-any */
+			const body = request.body as any;
```
Bl3f: Can't we type this? At least for what we need

apps/backend/src/middleware/auth.ts
```diff
@@ -21,3 +22,27 @@ export async function authMiddleware(request: FastifyRequest, reply: FastifyRepl
...
+	return crypto.timingSafeEqual(Buffer.from(mySignature), Buffer.from(requestSignature));
+}
+
+export async function slackAuthMiddleware(request: FastifyRequest, reply: FastifyReply) {
+	const rawBody = request.rawBody as string;
+	const timestamp = request.headers['x-slack-request-timestamp'] as string;
+	const signature = request.headers['x-slack-signature'] as string;
+
+	if (!verifySlackSignature(process.env.SLACK_SIGNING_SECRET!, signature, timestamp, rawBody)) {
+		console.log('❌ Invalid Slack signature');
```
Bl3f: remove this

apps/backend/src/middleware/auth.ts
```diff
@@ -21,3 +22,27 @@ export async function authMiddleware(request: FastifyRequest, reply: FastifyRepl
 	request.user = session.user;
 	request.session = session.session;
 }
+
+function verifySlackSignature(signingSecret: string, requestSignature: string, timestamp: string, rawBody: string) {
+	const currentTime = Math.floor(Date.now() / 1000);
+	if (Math.abs(currentTime - parseInt(timestamp)) > 300) {
+		console.log('❌ Timestamp is too old');
```
Bl3f: remove this

apps/backend/src/middleware/auth.ts
```diff
@@ -21,3 +22,27 @@ export async function authMiddleware(request: FastifyRequest, reply: FastifyRepl
 	request.user = session.user;
 	request.session = session.session;
 }
+
+function verifySlackSignature(signingSecret: string, requestSignature: string, timestamp: string, rawBody: string) {
```
Bl3f, line 26: can you point to a documentation (in a comment) that explain why we should do this? and what it means at least

apps/backend/src/db/sqliteSchema.ts
```diff
@@ -97,6 +97,7 @@ export const chat = sqliteTable(
 			.notNull()
 			.references(() => user.id, { onDelete: 'cascade' }),
 		title: text('title').notNull().default('New Conversation'),
+		slackThreadTs: text('slack_thread_ts'),
```
Bl3f: what Ts means? I feel this is poorly named and I'm not sure it should be in the chat table, maybe we should have a slackChat table (summon @Standlc to get a second thought)

apps/backend/src/utils/slack.ts
```diff
@@ -0,0 +1,101 @@
+import { WebClient } from '@slack/web-api';
```
Bl3f: should be a service I think and not an util as it connects to a specific service (maybe not everything goes into the service tho)

apps/backend/src/utils/slack.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import { generateResponse } from '../agents/agentService';
+import { User } from '../db/abstractSchema';
+import * as chatQueries from '../queries/chatQueries';
+import { getUser } from '../queries/userQueries';
+import { UIMessage } from '../types/chat';
+
+const slackClient = new WebClient(process.env.SLACK_BOT_TOKEN);
+const redirectUrl = 'http://localhost:3000/';
```
Bl3f: should also be an env variable

after we will put this first in the nao_config.yaml and then in the configuration page in the app

apps/backend/src/utils/slack.ts
```diff
@@ -0,0 +1,101 @@
...
+	const message: UIMessage = {
+		id: crypto.randomUUID(),
+		role,
+		parts: [{ type: 'text', text }],
+	};
+	return message;
+};
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const getSlackUser = async (body: any, channel: string, threadTs: string, reply: any): Promise<User> => {
```
Bl3f: some any could be type i feel

apps/backend/src/utils/slack.ts
```diff
@@ -0,0 +1,101 @@
...
+const getSlackUserEmail = async (userId: string): Promise<string | null> => {
+	const userProfile = await slackClient.users.profile.get({ user: userId });
+	return userProfile.profile?.email || null;
+};
+
+const sendSlackRequestAcknowledgement = async (channel: string, threadTs: string, reply: any) => {
+	reply.send({ ok: true });
+	await slackClient.chat.postMessage({
+		channel: channel,
+		text: '🔄 Nao is analyzing your request... This may take a moment.',
```
Bl3f:
```suggestion
		text: '🔄 nao is answering... please wait a few seconds.',
```

apps/backend/src/utils/slack.ts
```diff
@@ -0,0 +1,101 @@
...
+};
+
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const getSlackUser = async (body: any, channel: string, threadTs: string, reply: any): Promise<User> => {
+	reply.send({ ok: true });
+	const userEmail = await getSlackUserEmail(body.event?.user);
+
+	const user = await getUser({ email: userEmail! });
+	if (!user) {
+		const fullMessage = `❌ On nao-chat, create a user account with ${userEmail} to use this bot. \n\n Go to ${redirectUrl} to sign up.`;
```
Bl3f:
```suggestion
		const fullMessage = `❌ No user found. Create an user account with ${userEmail} on ${redirectUrl} to sign up.`;
```

apps/backend/src/agents/agentService.ts
```diff
@@ -0,0 +1,43 @@
+import { openai } from '@ai-sdk/openai';
+import { ModelMessage, ToolLoopAgent } from 'ai';
+
+import { tools } from './tools';
+
+const instructions = `
```
Bl3f: Really similar to prompts.ts file, should find a way to avoid repeating the instructions that are the same

apps/backend/src/agents/agentService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+function createAgent() {
+	return new ToolLoopAgent({
+		model: openai.chat('gpt-5.1'),
+		instructions,
+		tools,
+	});
+}
+
+export function streamResponse(messages: ModelMessage[]) {
```
Bl3f: do you need this?

Bl3f (commented):
A few comments

apps/backend/fastapi/main.py
MatLBS: Not really, I was just for my tests

apps/backend/src/agents/agentService.ts
MatLBS: Sure ! As you just added the file, I am going to use it

apps/backend/src/agents/agentService.ts
MatLBS: We will have to use it in chat.ts

apps/backend/src/db/sqliteSchema.ts
MatLBS: Ts stands for 'Time Stamp'. Slack sends a 'Ts' and 'threadTs' in its HTTP request. 

I had the idea to create a slackChat table but then I thought it would be cleaner to only have one 'chat' table

apps/backend/src/routes/slack.ts
MatLBS: No it's not. Do you have an idea on how to call this route ?

apps/backend/src/middleware/auth.ts
MatLBS: Verifying requests from Slack : verify whether requests from Slack are authentic

https://docs.slack.dev/authentication/verifying-requests-from-slack/#signing_secrets_admin_page

apps/backend/src/app.ts
MatLBS: I don't think so. 

I had to put `{ config: { rawBody: true }` in my route declaration so it won't impact the routes where it is not configured

apps/backend/src/routes/slack.ts
Bl3f: from what im seeing it's the Slack event that is named app_mention maybe it should be /api/slack/events/app_mention to be a bit more explicit (so here you add the events

apps/backend/src/middleware/auth.ts
Bl3f: put this as a comment in the code i'd suggest

apps/backend/src/db/sqliteSchema.ts
Bl3f: but what, why do we identify the thread with a timestamp? can't we use a generated id from Slack?

i'd be in favour of at least putting id in the name in the table as this is some kind of identifier, and even if we keep timestamp i'd put slackThreadTimestamp to be explicite

but in favour of having something called id, it's weird without

apps/backend/src/agents/agentService.ts
Bl3f: ah cool sorry thought it was already here when u did it

apps/backend/src/agents/agentService.ts
Bl3f: ah it's mb, it's not specific to Slack

apps/backend/src/db/sqliteSchema.ts
MatLBS: The time stamp is the id as it stays the same throughout the thread. But you are right i renamed it in my schemas

apps/backend/migrations-postgres/meta/_journal.json
```diff
@@ -8,6 +8,13 @@
       "when": 1768388561542,
       "tag": "0000_user_auth_and_chat_tables",
       "breakpoints": true
+    },
+    {
+      "idx": 1,
+      "version": "7",
+      "when": 1768493932001,
```
Bl3f: migration to bump

PR #13, opened 2026-01-15
default input when no messages

Bl3f:
Screenshot 2026-01-15 at 16 16 22

### DIFF
--- a/apps/frontend/src/routes/_chat-layout.index.tsx
+++ b/apps/frontend/src/routes/_chat-layout.index.tsx
@@ -1,10 +1,15 @@
 import { createFileRoute } from '@tanstack/react-router';
+import { useSession } from '@/lib/auth-client';
+import { capitalize } from '@/lib/utils';
 
 export const Route = createFileRoute('/_chat-layout/')({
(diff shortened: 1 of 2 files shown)


apps/frontend/src/routes/_chat-layout.index.tsx
```diff
@@ -1,10 +1,25 @@
...
+				<img
+					src='/nao-logo-greyscale.svg'
+					alt=''
+					className='w-[200px] h-auto select-none opacity-[0.05]'
+					aria-hidden='true'
+				/>
+			</div>
+
+			<h1 className='text-3xl md:text-4xl font-light tracking-tight text-slate-700 text-center px-6'>
+				Hey, what do you want to analyse?
```
MatLBS: Would be nice to add the username.  ( Hey ${username}, what do you want to analyse? )

You can use 'useSession' from auth-client.ts for that. 

```
const { data: session } = useSession();
const username = session?.user?.name;
```

MatLBS (approved):
Small improvement, otherwise LGTM

PR #14, opened 2026-01-15
Impement user feedback on chat messages

Standlc:
- show message actions on hover (last message is always visible)
- send feedback with optional message when negative
- fix some auth issues when accessing messages and chats

Screenshot 2026-01-15 at 5 56 19 PM
Screenshot 2026-01-15 at 5 56 28 PM

### DIFF
--- a/apps/backend/migrations-postgres/meta/_journal.json
+++ b/apps/backend/migrations-postgres/meta/_journal.json
@@ -8,6 +8,13 @@
       "when": 1768388561542,
       "tag": "0000_user_auth_and_chat_tables",
       "breakpoints": true
+    },
+    {
+      "idx": 1,
+      "version": "7",
+      "when": 1768496717318,
+      "tag": "0001_message_feedback",
+      "breakpoints": true
     }
   ]
 }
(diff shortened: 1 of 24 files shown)


apps/backend/migrations-postgres/meta/_journal.json
```diff
@@ -8,6 +8,13 @@
       "when": 1768388561542,
       "tag": "0000_user_auth_and_chat_tables",
       "breakpoints": true
+    },
+    {
+      "idx": 1,
+      "version": "7",
+      "when": 1768496717318,
+      "tag": "0001_message_feedback",
```
Bl3f, line 16: you gonna fight with Mateo about who gets the 0001 migration

apps/backend/src/trpc/feedback.routes.ts
```diff
@@ -0,0 +1,41 @@
...
+		.mutation(async ({ input, ctx }): Promise<MessageFeedback> => {
+			const ownerId = await chatQueries.getOwnerOfChatAndMessage(input.chatId, input.messageId);
+			if (!ownerId) {
+				throw new TRPCError({
+					code: 'NOT_FOUND',
+					message: `Message with id ${input.messageId} not found.`,
+				});
+			}
+
+			if (ownerId !== ctx.user.id) {
```
Bl3f, line 28: should this be in a middleware like for the auth?

PR #20, opened 2026-01-16
Fix/add input verif slack

MatLBS:
- check for error 400
- added zod schema validation
- moved the middleware to a proper file

### DIFF
--- a/apps/backend/src/middleware/slack.middleware.ts
+++ b/apps/backend/src/middleware/slack.middleware.ts
@@ -14,9 +14,17 @@ function verifySlackSignature(signingSecret: string, requestSignature: string, t
 }
 
 export async function slackAuthMiddleware(request: FastifyRequest, reply: FastifyReply) {
-	const rawBody = request.rawBody as string;
-	const timestamp = request.headers['x-slack-request-timestamp'] as string;
-	const signature = request.headers['x-slack-signature'] as string;
+	const rawBody = request.rawBody;
+	const timestamp = request.headers['x-slack-request-timestamp'];
+	const signature = request.headers['x-slack-signature'];
+
+	if (!rawBody || !timestamp || !signature) {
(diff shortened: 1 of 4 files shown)


apps/backend/src/middleware/slack.middleware.ts
```diff
@@ -0,0 +1,24 @@
...
+	const sigBasestring = `v0:${timestamp}:${rawBody}`;
+	const mySignature = 'v0=' + crypto.createHmac('sha256', signingSecret).update(sigBasestring).digest('hex');
+
+	return crypto.timingSafeEqual(Buffer.from(mySignature), Buffer.from(requestSignature));
+}
+
+export async function slackAuthMiddleware(request: FastifyRequest, reply: FastifyReply) {
+	const rawBody = request.rawBody as string;
+	const timestamp = request.headers['x-slack-request-timestamp'] as string;
+	const signature = request.headers['x-slack-signature'] as string;
```
Standlc: don't cast things that can be `undefined, return 400 if you expect these headers to be set.

apps/backend/src/services/slack.service.ts
```diff
@@ -16,22 +16,15 @@ export class SlackService {
...
 	private _abortController = new AbortController();
-	private _redirectUrl = process.env.REDIRECT_URL || 'http://localhost:5005/';
+	private _redirectUrl = process.env.REDIRECT_URL || '';
 	private _slackClient: WebClient;
 
-	constructor(body: SlackRequest) {
-		if (!body.event?.text || !body.event?.channel || !body.event?.ts || !body.event?.user) {
-			throw new Error('Invalid request: missing text, channel, thread timestamp, or user ID');
-		}
+	constructor(body: ValidatedSlackRequest) {
```
Standlc: it looks like you only need the `event`, not the entire `body`?

PR #24, opened 2026-01-19
add token usage db

MatLBS:
https://github.com/user-attachments/assets/65b37051-63c7-41cd-8e94-cb91648ba420

- Added token usage details in table 'message_part

### DIFF
--- a/apps/backend/src/services/agent.service.ts
+++ b/apps/backend/src/services/agent.service.ts
@@ -1,11 +1,18 @@
 import { anthropic, AnthropicProviderOptions } from '@ai-sdk/anthropic';
 import { openai, OpenAIResponsesProviderOptions } from '@ai-sdk/openai';
-import { convertToModelMessages, createUIMessageStream, ToolLoopAgent, ToolLoopAgentSettings } from 'ai';
+import {
+	convertToModelMessages,
+	createUIMessageStream,
+	StreamTextResult,
+	ToolLoopAgent,
+	ToolLoopAgentSettings,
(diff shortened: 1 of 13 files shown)


apps/backend/src/utils/chat.ts
```diff
@@ -0,0 +1,15 @@
+import { LanguageModelUsage } from 'ai';
+
+import { TokenUsage } from '../types/chat';
+
+export const spreadTokenUsage = (usage: LanguageModelUsage): TokenUsage => ({
```
Standlc, line 5: `convertToTokenUsage` is better

apps/backend/src/services/agent.service.ts
```diff
@@ -48,6 +56,7 @@ class AgentManager {
 			tools,
 			instructions: getInstructions(),
 		});
+		this._result = {} as StreamTextResult<typeof tools, never>;
```
Standlc, line 59: you don't need `_result` to be a class member. You can simply define a `let result` inside the `stream` method.

apps/backend/src/services/agent.service.ts
Standlc, line 1: No token usage saved when generating? I assume it'll be fixed in a future PR or when you merge the fixes the `generate` method.

Standlc (approved):
Small comments, lgtm otherwise

apps/backend/src/services/agent.service.ts
MatLBS: Yes it will be fixed in the next slack PR

PR #25, opened 2026-01-19
chore: readme update semi colon missing

StanGirard:
Added a semi colon missing


README.md
Standlc, line 1: Semicolon before the last 'and' is not very common. I personally like it better without it :)

PR #29, opened 2026-01-19
Feat/add snowflake and databricks connections

MatLBS:
- Added snowflake connection with username/password
- Added snowflake connection with key-pair (encrypted or not)
- Added databricks connection with personnal access token

### DIFF
--- a/cli/nao_core/commands/init.py
+++ b/cli/nao_core/commands/init.py
@@ -19,15 +19,10 @@ from nao_core.config import (
     SlackConfig,
     SnowflakeConfig,
 )
+from nao_core.config.exceptions import InitError
 from nao_core.config.repos import RepoConfig
 
 console = Console()
-
-
-class InitError(Exception):
-    """Base exception for init command errors."""
-
-    pass
 
 
 class EmptyProjectNameError(InitError):
--- a/cli/nao_core/config/databases/snowflake.py
+++ b/cli/nao_core/config/databases/snowflake.py
@@ -7,6 +7,8 @@ from ibis import BaseBackend
 from pydantic import Field
 from rich.prompt import Confirm, Prompt
 
+from nao_core.config.exceptions import InitError
+
 from .base import DatabaseConfig, console
 
 
@@ -38,15 +40,15 @@ class SnowflakeConfig(DatabaseConfig):
 
         username = Prompt.ask("[bold]Snowflake username[/bold]")
         if not username:
-            raise ValueError("Snowflake username cannot be empty.")
+            raise InitError("Snowflake username cannot be empty.")
 
         account_id = Prompt.ask("[bold]Snowflake account identifier[/bold]")
         if not account_id:
-            raise ValueError("Snowflake account identifier cannot be empty.")
+            raise InitError("Snowflake account identifier cannot be empty.")
 
         database = Prompt.ask("[bold]Snowflake database[/bold]")
         if not database:
-            raise ValueError("Snowflake database cannot be empty.")
+            raise InitError("Snowflake database cannot be empty.")
 
         warehouse = Prompt.ask(
             "[bold]Snowflake warehouse[/bold] [dim](optional, press Enter to skip)[/dim]", default=None
@@ -59,7 +61,7 @@ class SnowflakeConfig(DatabaseConfig):
         if key_pair_auth:
             private_key_path = Prompt.ask("[bold]Path to private key file[/bold]")
             if not private_key_path:
-                raise ValueError("Path to private key file cannot be empty.")
+                raise InitError("Path to private key file cannot be empty.")
             passphrase = Prompt.ask(
                 "[bold]Passphrase for the private key[/bold] [dim](optional, press Enter to skip)[/dim]",
                 default=None,
@@ -68,7 +70,7 @@ class SnowflakeConfig(DatabaseConfig):
         else:
             password = Prompt.ask("[bold]Snowflake password[/bold]", password=True)
             if not password:
-                raise ValueError("Snowflake password cannot be empty.")
+                raise InitError("Snowflake password cannot be empty.")
 
         return SnowflakeConfig(
             name=name,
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -13,6 +13,8 @@
 		"lint:fix": "npm run lint:fix -w backend && npm run lint:fix -w frontend",
 		"format": "prettier --write .",
 		"format:check": "prettier --check .",
+		"cli:fix": "cd cli && uv run ruff check --select I --fix . && uv run ruff format .",
```
Bl3f: u should use directly the make shortcuts 'make lint-fix' and 'make lint'

cli/nao_core/config/databases/snowflake.py
```diff
@@ -0,0 +1,68 @@
...
+    private_key_path: str | None = Field(
+        default=None,
+        description="Path to private key file for key-pair authentication",
+    )
+    passphrase: str | None = Field(
+        default=None,
+        description="Passphrase for the private key if it is encrypted",
+    )
+
+    key_pair_auth: bool = Field(
```
Bl3f: why do we need a bool for this, if the key is defined use the key?

MatLBS: It was cleaner like this. 

in init.py i added this : 

```
key_pair_auth = Confirm.ask("[bold]Use key-pair authentication for authentication?[/bold]", default=False)

if key_pair_auth:
    private_key_path = Prompt.ask("[bold]Path to private key file[/bold]")
    if not private_key_path:
        raise InitError("Path to private key file cannot be empty.")
    passphrase = Prompt.ask(
        "[bold]Passphrase for the private key[/bold] [dim](optional, press Enter to skip)[/dim]",
        default=None,
        password=True,
    )
else:
    password = Prompt.ask("[bold]Snowflake password[/bold]", password=True)
    if not password:
        raise InitError("Snowflake password cannot be empty.")
```

I ask for different variables depending on the user wish

Bl3f: i don't think we need the bool variable to be stored in the config, even if it exists when asking

.husky/pre-commit
```diff
@@ -1,3 +1,5 @@
 npm run lint
 npm run format:check
-npm run db:check-migration-files -w backend
\ No newline at end of file
+npm run db:check-migration-files -w backend
+npm run cli:fix
```
Standlc: remove the `cli:fix`, that will fix the problems, which means `cli:check` won't detect errors and you'll have to do a second commit with the fixes.

cli/nao_core/config/databases/__init__.py
```diff
@@ -1,23 +1,26 @@
...
-# AnyDatabaseConfig = Annotated[
-#     Union[
-#         Annotated[BigQueryConfig, Tag("bigquery")],
-#         Annotated[PostgresConfig, Tag("postgres")],
-#     ],
-#     Discriminator(lambda x: x.get("type", "bigquery")),
-# ]
-
-AnyDatabaseConfig = Union[BigQueryConfig, DuckDBConfig]
+AnyDatabaseConfig = Annotated[
```
Standlc, line 15: What's the point of `Annotated` here? Looks kinda useless

cli/nao_core/commands/init.py
```diff
@@ -120,6 +122,90 @@ def setup_duckdb() -> DuckDBConfig:
...
+            "[bold]Passphrase for the private key[/bold] [dim](optional, press Enter to skip)[/dim]",
+            default=None,
+            password=True,
+        )
+    else:
+        password = Prompt.ask("[bold]Snowflake password[/bold]", password=True)
+        if not password:
+            raise InitError("Snowflake password cannot be empty.")
+
+    return SnowflakeConfig(
```
Standlc: Wouldn't it be cleaner to have each `XXXConfig` handle the user prompting with a method? 

For example:
```python
config = SnowflakeConfig.promptConfig()
```

cli/nao_core/commands/sync/databases.py
```diff
@@ -164,6 +164,155 @@ def sync_duckdb(
...
+            base_path: Base output path
+            progress: Rich progress instance
+            accessors: List of data accessors to run
+
+    Returns:
+            Tuple of (schemas_synced, tables_synced)
+    """
+    conn = db_config.connect()
+    catalog = db_config.catalog or "main"
+    db_path = base_path / "type=databricks" / f"database={catalog}"
```
Standlc, line 186: There's a lot of repetition in the directory names. There should be abstracted logic for all providers. Also, shouldn't that code be in a databricks specific directory?

.husky/pre-commit
MatLBS: ? `clix:fix` will correct the errors before `cli:check` as I added both of them in the pre-commit action from huski so no need for a second commit

cli/nao_core/config/databases/__init__.py
MatLBS: Here's what they show is the pydantic doc : 

```
dessert: Annotated[
    Union[
        Annotated[ApplePie, Tag('apple')],
        Annotated[PumpkinPie, Tag('pumpkin')],
    ],
    Discriminator(get_discriminator_value),
]
```

cli/nao_core/config/databases/snowflake.py
MatLBS: Done

cli/nao_core/commands/init.py
MatLBS: Good idea

package.json
Bl3f: you did not fix my comment here?

cli/nao_core/config/databases/__init__.py
Standlc: ok but do you know what it's for?

cli/nao_core/commands/sync/databases.py
MatLBS: Could that be done in another PR ?

cli/nao_core/config/databases/__init__.py
MatLBS: The first one is for the 'Discriminator' and the other ones for the tags

PR #36, opened 2026-01-20
Feat/improve slack integration

MatLBS:
- Use agent.stream for Slack response
- Tools and tokens stored to db
- Agent has memory

Capture d’écran 2026-01-20 à 17 33 22

https://github.com/user-attachments/assets/268275a9-99bb-4c7a-8630-459e1a4678de

### DIFF
--- a/apps/backend/src/routes/slack.ts
+++ b/apps/backend/src/routes/slack.ts
@@ -1,8 +1,9 @@
 import type { App } from '../app';
 import { slackAuthMiddleware } from '../middleware/slack.middleware';
+import * as chatQueries from '../queries/chat.queries';
+import { agentService } from '../services/agent.service';
 import { SlackService } from '../services/slack.service';
-import { SlackInteractionPayload, SlackRequestSchema } from '../types/slack';
-import { activeSlackStreams } from '../utils/slack';
+import { SlackInteractionBodySchema, SlackInteractionPayloadSchema, SlackRequestSchema } from '../types/slack';
 
 export const slackRoutes = async (app: App) => {
 	// Verifying requests from Slack : verify whether requests from Slack are authentic
@@ -22,10 +23,8 @@ export const slackRoutes = async (app: App) => {
 				return reply.send({ challenge: body.challenge });
 			}
 
-			if (!process.env.SLACK_BOT_TOKEN || !process.env.SLACK_SIGNING_SECRET) {
-				return reply
-					.status(400)
-					.send({ error: 'SLACK_BOT_TOKEN or SLACK_SIGNING_SECRET is not defined in environment variables' });
+			if (!process.env.SLACK_BOT_TOKEN) {
+				return reply.status(400).send({ error: 'SLACK_BOT_TOKEN is not defined in environment variables' });
 			}
 
 			if (!body.event) {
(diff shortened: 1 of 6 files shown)


package-lock.json
```diff
@@ -62,6 +62,7 @@
 			"integrity": "sha512-W609buLVRVmeW693xKfzHeIV6nJGGz98uCPfeXI1ELMLXVeKYZ9m15fAMSaUPBHYLGFsVRcMmSCksQOrZV9BYA==",
 			"dev": true,
 			"license": "MIT",
+			"peer": true,
```
Bl3f, line 66: why these changes again?

MatLBS: idk, I'm going to revert them

apps/backend/src/routes/slack.ts
```diff
@@ -38,9 +39,42 @@ export const slackRoutes = async (app: App) => {
...
 		},
 	);
+
+	app.post(
+		'/interactions',
+		{
+			config: { rawBody: true },
+		},
+		async (request, reply) => {
+			const body = request.body as { payload: string };
```
Standlc: why `rawBody` in the config? Why don't you validate the body like normal and then you won't need to parse it nor cast is `as { payload: string }`.

apps/backend/src/services/slack.service.ts
```diff
@@ -77,37 +89,92 @@ export class SlackService {
...
+
+		await this._processStream(stream);
+		await this._replaceStopButtonWithLink(chatId);
+		activeSlackStreams.delete(this._threadId);
+	}
 
-		await this._slackClient.chat.postMessage({
+	private _createAgentStream(chat: UIChat, isNew: boolean) {
+		const agent = agentService.create({ ...chat, userId: this._user.id }, this._abortController);
+		return agent.stream(chat.messages as UIMessage[], { sendNewChatData: !!isNew });
```
Standlc: remove the cast

Standlc: I don't think you need to send the `data-newChat` event at all. You can disregard this option or always send `false` here.

apps/backend/src/utils/slack.ts
```diff
@@ -22,9 +22,40 @@ const createTextMessage = (text: string, role: 'system' | 'user' | 'assistant'):
 	return message;
 };
 
-const saveSlackAgentResponse = async (chatId: string, responseText: string) => {
-	const assistantMessage = createTextMessage(responseText, 'assistant');
-	await chatQueries.upsertMessage(assistantMessage, { chatId });
+const extractLastTextFromMessage = (message: { parts: { type: string; text?: string }[] }): string => {
```
Standlc: that function should be in a ai related `utils.ts`, not slack

apps/backend/src/services/slack.service.ts
```diff
@@ -77,37 +89,92 @@ export class SlackService {
...
 			channel: this._channel,
-			text: fullMessage,
+			text: 'Generating response... ',
+			blocks: [addButtonStopBlock()],
 			thread_ts: this._threadTs,
 		});
+		this._buttonTs = buttonMessage.ts;
+	}
+
+	private async _processStream(stream: ReadableStream<UIMessageChunk<unknown, UIDataTypes>>): Promise<void> {
```
Standlc: rename to `readStreamAndUpdateSlackMessage`

apps/backend/src/routes/slack.ts
```diff
@@ -38,9 +39,42 @@ export const slackRoutes = async (app: App) => {
...
+
+			if (!body.payload) {
+				return reply.status(400).send({ error: 'Missing payload' });
+			}
+
+			const payload = JSON.parse(body.payload) as SlackInteractionPayload;
+
+			if (payload.type === 'block_actions' && payload.actions) {
```
Standlc: use early returns please

apps/backend/src/utils/slack.ts
Standlc: The functions are too factorized lol. Delete this function and create your object inline. When you create a function `createTextMessage` we don't know if it creates it in the db, so we have to look around to check what it does and end up spending time understand it's purpose when the only thing it does is return an object.

Standlc: this shouldn't be a function, it already does one thing. Move the code to where it's used in the slack service.

apps/backend/src/routes/slack.ts
```diff
@@ -38,9 +39,42 @@ export const slackRoutes = async (app: App) => {
...
+					if (action.action_id === 'stop_generation') {
+						const channel = payload.channel?.id;
+						const threadTs = payload.message?.thread_ts || payload.message?.ts;
+
+						if (channel && threadTs) {
+							const threadId = [channel, threadTs.replace('.', '')].join('/p');
+							const abortController = activeSlackStreams.get(threadId);
+
+							if (abortController) {
+								abortController.abort();
```
Standlc: why don't you use the `agentService` which already has a map of running agents?

apps/backend/src/services/slack.service.ts
MatLBS: But for the first message in a slack conv, should not we send `data-newChat` ?

apps/backend/src/utils/slack.ts
MatLBS: Okay so I should just renamed it. If I create object inline I will just duplicated code ?

MatLBS: I moved the code into slack.service ( _saveOrUpdateUserMessage() ) but I don't really understand your point.

As the parent function does 2 things, create or update, that's why I created 2 separated functions...

PR #45, opened 2026-01-22
remove the cache on logout which fix the chat initialization on login

### LINKED ISSUE #44: When logout then login, the app is wrongly initialised
When you logout and then login in the same browser, the app is initialised with the data from the account that has been logged out.

MatLBS:
Fixes #44 

- Clear cache on logout
- Chats in the sidebar are initialized correctly when on another user connection

### DIFF
--- a/apps/frontend/src/components/header.tsx
+++ b/apps/frontend/src/components/header.tsx
@@ -1,4 +1,5 @@
 import { useNavigate, Link } from '@tanstack/react-router';
+import { useQueryClient } from '@tanstack/react-query';
 import { LogOut } from 'lucide-react';
 import { ButtonConnection } from './ui/button';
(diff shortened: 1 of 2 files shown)


apps/frontend/src/main.tsx
```diff
@@ -29,7 +29,7 @@ const router = createRouter({
 });
 
 /** Query client for state management */
-const queryClient = new QueryClient({
+export const queryClient = new QueryClient({
```
Standlc: i'm not again exporting it, but the defacto in the codebase is to get a hold of the `queryClient` through `useQueryClient`. Either you revert the export and use the hook where you need it, or remove the hook everywhere it's used and import the `queryClient` directly instead.

MatLBS: I removed the export and use 'useQueryClient'

PR #46, opened 2026-01-22
Display charts in Chat with agent tool

Standlc:
## Charts in Chat

- agent can display charts with the `display_chart` tool
  - the tool takes the config for the chart (type, series, etc) and the `id` of a `execute_sql` tool result
  - supports: line, bar and pie

- charts are rendered with `recharts` and wrapped with `shadcn` components for style

TODO:
- improve loading and error state in chart component
- add nicer chart colors
- check that the chart is displayable in the `display_chart` execute function
- show multiple charts inside a carousel (navigate left and right)?
- fix annoying tooltip jumping when hovering


apps/frontend/src/components/tool-call/tool-call-wrapper.tsx
```diff
@@ -67,20 +67,24 @@ export const ToolCallWrapper = ({
...
 		>
-			<AccordionItem value='tool-content' className={cn('border-b-0')} style={{ padding: 0 }} id={'hey'}>
+			<AccordionItem value='tool-content' className={cn('border-b-0')} style={{ padding: 0 }}>
 				{isBordered ? (
 					<div
 						className={cn(
-							'flex items-center justify-between gap-2 px-3 py-2',
+							'flex items-center justify-between gap-2 py-2',
 							canExpand && 'cursor-pointer',
+							isBordered && 'px-3',
```
Bl3f: as you fixed the padding can you also fix:
1. the padding for the thinking display
2. make the left chevron aligned with the all the text, looks like it isn't

Image

apps/frontend/src/components/tool-call/display-chart.tsx
```diff
@@ -0,0 +1,220 @@
...
+import { useToolCallContext } from '../../contexts/tool-call.context';
+import { useAgentContext } from '../../contexts/agent.provider';
+import { ChartContainer, ChartTooltip, ChartTooltipContent, ChartLegend, ChartLegendContent } from '../ui/chart';
+import { ToolCallWrapper } from './tool-call-wrapper';
+import type { CategoricalChartProps } from 'recharts/types/chart/generateCategoricalChart';
+import type { ChartConfig } from '../ui/chart';
+import type { displayChartSchemas, executeSqlSchemas } from 'backend/tools';
+import { toLabel } from '@/lib/charts.utils';
+
+const DEFAULT_COLORS = ['var(--chart-1)', 'var(--chart-2)', 'var(--chart-3)', 'var(--chart-4)', 'var(--chart-5)'];
```
Bl3f, line 12: Will be great next option to give the possibility to user to give color palette - next thing is even to have consistent color paletter which makes a specific item always the same color across all charts

apps/frontend/src/components/tool-call/display-chart.tsx
```diff
@@ -0,0 +1,220 @@
...
+				<div className='p-4 text-center text-foreground/50 text-sm'>No data to display</div>
+			</ToolCallWrapper>
+		);
+	}
+
+	const data = sourceData.data;
+
+	const renderChart = (options: CategoricalChartProps, tooltipOptions: React.ComponentProps<typeof ChartTooltip>) => {
+		if (input.chart_type === 'bar' || input.chart_type === 'line') {
+			const Chart = input.chart_type === 'bar' ? BarChart : AreaChart;
```
Bl3f, line 112: would be great to add an option where we can see the settings that have been passed to the chart so we can debug (and later reproduce) a bit like the <> on the execute sql where you see the SQL code

visually if we can also link the chart generated to the sql query would be banger, at least with a jump?

apps/backend/src/agents/tools/schema/display-chart.ts
```diff
@@ -0,0 +1,40 @@
+import z from 'zod/v3';
+
+export const chartTypeEnum = z.enum(['bar', 'line', 'pie']);
```
Bl3f, line 3: pie chat is more of a joke than something useful i think, but tho in the long run we have to think better how to structure this because im ok to start with a type but maybe it's a composable thing we have to do where we let the LLM pick what we need in axes and then series representations like
- x axis
- y axis left
- y axis right
- serie 1 bar
- serie 2 line

also we need stacked charts, etc. 

what i mean is type is unidimension where actually chart type can have either subtype or be composable

apps/frontend/src/components/ui/chart.tsx
```diff
@@ -0,0 +1,309 @@
...
+	const uniqueId = React.useId();
+	const chartId = `chart-${id || uniqueId.replace(/:/g, '')}`;
+
+	return (
+		<ChartContext.Provider value={{ config }}>
+			<div
+				data-slot='chart'
+				data-chart={chartId}
+				className={cn(
+					"[&_.recharts-cartesian-axis-tick_text]:fill-muted-foreground [&_.recharts-cartesian-grid_line[stroke='#ccc']]:stroke-border/50 [&_.recharts-curve.recharts-tooltip-cursor]:stroke-border [&_.recharts-polar-grid_[stroke='#ccc']]:stroke-border [&_.recharts-radial-bar-background-sector]:fill-muted [&_.recharts-rectangle.recharts-tooltip-cursor]:fill-muted [&_.recharts-reference-line_[stroke='#ccc']]:stroke-border flex aspect-video justify-center text-xs [&_.recharts-dot[stroke='#fff']]:stroke-transparent [&_.recharts-layer]:outline-hidden [&_.recharts-sector]:outline-hidden [&_.recharts-sector[stroke='#fff']]:stroke-transparent [&_.recharts-surface]:outline-hidden",
```
Bl3f, line 51: 🙃

PR #48, opened 2026-01-22
Add env support in `nao_config.yaml` file

### LINKED ISSUE #34: Support {{ env_var }} in the nao_config.yaml
We should support env variable read in the nao_config.yaml so we can define some values from the process deploying the project.

The syntax should be for instance 

```
llm:
    - provider: openai
      api_key: {{ env_var('OPENAI_API_KEY') }}
```

Can also be `{{ env('KEY') }}` function, depends which one feels more natural.

paulsojan:
Fixes #34

Now `nao_config.yaml` can read env variables.

Demo: https://paulsojan.neetorecord.com/watch/487a05fb213ccf5c4c21

I have kept the syntax like this:

```yaml
llm:
    provider: openai
    api_key: ${{ env('OPENAI_API_KEY') }}
```
I used the GitHub workflow syntax style to access the environment from a YAML file.
Ref: https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#using-contexts-to-access-variable-values

### DIFF
--- a/cli/nao_core/config/base.py
+++ b/cli/nao_core/config/base.py
@@ -49,7 +49,6 @@ class NaoConfig(BaseModel):
         """Load the configuration from a YAML file."""
         config_file = path / "nao_config.yaml"
         content = config_file.read_text()
-
         content = cls._process_env_vars(content)
         data = yaml.safe_load(content)
         return cls.model_validate(data)
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     "uvicorn>=0.40.0",
     "cryptography>=46.0.3",
     "dotenv>=0.9.9",
+    "pytest>=9.0.2",
 ]
 
 [project.scripts]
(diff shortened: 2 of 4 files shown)


cli/nao_core/config/base.py
```diff
@@ -43,8 +48,10 @@ def save(self, path: Path) -> None:
     def load(cls, path: Path) -> "NaoConfig":
         """Load the configuration from a YAML file."""
         config_file = path / "nao_config.yaml"
-        with config_file.open() as f:
-            data = yaml.safe_load(f)
+        content = config_file.read_text()
+
```
Bl3f: This is a useless newline in this context I think.

cli/nao_core/config/base.py
```diff
@@ -76,3 +83,13 @@ def try_load(cls, path: Path | None = None) -> "NaoConfig | None":
     def json_schema(cls) -> dict:
         """Generate JSON schema for the configuration."""
         return cls.model_json_schema()
+
+    @staticmethod
+    def _process_env_vars(content: str) -> str:
```
Bl3f, line 87: We don't have unit tests yet (it's a task we have to work on) but it would be great to have this method tested at least.

paulsojan: @Bl3f I have added tests for the method.

PR #49, opened 2026-01-22
Add project management + llm and slack settings in the user page

Bl3f:
Add project management in the Settings page, how it works:
- if the `NAO_DEFAULT_PROJECT_PATH` is defined then at first login/signup it will attribute this project to the user (first one gets admin)
- then you can edit this project params (LLM or Slack)
- there is for the moment only one project per instance launched (multi project will come after)

Screenshot 2026-01-22 at 16 03 35

Screenshot 2026-01-22 at 16 04 11

### DIFF
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -2,7 +2,6 @@ import { and, eq } from 'drizzle-orm';
 
 import s, { DBProject, DBProjectMember, NewProject, NewProjectMember } from '../db/abstractSchema';
 import { db } from '../db/db';
-import * as llmConfigQueries from './project-llm-config.queries';
 import * as userQueries from './user.queries';
 
 export const getProjectByPath = async (path: string): Promise<DBProject | null> => {
@@ -94,21 +93,6 @@ export const initializeDefaultProjectForFirstUser = async (userId: string): Prom
 		type: 'local',
 		path: projectPath,
 	});
-
-	const openaiKey = process.env.OPENAI_API_KEY;
-	const anthropicKey = process.env.ANTHROPIC_API_KEY;
-
-	if (openaiKey) {
-		await llmConfigQueries.upsertProjectLlmConfig({ projectId: project.id, provider: 'openai', apiKey: openaiKey });
-	}
-
-	if (anthropicKey) {
-		await llmConfigQueries.upsertProjectLlmConfig({
-			projectId: project.id,
-			provider: 'anthropic',
-			apiKey: anthropicKey,
-		});
-	}
 
 	await addProjectMember({
 		projectId: project.id,
--- a/apps/frontend/src/routes/user.tsx
+++ b/apps/frontend/src/routes/user.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 import { createFileRoute, useNavigate } from '@tanstack/react-router';
-import { useQueryClient, useQuery  } from '@tanstack/react-query';
+import { useQueryClient, useQuery } from '@tanstack/react-query';
 import { Input } from '@/components/ui/input';
 import { Sidebar } from '@/components/sidebar';
 import { signOut, useSession } from '@/lib/auth-client';


apps/backend/src/db/pgSchema.ts
```diff
@@ -84,6 +108,9 @@ export const chat = pgTable(
 		userId: text('user_id')
 			.notNull()
 			.references(() => user.id, { onDelete: 'cascade' }),
+		projectId: text('project_id')
```
Standlc, line 124: Shouldn't a `chat` belong to a `project_member`? If it has a `user_id` and `project_id`, it could, if some mistake happens when saving a chat, belong to a `project_id` who the `user_id` doesn't belong to.

I think it's acceptable this way, but it would be more normalized with a `project_member_id`.

apps/backend/src/db/pgSchema.ts
```diff
@@ -189,3 +220,60 @@ export const messageFeedback = pgTable('message_feedback', {
...
+	(t) => [primaryKey({ columns: [t.projectId, t.userId] }), index('project_member_userId_idx').on(t.userId)],
+);
+
+export const projectLlmConfig = pgTable(
+	'project_llm_config',
+	{
+		id: text('id')
+			.$defaultFn(() => crypto.randomUUID())
+			.primaryKey(),
+		projectId: text('project_id')
```
Standlc, line 258: unique constraint?

apps/backend/src/middleware/auth.ts
```diff
@@ -20,4 +23,5 @@ export async function authMiddleware(request: FastifyRequest, reply: FastifyRepl
 
 	request.user = session.user;
 	request.session = session.session;
+	request.project = await ensureDefaultProjectMembership(session.user.id);
```
Standlc: i think `projectQueries.ensureDefaultProjectMembership()` reads better

apps/backend/src/middleware/slack.middleware.ts
```diff
@@ -13,13 +15,57 @@ function verifySlackSignature(signingSecret: string, requestSignature: string, t
...
+ * Get Slack configuration from project config with env var fallbacks.
+ * This is the single source of truth for all Slack config values.
+ */
+export async function getSlackConfig(): Promise<SlackConfig | null> {
+	const projectPath = process.env.NAO_DEFAULT_PROJECT_PATH;
+	if (!projectPath) {
+		return null;
+	}
+
+	const project = await getProjectByPath(projectPath);
```
Standlc: add `projectQueries` so we know it's a db query.

apps/backend/src/middleware/slack.middleware.ts
```diff
@@ -13,13 +15,57 @@ function verifySlackSignature(signingSecret: string, requestSignature: string, t
...
+	if (!project) {
+		return null;
+	}
+
+	const dbConfig = await getProjectSlackConfig(project.id);
```
Standlc: you could left join the slack config and do only one query

apps/backend/src/middleware/auth.ts
Standlc: Also, as we discussed, a middleware is usually meant to protect a route, not to mutate the db as a side effect.

Why not add this to the chat route?

apps/backend/src/queries/project.queries.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export const getUserRoleInProject = async (
+	projectId: string,
+	userId: string,
+): Promise<'admin' | 'user' | 'viewer' | null> => {
+	const member = await getProjectMember(projectId, userId);
+	return member?.role ?? null;
+};
+
+export const ensureDefaultProjectMembership = async (userId: string): Promise<DBProject | null> => {
```
Standlc: should be called `checkUserHasProjectOrCreate`

apps/backend/src/queries/project.queries.ts
```diff
@@ -0,0 +1,167 @@
...
+			projectId: project.id,
+			userId,
+			role: 'user',
+		});
+	}
+
+	return project;
+};
+
+// LLM Config queries
```
Standlc: you said it in the comment... These are llm config queries, not project queries.

apps/backend/src/middleware/slack.middleware.ts
```diff
@@ -13,13 +15,57 @@ function verifySlackSignature(signingSecret: string, requestSignature: string, t
...
+	botToken: string;
+	signingSecret: string;
+	redirectUrl: string;
+}
+
+/**
+ * Get Slack configuration from project config with env var fallbacks.
+ * This is the single source of truth for all Slack config values.
+ */
+export async function getSlackConfig(): Promise<SlackConfig | null> {
```
Standlc: that's not a middleware but a util

apps/frontend/src/components/settings-slack-config-section.tsx
```diff
@@ -0,0 +1,262 @@
...
+import { Check, Copy, ExternalLink, Pencil, Plus, Trash2 } from 'lucide-react';
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+import { trpc } from '@/main';
+
+interface SlackConfigSectionProps {
+	isAdmin: boolean;
+}
+
+function CopyableUrl({ label, url }: { label: string; url: string }) {
```
Standlc: seems pretty abstract, should go in `components/ui/`

apps/frontend/src/components/settings-slack-config-section.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+	const projectConfig = slackConfig.data?.projectConfig;
+	const hasEnvConfig = slackConfig.data?.hasEnvConfig ?? false;
+
+	const handleSaveConfig = async () => {
+		if (!botToken || !signingSecret) {
+			return;
+		}
+		await upsertSlackConfig.mutateAsync({ botToken, signingSecret });
+		queryClient.invalidateQueries(trpc.project.getSlackConfig.queryOptions());
```
Standlc, line 105: tip: after a mutation, you can either invalidate the query/cache data (which will trigger a re-fetch) or `queryClient.setQueryData(trpc.project.getSlackConfig.queryKey(), ...new data...)` to change the data in local without doing another request to the server.

apps/frontend/src/components/ui/settings-card.tsx
```diff
@@ -0,0 +1,22 @@
+import { cn } from '@/lib/utils';
+
+interface SettingsCardProps {
+	title?: string;
+	badge?: React.ReactNode;
+	children: React.ReactNode;
+	className?: string;
+}
+
+export function SettingsCard({ title, badge, children, className }: SettingsCardProps) {
```
Standlc, line 10: in `components/ui`?

apps/backend/src/auth.ts
```diff
@@ -3,6 +3,51 @@ import { drizzleAdapter } from 'better-auth/adapters/drizzle';
 
 import { db } from './db/db';
 import dbConfig, { Dialect } from './db/dbConfig';
+import { addProjectMember, createProject, getProjectByPath } from './queries/project.queries';
+import { countUsers } from './queries/user.queries';
+
+const initializeDefaultProjectForFirstUser = async (userId: string): Promise<void> => {
```
Standlc: move to `user.queries` or some utils file?

apps/backend/src/trpc/router.ts
```diff
@@ -1,17 +1,26 @@
...
 	chat: chatRoutes,
 	feedback: feedbackRoutes,
+	project: projectRoutes,
 	user: userRoutes,
 	hasGoogleSetup: publicProcedure.query(() => {
 		return !!(process.env.GOOGLE_CLIENT_ID && process.env.GOOGLE_CLIENT_SECRET);
 	}),
-	getModelProvider: publicProcedure.query(() => {
-		return process.env.ANTHROPIC_API_KEY ? 'anthropic' : process.env.OPENAI_API_KEY ? 'openai' : undefined;
+	getModelProvider: projectProtectedProcedure.query(async ({ ctx }) => {
```
Standlc: should be moved to `projectRoutes` so you can do `trpc.project.getModelProvider`

apps/backend/migrations-sqlite/meta/0005_snapshot.json
Standlc, line 1: No migration for the required `project_id` in the chats? The migration probably won't work unless you delete the db

Bl3f: Actually in the migrations I added `DELETE FROM chat`, that ensure there a no leftover conversations. It's brutal, but as it's the beginning i don't care about deleting chats(?)

apps/backend/src/middleware/auth.ts
Bl3f: I added a comment, the mutation is done only for the signup of user other than admin. I will remove the mutation as anyway we have to do differently the signup process.

To me it doesnt make sense to be a route as it's not something that is triggered by the frontend but more backend hook

apps/backend/src/queries/project.queries.ts
Bl3f: but llm are project related objects so 🤔

apps/backend/src/db/pgSchema.ts
Bl3f: Not sure about this, will think a bit about it

Bl3f: Let's keep it this way for now.

PR #50, opened 2026-01-22
fixed relative path issue duckdb

### LINKED ISSUE #33: DuckDB path relative to FastAPI
When using a path in the `nao_config.yaml` for a DuckDB database, the path should be relative the nao_config file rather than from where the FastAPI server is launched.

MatLBS:
Fix #33 

- Pass 'project_path' for duckdb connection which is the absolute path to  'nao_config.yaml'. Then join it with db path in the config file


apps/backend/fastapi/main.py
```diff
@@ -83,7 +83,10 @@ async def execute_sql(request: ExecuteSQLRequest):
                 }
             )
 
-        connection = db_config.connect()
+        if db_config.type == "duckdb":
```
Bl3f: why this if?

Bl3f (commented):
I'd have tried to fix the issue globally at FastAPI level using this DuckDB example, because if we have another path managed in a configuration it will be the same issue

PR #53, opened 2026-01-23
Display errors in the chat

Bl3f:
https://github.com/user-attachments/assets/c24619d2-bdd9-4eb0-b486-ef12fca6e419

### DIFF
--- a/apps/frontend/src/hooks/use-agent.ts
+++ b/apps/frontend/src/hooks/use-agent.ts
@@ -1,6 +1,6 @@
 import { useNavigate, useParams } from '@tanstack/react-router';
 import { useMutation, useQueryClient } from '@tanstack/react-query';
-import { useMemo, useEffect, useRef, useCallback, useState } from 'react';
+import { useMemo, useEffect, useRef, useCallback } from 'react';
 import { Chat as Agent, useChat } from '@ai-sdk/react';
 import { DefaultChatTransport } from 'ai';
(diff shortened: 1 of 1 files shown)


apps/frontend/src/hooks/use-agent.ts
```diff
@@ -31,6 +33,7 @@ export const useAgent = (): AgentHelpers => {
 	const queryClient = useQueryClient();
 	const chatIdRef = useCurrent(chatId);
 	const scrollDownService = useScrollDownCallbackService();
+	const [error, setError] = useState<Error | null>(null);
```
Standlc: `agentInstance` already has an `error` and a `clearError` method. Any reason not to use it?

apps/frontend/src/components/chat-error.tsx
```diff
@@ -0,0 +1,46 @@
+import { AlertCircleIcon } from 'lucide-react';
+
+import { useAgentContext } from '@/contexts/agent.provider';
+import { cn } from '@/lib/utils';
+
+export interface Props {
+	className?: string;
+}
+
+type ParsedError = {
```
Standlc, line 10: is this the error format what the ai sdk sends?

Bl3f: Yes it's an APICallError which inherits from AISDKError, but error is not always filled, depends how it's raised from what i understood

PR #55, opened 2026-01-23
Group and collpase consecutive tool calls

Standlc:
- group subsequent tool calls
- expand while streaming & auto collapse when done

https://github.com/user-attachments/assets/cef7451c-2191-4d31-9039-969b2614e7ff

### DIFF
--- a/apps/frontend/src/main.tsx
+++ b/apps/frontend/src/main.tsx
@@ -16,6 +16,9 @@ declare module '@tanstack/react-router' {
 	interface Register {
 		router: typeof router;
 	}
+	interface HistoryState {
+		fromMessageSend?: boolean;
+	}
 }
 
 // Create a new router instance
--- a/apps/frontend/src/components/chat-message-reasoning-accordion.tsx
+++ b/apps/frontend/src/components/chat-message-reasoning-accordion.tsx
@@ -23,7 +23,7 @@ export const ReasoningAccordion = ({ text, isStreaming }: ReasoningAccordionProp
 
 	return (
 		<Expandable
-			title={isStreaming ? 'Thinking...' : 'Thought'}
+			title={isStreaming ? 'Thinking' : 'Thought'}
 			expanded={isExpanded}
 			onExpandedChange={setIsExpanded}
 			isLoading={isStreaming}
--- a/apps/frontend/src/components/chat-messages.tsx
+++ b/apps/frontend/src/components/chat-messages.tsx
@@ -1,6 +1,6 @@
 import { Streamdown } from 'streamdown';
 import { useEffect, useMemo, useRef } from 'react';
-import { useParams } from '@tanstack/react-router';
+import { useParams, useRouterState } from '@tanstack/react-router';
 import { useStickToBottomContext } from 'use-stick-to-bottom';
 import { ToolCall } from './tool-call';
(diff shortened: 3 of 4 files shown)


apps/frontend/src/lib/ai.ts
```diff
@@ -55,6 +58,63 @@ export const checkIsAgentGenerating = (
 	});
 };
 
-export const checkIsAgentRunning = (status: UseChatHelpers<UIMessage>['status']) => {
-	return status === 'streaming' || status === 'submitted';
+export const checkIsAgentRunning = (agent: Pick<UseChatHelpers<UIMessage>, 'status'>) => {
```
Bl3f, line 61: What does it brings to change the typing like except the fact it's more handy to pass an agent and dont care from the outside?

apps/frontend/src/lib/ai.ts
```diff
@@ -55,6 +58,63 @@ export const checkIsAgentGenerating = (
...
+	return agent.status === 'streaming' || agent.status === 'submitted';
+};
+
+/** Tools that should NOT be collapsed (important UI elements) */
+export const NON_COLLAPSIBLE_TOOLS: StaticToolName[] = ['execute_sql', 'display_chart'];
+
+/** Check if a part is a reasoning part */
```
Bl3f, line 68: this kind of comment is a bit useless, sometimes i also forget to remove them but not sure they bring important information

apps/frontend/src/lib/ai.ts
```diff
@@ -55,6 +58,63 @@ export const checkIsAgentGenerating = (
...
+				result.push({ type: 'tool-group', parts: [...currentGroup] });
+			}
+			currentGroup = [];
+		}
+	};
+
+	for (const part of parts) {
+		if (isCollapsiblePart(part)) {
+			currentGroup.push(part);
+		} else if (part.type === 'text' || isToolUIPart(part)) {
```
Bl3f, line 100: is a condition mandatory in this if?

apps/frontend/src/main.tsx
```diff
@@ -16,6 +16,9 @@ declare module '@tanstack/react-router' {
 	interface Register {
 		router: typeof router;
 	}
+	interface HistoryState {
```
Bl3f, line 19: is it used?

apps/frontend/src/components/tool-call/tool-calls-group.tsx
```diff
@@ -0,0 +1,61 @@
...
+	useEffect(() => {
+		setIsExpanded(isLoading);
+	}, [isLoading]);
+
+	const errorCount = useMemo(() => {
+		return parts.filter((part) => !isReasoningPart(part) && !!part.errorText).length;
+	}, [parts]);
+
+	const title = isLoading
+		? `Exploring${errorCount ? ` (${errorCount} error${errorCount > 1 ? 's' : ''})` : ''}`
```
Bl3f, line 29: should have a plural() method for this somewhere

apps/frontend/src/lib/ai.ts
Standlc: it's to include text or other tools as parts, other part types get disregarded

apps/frontend/src/lib/ai.ts
Standlc: true

PR #57, opened 2026-01-23
Improved scroll down to actual message + indicator when there is message history

SIMOUNIX:
- change the animation to look better and actually be smooth 

[Screencast from 2026-01-23 16-47-26.webm](https://github.com/user-attachments/assets/d2df030a-8548-4a1f-ab1f-5b62cbf66330)

looks scrappy within the video but mainly due of the computer performance when recording.. animation works correctly with big scrolling height but less when very short distance between actual position and bottom position tho

### DIFF
--- a/apps/frontend/src/components/sidebar-user-menu.tsx
+++ b/apps/frontend/src/components/sidebar-user-menu.tsx
@@ -24,7 +24,7 @@ export function SidebarUserMenu({ isCollapsed }: SidebarUserMenuProps) {
 			}}
 			className={cn(
 				'flex items-center justify-between border-sidebar-border cursor-pointer rounded-lg',
-				'hover:bg-sidebar-accent transition-[background-color,padding]',
+				'hover:bg-sidebar-accent transition-[background-color,padding] duration-300',
 				isCollapsed ? 'p-1.5' : 'p-3 py-2',
 			)}
 		>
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/chat-messages.tsx
```diff
@@ -29,6 +29,10 @@ export function ChatMessages() {
 	const chatId = useParams({ strict: false }).chatId;
 	const contentRef = useRef<HTMLDivElement>(null);
 	const containerHeight = useHeight(contentRef, [chatId]);
+	const { messages, status } = useAgentContext();
+	const isAgentGenerating = checkIsAgentGenerating({ status, messages });
+	const lastMessageRole = messages.at(-1)?.role;
+	const shouldSmoothResize = !isAgentGenerating && lastMessageRole === 'user';
```
Standlc: rename to `shouldResizeSmootly`

apps/frontend/src/components/chat-messages.tsx
```diff
@@ -29,6 +29,10 @@ export function ChatMessages() {
 	const chatId = useParams({ strict: false }).chatId;
 	const contentRef = useRef<HTMLDivElement>(null);
 	const containerHeight = useHeight(contentRef, [chatId]);
+	const { messages, status } = useAgentContext();
+	const isAgentGenerating = checkIsAgentGenerating({ status, messages });
```
Standlc, line 33: you can probably just use the `isRunning` property from the `useAgentContext` here.

Standlc: If not, make sure to call `checkIsAgentGenerating` once and pass the value down to `ChatMessagesContent` instead of calculating it twice.

Issue #56, opened 2026-01-23
Provide Docker images to deploy nao


Bl3f:
We should provide docker images for users to be able to deploy nao as a container.

# How should it work

In order to launch nao, users should be able to `docker run getnao/nao` and be able to run the image.

The env variables available to configure the Docker file should be documented in the README and on Dockerhub.

PR #59, opened 2026-01-23
Feat/add docker

Bl3f:
Add Dockerfile + CI script to provide docker images to deploy nao chat UI (#56)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -54,6 +54,7 @@ jobs:
               uses: docker/build-push-action@v6
               with:
                   context: .
+                  platforms: linux/amd64,linux/arm64
                   push: ${{ github.event_name != 'pull_request' }}
                   tags: ${{ steps.meta.outputs.tags }}
                   labels: ${{ steps.meta.outputs.labels }}


.github/workflows/docker.yml
```diff
@@ -0,0 +1,98 @@
+name: Docker Build & Deploy
+
+on:
+    push:
+        branches:
+            - main
+            - '**docker'
+    pull_request:
```
Standlc, line 8: why do we want to run this on PRs?

Dockerfile
```diff
@@ -0,0 +1,112 @@
...
+FROM base AS frontend-builder
+WORKDIR /app
+
+COPY package.json package-lock.json bun.lock ./
+COPY apps/frontend/package.json ./apps/frontend/
+COPY apps/backend/package.json ./apps/backend/
+
+RUN npm ci
+
+COPY apps/frontend ./apps/frontend
```
Standlc, line 21: why not copy the apps before `npm ci`?

apps/backend/src/cli.ts
Standlc, line 176: so launching the server from here doesn't do the "user belongs to a project" check?

docker/supervisord.conf
```diff
@@ -0,0 +1,27 @@
...
+user=nao
+autostart=true
+autorestart=true
+stdout_logfile=/dev/stdout
+stdout_logfile_maxbytes=0
+stderr_logfile=/dev/stderr
+stderr_logfile_maxbytes=0
+
+[program:backend]
+command=bun run apps/backend/src/cli.ts serve
```
Standlc, line 19: I don't think running a Fastify server with `bun` is recommended (Fastify is really built for Node). Why don't we build the server and run it with node?

Dockerfile
Bl3f: so npm ci step can be cached

to explain: if you just copy the package.json then npm ci, the npm ci will only rerun if package.json changes, whereas if you copy the whole apps/ folder npm ci will have to rerun everytime you change something in apps

docker/supervisord.conf
Bl3f: Because we can't for now as we moved some bindings (sqlite) to bun for the standalone server

apps/backend/src/cli.ts
Bl3f: that's true will refactor

.github/workflows/docker.yml
Bl3f: so we can have on PR a docker image build that we could eventually deploy to have a live check of contributors PRs

docker/supervisord.conf
Bl3f: rps vs stability lol https://sharkbench.dev/web/javascript-fastify/bun-vs-nodejs

PR #62, opened 2026-01-25
Remove unused repositories and databases directories during nao sync

### LINKED ISSUE #41: nao sync after changing the nao_config.yaml only adds the new file but doesn’t remove the ones deleted
ex if you removed a table or repo from the config it stays in the folder tree if it had already been synced

paulsojan:
Fixes #41

The issue with the current stale database cleanup logic is that when multiple databases share the same type, removing a single database of that type does not work correctly. The existing implementation only checks whether the database type exists in the configuration, but it does not verify whether a specific database under that type still exists. Refactored the current logic to include that as well.

Demo: https://paulsojan.neetorecord.com/watch/afae298321837c18a47f

### DIFF
--- a/cli/tests/nao_core/commands/sync/test_cleanup.py
+++ b/cli/tests/nao_core/commands/sync/test_cleanup.py
@@ -21,9 +21,7 @@ class DBConfig:
     database: str | None = None
 
     def get_database_name(self) -> str:
-        """Extract database name from path or return database attribute."""
         if self.path:
-            # Extract filename without extension from path
             return Path(self.path).stem
         if self.database:
             return self.database


cli/nao_core/commands/sync/__init__.py
```diff
@@ -48,13 +51,19 @@ def sync(
...
-
         # Get output directory (custom or default)
         output_dir = output_dirs.get(provider.name, provider.default_output_dir)
         output_path = Path(output_dir)
 
+        if isinstance(provider, DatabaseSyncProvider):
+            cleanup_stale_databases(config.databases, output_path, verbose=True)
+
+        if isinstance(provider, RepositorySyncProvider):
+            cleanup_stale_repos(config.repos, output_path, verbose=True)
```
paulsojan: The reason cleanup methods are implemented outside of the respective provider files is that when the configuration becomes empty and the repositories or database folders contain files, the cleanup function inside the provider file will not be executed when we run the `nao sync` command.

https://paulsojan.neetorecord.com/watch/83b73549e358b35aff43

Bl3f: That's a great catch, tho in term of design we wanted to create this provider abstraction so the sync/__init__.py does not have to implement if/else logic depending on database. Would be awesome if we can find a code structure that is more generic in this file to avoid the isinstance(provider) this way it gonna be more agnostic.

I say this because in the long term vision we want to be able to give this provider system as an extensible system so people can develop provider to pull system context in the filesystem.

cli/nao_core/commands/sync/cleanup.py
```diff
@@ -105,29 +107,69 @@ def cleanup_stale_paths(state: DatabaseSyncState, verbose: bool = False) -> int:
...
+                shutil.rmtree(db_dir)
                 if verbose:
-                    console.print(f"  [dim red]removing stale database type:[/dim red] {db_type_dir.name}")
-                shutil.rmtree(db_type_dir)
-                removed_count += 1
+                    console.print(f"\n[yellow] Removed unused database:[/yellow] {type_folder_name}/{db_dir.name}")
 
-    return removed_count
+
+def get_database_identifier(db):
```
Bl3f: Can't we move this to each database provider?

cli/nao_core/commands/sync/cleanup.py
```diff
@@ -105,29 +107,69 @@ def cleanup_stale_paths(state: DatabaseSyncState, verbose: bool = False) -> int:
...
+    if db.type == "bigquery":
+        return db.project_id
+
+    elif db.type == "duckdb":
+        if db.path == ":memory:":
+            return "memory"
+        return Path(db.path).stem
```
Bl3f: should we return this or the database name inside the instance instead?

paulsojan: This is the existing logic. This is how the database name for type `duckdb` is generated. I haven't made any additional changes here.

cli/nao_core/commands/sync/cleanup.py
paulsojan: Yes, refactored the logic

PR #82, opened 2026-01-28
Charts: date range filtering, hide/show series, improve error and loa…

Standlc:
Charts:
- add date filtering on chart data (7d, 30d, 3m, etc)
- toggle series visibility (choose which series to draw)
- show skeleton when loading and improve error state UI

Todo:
- better handle date filtering:
  - make date filtering options dynamic based on graph data
  - don't filter/show filtering options if not enough data (and prevent to show only one x tick)

https://github.com/user-attachments/assets/fd46c223-b7dd-46c5-b50a-f66805e56cd6

https://github.com/user-attachments/assets/4a5f99d8-55cc-4bc9-9141-ff1a18080274


apps/frontend/src/lib/charts.utils.ts
```diff
@@ -1,6 +1,87 @@
...
+	});
+}
+
+function isValidDate(date: Date): boolean {
+	return !isNaN(date.getTime());
+}
+
+/** Checks if a string is in ISO 8601 date format (e.g., 2024-01-15 or 2024-01-15T12:30:00Z) */
+function isISODateString(value: string): boolean {
+	return /^\d{4}-\d{2}-\d{2}(T\d{2}:\d{2}:\d{2}(\.\d{3})?(Z|[+-]\d{2}:\d{2})?)?$/.test(value);
```
Bl3f, line 70: isn't better to parse the date and check if it works rather than a regexp that can validate dates that makes no sense like "2026-02-30", this way you merge isValidDate and isIsoDateString

Standlc: `Date` parse "01/12/2025" as "Jan 12 2025" because it interprets the US pattern of `MM/DD/YYYY` and not `DD/MM/YYYY`. Because of these cases, i'd rather only parse ISO which is stable and ambiguous

PR #84, opened 2026-01-28
fix: error when BETTER_AUTH_URL not defined + validate env variables before server starts

SIMOUNIX:
- was getting an error because the BETTER_AUTH_URL was not defined properly and was hard to catch among the server logs -> do not start if everything is properly set up
- was calling backend before the sessions was properly initialized (getAvailableModels was called when user not yet connected = no session)
- was not able to build the server because the bun compiled check was not good

Screenshot from 2026-01-28 22-57-41

could not get to catch the error and display a proper message using `console.print`, I will try to make it better

### DIFF
--- a/apps/backend/src/env.ts
+++ b/apps/backend/src/env.ts
@@ -11,7 +11,7 @@ const envSchema = z.object({
 		.enum(['true', 'false'])
 		.optional()
 		.transform((val) => val === 'true'),
-	BETTER_AUTH_SECRET: z.string().min(1).or(z.literal('').optional()), // try to make min 1 if set and optional otherwise
+	BETTER_AUTH_SECRET: z.string().min(20).or(z.literal('').optional()), // try to make min 1 if set and optional otherwise
 	TRUSTED_ORIGINS: z
 		.string()
 		.optional()


apps/frontend/src/routes/__root.tsx
```diff
@@ -10,7 +10,7 @@ function RootComponent() {
 	const session = useSessionOrNavigateToLoginPage();
 	useDisposeInactiveAgents();
 
-	if (session.isPending) {
+	if (session.isPending || !session.data) {
```
Standlc: Why? If i'm not mistaken that means we're not rendering anything if the user isn't connected. How do they get to the login page?

apps/backend/src/cli.ts
```diff
@@ -27,11 +27,12 @@ interface BuildInfo {
 
 function getExecutableDir(): string {
 	// Check if running as compiled binary or as script
-	// When compiled, process.execPath === Bun.main (both point to the binary)
-	// When running as script, process.execPath is bun, Bun.main is the script
-	const isCompiled = process.execPath === Bun.main;
+	// When compiled with Bun, Bun.main points to /$bunfs/... (internal virtual filesystem)
+	// When running as script, Bun.main points to the actual .ts file path
+	const isCompiled = Bun.main.startsWith('/$bunfs/');
```
Standlc, line 38: @Bl3f what do you think?

apps/frontend/src/routes/__root.tsx
SIMOUNIX: This resolves the bug where we were trying to fetch availlableModels before the user is connected! Maybe we can guard it in the children components.

apps/backend/src/cli.ts
Bl3f: Yes actually it was a bug i introduced while working on Docker and did not notice at that time. I've fixed it in another PR, so it can be reverted.

apps/backend/src/env.ts
```diff
@@ -0,0 +1,38 @@
...
+	DB_URI: z.string().default('sqlite:./db.sqlite'),
+	REDIRECT_URL: z.url({ message: 'REDIRECT_URL must be a valid URL' }).default('http://localhost:3000/'),
+	BETTER_AUTH_URL: z.url({ message: 'BETTER_AUTH_URL must be a valid URL' }),
+
+	// Optional
+	DB_SSL: z
+		.enum(['true', 'false'])
+		.optional()
+		.transform((val) => val === 'true'),
+	BETTER_AUTH_SECRET: z.string().min(1).or(z.literal('').optional()), // try to make min 1 if set and optional otherwise
```
Bl3f: Maybe the min should be a bit higher if we want to consider it a secret?

apps/backend/src/env.ts
```diff
@@ -0,0 +1,38 @@
...
+	TRUSTED_ORIGINS: z
```
Bl3f, line 15: There is something to factorise with the URLs because BETTER_AUTH_URL, TRUSTED_ORIGINS and REDIRECT_URL are actually kinda the same thing

BETTER_AUTH_URL and TRUSTED I think are used to validated the origin of login, and REDIRECT_URL is when sending a Slack message to get the url it's deployed on.

But actually 3 urls should be the same most of the time so, if working on this I'd in favour of converging to 1 url.

Can be done in another PR so we merge this one and separte concer tho. I'm opening a ticket for later.

Bl3f: cf. #96

SIMOUNIX: alright I will look into this in a new PR to make things less confusing

Issue #81, opened 2026-01-28
Support Gemini API keys


ClaireGz:
As a user, I want to use Gemini as an LLM provider

PR #85, opened 2026-01-28
Add Gemini and Mistral + small typing refacto

Bl3f:
* Added Gemini and Mistral
* Small types refactor
* Added their icons

Completes #81

### DIFF
--- a/apps/backend/src/queries/project-llm-config.queries.ts
+++ b/apps/backend/src/queries/project-llm-config.queries.ts
@@ -3,7 +3,7 @@ import { and, eq } from 'drizzle-orm';
 import s, { DBProjectLlmConfig, NewProjectLlmConfig } from '../db/abstractSchema';
 import { db } from '../db/db';
 import { LlmProvider } from '../types/llm';
-import { getDefaultEnvProvider, getDefaultModelId, getEnvModelSelections } from '../utils/llm';
+import { getDefaultEnvProvider, getDefaultModelId } from '../utils/llm';
 
 export const getProjectLlmConfigs = async (projectId: string): Promise<DBProjectLlmConfig[]> => {
 	return db.select().from(s.projectLlmConfig).where(eq(s.projectLlmConfig.projectId, projectId)).execute();


apps/backend/src/queries/project-llm-config.queries.ts
```diff
@@ -115,12 +115,8 @@ export const getProjectAvailableModels = async (
 	}
 
 	// Also add env-configured providers with their defaults
-	const envProviders: LlmProvider[] = ['anthropic', 'openai'];
-	for (const provider of envProviders) {
-		if (hasEnvApiKey(provider) && !configs.some((c) => c.provider === provider)) {
-			models.push({ provider, modelId: getDefaultModelId(provider) });
-		}
-	}
+	const envSelections = getEnvModelSelections().filter((s) => !configs.some((c) => c.provider === s.provider));
```
Standlc: `getEnvModelSelections` is not a db query, shouldn't be here.

apps/frontend/src/components/settings-llm-provider-card.tsx
```diff
@@ -35,6 +36,7 @@ export function ProviderCard({
 			<div className='flex items-center gap-4'>
 				<div className='flex-1 grid gap-1'>
 					<div className='flex items-center gap-2'>
+						<ProviderIcon provider={provider} className='size-3.5' />
```
Standlc: `ProviderIcon` is very generic, make it `LlmProviderIcon` maybe

apps/backend/src/types/llm.ts
```diff
@@ -49,3 +109,14 @@ export function getDefaultModelId(provider: LlmProvider): string {
 	const defaultModel = models.find((m) => m.default);
 	return defaultModel?.id ?? models[0].id;
 }
+
+export function getProviderModelConfig<P extends LlmProvider>(provider: P, modelId: string): ProviderConfigMap[P] {
```
Standlc: move to utils/llm.ts

apps/backend/src/types/llm.ts
```diff
@@ -13,14 +17,35 @@ export const llmConfigSchema = z.object({
...
+
 /** Provider configuration with env var names and known models */
-export const LLM_PROVIDERS: Record<
-	LlmProvider,
-	{
-		envVar: string;
-		models: readonly { id: string; name: string; default?: boolean }[];
-	}
-> = {
+export const LLM_PROVIDERS: LlmProvidersType = {
```
Standlc: that's not a type, move to `utils/llm.ts` or `agents/providers.ts`

apps/backend/src/services/agent.service.ts
Standlc: Lots of repetition. All the clients are created with the same pattern. Can you factorize better with an extra method that takes a client function (`createAnthropic`, `createGoogleGenerativeAI`, etc) and an LLM config as parameters?

apps/backend/src/types/llm.ts
Bl3f: Moving to  `agents/providers.ts` then, i dont think it belongs to `utils` because it's something static and it's more logical i feel to put a static content in `types` than in `utils`. `agents/providers` works ok even if i'd prefer in types even if it's not exactly types.

apps/backend/src/services/agent.service.ts
Bl3f: Ok did something. Tbf having multiple ifs like this made it way easier to read, but i think i found a pattern that still is easy to read without being a pain in this ass

PR #88, opened 2026-01-29
Refactor `Create User` component

paulsojan:
- I have replaced normal form with tanstack form. Also added form validation.
- Moved hardcoded strings to constants.
- Kept the create user component under the `settings` folder.  The remaining settings related components can be added to this folder.

TODO
- Currently, we need to explicitly handle the backend error for each api, instead we can create a global error handling mechanism in the frontend to catch and display api errors.

### DIFF
--- a/apps/frontend/package.json
+++ b/apps/frontend/package.json
@@ -39,7 +39,6 @@
 		"proxy-from-env": "^1.1.0",
 		"react": "^19.2.0",
 		"react-dom": "^19.2.0",
-		"react-hook-form": "^7.71.1",
 		"recharts": "2.15.4",
 		"streamdown": "^2.0.1",
 		"superjson": "^2.2.6",
--- a/package-lock.json
+++ b/package-lock.json
@@ -340,7 +340,6 @@
 				"proxy-from-env": "^1.1.0",
 				"react": "^19.2.0",
 				"react-dom": "^19.2.0",
-				"react-hook-form": "^7.71.1",
 				"recharts": "2.15.4",
 				"streamdown": "^2.0.1",
 				"superjson": "^2.2.6",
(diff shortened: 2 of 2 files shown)


apps/frontend/src/components/settings/create-user/create-user.tsx
paulsojan, line 1: I have created this folder structure for better separation of concerns and kept all logic related to that component within its closest scope. Let me know your thoughts.

MatLBS:
Could you send some screenshots in order to see the changes plz

apps/frontend/src/components/settings/create-user/constants.ts
```diff
@@ -0,0 +1,17 @@
+import * as z from 'zod';
+
+export const USER_VALIDATION_SCHEMA = z.object({
```
MatLBS, line 3: No need for schema validation here. 

We already have this in the user.routes.ts : 

```
createUserAndAddToProject: adminProtectedProcedure
		.input(
			z.object({
				name: z.string().min(1),
				email: z.string().min(1),
			}),
		)
```

apps/frontend/src/components/settings/create-user/constants.ts
```diff
@@ -0,0 +1,17 @@
...
+	name: z.string().min(2, 'Name must be at least 2 characters'),
+	email: z.email('Invalid email address'),
+});
+
+export const USER_FORM_INITIAL_VALUES = {
```
MatLBS, line 8: In my upcoming PR, I moved everything in a shared context between user page components

apps/frontend/src/components/settings/create-user/constants.ts
paulsojan: This schema is for frontend form validation. So it will only allow the user to submit a valid form.

Issue #71, opened 2026-01-27
We should add an organisation level in the application


Bl3f:
# Problem

Currently we you launch nao it start attached to the default project given and gives the admin to the first user to log in this project. To prepare for the multi project setup we should create an organisation level with the corresponding permissions which would make sense to attach configurations at org level like the SSO, the app name, logo etc.

PR #93, opened 2026-01-30
feat: added organization level

SIMOUNIX:
start implementation of #71 

[Screencast from 2026-01-30 11-17-12.webm](https://github.com/user-attachments/assets/bc9d2030-b1f9-4706-b4f1-4a84253881f0)

### DIFF
--- a/apps/backend/src/index.ts
+++ b/apps/backend/src/index.ts
@@ -1,24 +1,9 @@
 import './env';
 
 import { startServer } from './app';
-import app from './app';
-import { ensureOrganizationSetup } from './queries/organization.queries';
 
 startServer({ port: 5005, host: '0.0.0.0' }).catch((err) => {
 	console.error('\n❌ Server failed to start:\n');
 	console.error(`   ${err.message}\n`);
 	process.exit(1);
 });
-
-ensureOrganizationSetup()
-	.then(() => {
-		return app.listen({ host: '0.0.0.0', port: 5005 });
-	})
-	.then((address) => {
-		console.log(`Server is running on ${address}`);
-	})
-	.catch((err) => {
-		console.error('\n❌ Server failed to start:\n');
-		console.error(`   ${err.message}\n`);
-		process.exit(1);
-	});


apps/backend/src/types/organization.ts
```diff
@@ -0,0 +1,3 @@
+export type OrgRole = 'org_admin' | 'org_user';
```
MatLBS: I think we should keep 'admin' and 'user'

apps/backend/src/db/pgSchema.ts
```diff
@@ -88,12 +89,47 @@ export const verification = pgTable(
...
+export const organization = pgTable('organization', {
+	id: text('id')
+		.$defaultFn(() => crypto.randomUUID())
+		.primaryKey(),
+	name: text('name').notNull(),
+	slug: text('slug').unique(), // for future multi-org URL routing
+	logo: text('logo'),
+	appName: text('app_name'),
+	// SSO config
+	googleClientId: text('google_client_id'),
```
MatLBS, line 100: We need to discuss if the SSO config is in the org scope or project scope. 

Maybe you want to allow SSO in one project but not another

apps/backend/src/db/sqliteSchema.ts
```diff
@@ -89,12 +90,52 @@ export const verification = sqliteTable(
 	(table) => [index('verification_identifier_idx').on(table.identifier)],
 );
 
+export const organization = sqliteTable('organization', {
+	id: text('id')
+		.$defaultFn(() => crypto.randomUUID())
+		.primaryKey(),
+	name: text('name').notNull(),
+	slug: text('slug').unique(), // for future multi-org URL routing
```
MatLBS: What is the purpose of the column 'slug' ? 

We want to implement multi-project in an organization but not sure we want to implement multi-organization

apps/backend/src/db/sqliteSchema.ts
```diff
@@ -89,12 +90,52 @@ export const verification = sqliteTable(
...
+	logo: text('logo'),
+	appName: text('app_name'),
```
MatLBS: What is the difference between 'name' and 'appName' ?

apps/backend/src/auth.ts
```diff
@@ -34,8 +35,11 @@ export const auth = betterAuth({
 					return true;
 				},
 				async after(user) {
-					// Handle first user signup: create default project and add user as admin
-					await projectQueries.initializeDefaultProjectForFirstUser(user.id);
+					// Handle first user signup: create default org, then default project
+					const org = await orgQueries.initializeDefaultOrganizationForFirstUser(user.id);
```
MatLBS: You should create the default project in the same function as the organization. 

Implement a db transaction in order to create both

apps/backend/src/db/pgSchema.ts
Bl3f: No I think the SSO is at org level tbf

apps/backend/src/queries/organization.queries.ts
```diff
@@ -0,0 +1,141 @@
...
+		name: 'Default Organization',
+	});
+
+	await addOrgMember({
+		orgId: org.id,
+		userId,
+		role: 'org_admin',
+	});
+
+	return org;
```
MatLBS: If you create also the default project in this function, no need to return 'org'

apps/backend/src/db/sqliteSchema.ts
SIMOUNIX: honestly none

apps/backend/src/db/sqliteSchema.ts
SIMOUNIX: misread the issue.. reverted it

SIMOUNIX:
> Hi, thanks for your PR !
> 
> I left some comments.
> 
> We need to discuss and decide how we are gonna handle organizations and projects.
> 
> At the same time I also pushed a PR with a db migration and I would like to merge it before diving on the organization level.

no problem, tell me more when your work is merged I will rebase and eventually work around what you did! 

also I just pushed some new stuff!

MatLBS (changes requested):
@SIMOUNIX careful, your `organization tables` are not the same between sqlite and pg. 

Keep `slug` and `name` but remove`logo` and `appName`

Recreate your migration and everything will be good to merge

PR #98, opened 2026-01-30
Feat/improve signin flow

### LINKED ISSUE #51: Work on the signup flow
The signup flow should be better than what it is today.

# Current situation

There are 2 ways to signup in nao:
* Using the signup form
* Using the Google SSO if `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` are setup

As we are moving towards self-hosted deployment people will deploy nao in their infrastructure, as per Slack requirement it means nao have to be exposed to the outside so Slack can reach the backend. Because of this it means nao application could be accessible online, any random person online could signup to a nao instance.

We don't want this.

# What we should do:
1. Signup form should be disabled, only the first user can signup and become admin of the `NAO_DEFAULT_PROJECT_FOLDER` setup
2. When using the SSO the setup admin should pass an env variable named GOOGLE_DOMAINS_ALLOW_LIST
3. Project admin should be able to add users in the admin

## Signup form should be disabled

As a project admin, if the user count is currently 0 i should be proposed a signup form with username, email, password the first time i come to the login page (no login should be proposed then) — it should also offer the Google login (if activated)
As a user, if the user count is > 0 it should propose login form or Google login (if activated)

## GOOGLE_DOMAINS_ALLOW_LIST

Should be a comma separated list of domains authorised to login through Google SSO. This variable has to be taken into account everytime

if GOOGLE_DOMAINS_ALLOW_LIST='' then no domain will be allowed

other example
GOOGLE_DOMAINS_ALLOW_LIST=getnao
GOOGLE_DOMAINS_ALLOW_LIST=getnao,heynao

## Project admin member page

in the /user page in the project section there should be a member management section where you can add member to project

### Add a member

As an admin i can add a user in the project, the username and the email have to be filled, the first password as to be randomly generated
As an user who got credentials i can connect to nao using the login form, if it's my first connection i should be prompted to change my password (old password, new password)

MatLBS:
Fix #51 

Integrated new feature : 

- Created a context to handle the user interactions

- Admin can generate a new password for a user
- Admin can modify the role of any user. Must be at least one admin per project
- Admin can remove a user from a project
- Implemented a SMTP server with 3 credentials : EMAIL_SERVICE (default to gmail), GOOGLE_APP_PASSWORD and EMAIL_USER
-  When adding a user to the project, check if he already exists
- When user first login or its password has been reset, need to modify password

### DIFF
--- a/apps/frontend/src/components/settings-modify-user-form.tsx
+++ b/apps/frontend/src/components/settings-modify-user-form.tsx
@@ -1,5 +1,6 @@
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { useForm } from '@tanstack/react-form';
+import { useEffect } from 'react';
 import { ChevronDown } from 'lucide-react';
 import { DropdownMenu, DropdownMenuItem, DropdownMenuContent, DropdownMenuTrigger } from './ui/dropdown-menu';
 import { trpc } from '@/main';
@@ -17,6 +18,7 @@ export function ModifyUserForm({ isAdmin }: ModifyUserInfoProps) {
 	const { userInfo, isModifyUserFormOpen, setIsModifyUserFormOpen, error, setError } = useUserPageContext();
 	const { refetch } = useSession();
 	const queryClient = useQueryClient();
+
 	const modifyUser = useMutation(
 		trpc.user.modify.mutationOptions({
 			onSuccess: async () => {
@@ -34,7 +36,7 @@ export function ModifyUserForm({ isAdmin }: ModifyUserInfoProps) {
 
 	const form = useForm({
 		defaultValues: {
-			name: userInfo.name || '',
+			name: userInfo.name,
 			role: userInfo.role,
 		},
 		onSubmit: async ({ value }) => {
@@ -47,6 +49,14 @@ export function ModifyUserForm({ isAdmin }: ModifyUserInfoProps) {
 			});
 		},
 	});
+
+	useEffect(() => {
+		if (isModifyUserFormOpen) {
+			form.reset();
+			form.setFieldValue('name', userInfo.name);
+			form.setFieldValue('role', userInfo.role);
+		}
+	}, [isModifyUserFormOpen, userInfo, form]);
 
 	return (
 		<Dialog open={isModifyUserFormOpen} onOpenChange={setIsModifyUserFormOpen}>


apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+
+	constructor() {
+		this.initialize();
+	}
+
+	private initialize() {
+		const { EMAIL_SERVICE, EMAIL_USER, GOOGLE_APP_PASSWORD } = process.env;
+
+		if (!EMAIL_USER || !GOOGLE_APP_PASSWORD) {
+			return;
```
Standlc, line 21: Should we throw an error in this case or is the email service not a mandatory service? In other words, can the backend work if the mail service doesn't?

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
+import type { Transporter } from 'nodemailer';
+import nodemailer from 'nodemailer';
+
+import type { CreatedEmailData, EmailData, SendEmailParams } from '../types/email';
+
+class EmailService {
+	private transporter: Transporter | null = null;
```
Standlc: make it `undefined` by default. We don't really use null in the codebase

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+			return await this.createUserAddedToProjectEmail(data);
+		}
+	}
+
+	public async sendEmail(params: SendEmailParams): Promise<void> {
+		if (!this.enabled || !this.transporter) {
+			return;
+		}
+
+		const { user, projectName, temporaryPassword, type } = params;
```
Standlc: you can also spread the params object above instead of here.
```typescript
public async sendEmail({ user, projectName, temporaryPassword, type }: SendEmailParams): Promise<void> {
```

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+				},
+			});
+
+			this.enabled = true;
+		} catch {
+			this.enabled = false;
+		}
+	}
+
+	async createUserAddedToProjectEmail(data: EmailData): Promise<CreatedEmailData> {
```
Standlc: this function doesn't have to be `async`

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+		try {
+			await this.transporter.sendMail({
+				from: process.env.EMAIL_USER,
+				to: user.email,
+				subject: email.subject,
+				html: email.html,
+				text: email.text,
+			});
+		} catch (error) {
+			console.error(`❌ Failed to send email to ${user.email}:`, error);
```
Standlc: remove the try and catch and let the client of the service handle the error (to log or send in the response body)

apps/backend/src/services/email.service.ts
Standlc, line 1: Generally pretty good! Some improvement tho:
- add the `private` keyword to private methods
- sort your methods by placing the public ones first, then the ones used in order of usage. That will improve readability
- you could use `renderToString` from `react-dom/server` to render the html with React. Optional but html as text is really sketchy here

apps/frontend/src/routes/modifyPassword.tsx
```diff
@@ -0,0 +1,87 @@
...
+	const navigate = useNavigate();
+	const { data: sessionData, refetch } = useSession();
+	const [newPassword, setNewPassword] = useState('');
+	const [confirmPassword, setConfirmPassword] = useState('');
+	const [error, setError] = useState('');
+
+	const modifyUserPassword = useMutation(
+		trpc.account.modifyPassword.mutationOptions({
+			onSuccess: async () => {
+				await refetch();
```
Standlc: why do `refetch` here? The password is not loaded by the frontend anyway

apps/frontend/src/routes/modifyPassword.tsx
```diff
@@ -0,0 +1,87 @@
+import { createFileRoute, useNavigate } from '@tanstack/react-router';
+import { useState } from 'react';
+import { useMutation } from '@tanstack/react-query';
+import { useSession } from '@/lib/auth-client';
+import { trpc } from '@/main';
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+
+export const Route = createFileRoute('/modifyPassword')({
```
Standlc: Not sure this page should be a route. Instead, can we show this component programmatically based on the user state that the backend sends us? Otherwise the user could skip this page.
However, if you want to keep it as a page, rename it to `modify-password`.

.env.example
```diff
@@ -27,6 +27,12 @@ REDIRECT_URL='http://localhost:3000/'
 # Context folder (to change for your setup)
 NAO_DEFAULT_PROJECT_PATH=/Users/blef/Work/naolabs/chat/example
 
+# FastAPI URL
 FASTAPI_URL=http://localhost:8005
 
+# SMTP server Configuration
+EMAIL_SERVICE=                  # Email service: gmail, outlook, yahoo, etc. (defaults to gmail)
+GOOGLE_APP_PASSWORD=            # Your app password (for Gmail: https://myaccount.google.com/apppasswords)
+EMAIL_USER=                     # Sender address (e.g., your email)
```
Standlc: Why do you need a user email? The emails should be sent to the email of any user, not the one you setup in your env.

apps/backend/src/trpc/account.routes.ts
```diff
@@ -0,0 +1,76 @@
...
+				throw new TRPCError({
+					code: 'NOT_FOUND',
+					message: 'User account not found or user does not use password authentication.',
+				});
+			}
+
+			const password = crypto.randomUUID().slice(0, 8);
+			const hashedPassword = await hashPassword(password);
+
+			await accountQueries.updateAccountPassword(account.id, hashedPassword, input.userId);
```
Standlc, line 40: you need to check that the user doing the password reset is allowed to do it, otherwise an admin could reset the password of a user from a different project

apps/backend/src/trpc/account.routes.ts
```diff
@@ -0,0 +1,76 @@
...
+		.mutation(async ({ input }) => {
+			const account = await accountQueries.getAccountById(input.userId);
+			if (!account || !account.password) {
+				throw new TRPCError({
+					code: 'NOT_FOUND',
+					message: 'User account not found or user does not use password authentication.',
+				});
+			}
+
+			if (input.newPassword !== input.confirmPassword) {
```
Standlc, line 72: actually I think only the frontend should check that and only send `newPassword`. Let's not waste server resources when the frontend can do it.

apps/backend/src/trpc/account.routes.ts
```diff
@@ -0,0 +1,76 @@
...
+
+			if (!regexPassword.test(input.newPassword)) {
+				throw new TRPCError({
+					code: 'BAD_REQUEST',
+					message:
+						'New password must be at least 8 characters long and include uppercase, lowercase, number, and special character.',
+				});
+			}
+
+			const hashedPassword = await hashPassword(input.newPassword);
```
Standlc, line 87: the pattern of checking a password's validity and hashing it is very redundant. You should create a function or class that handles password creation/validation.

apps/backend/src/trpc/user.routes.ts
```diff
@@ -23,48 +22,36 @@ export const userRoutes = {
...
-			if (input.previousPassword && input.newPassword) {
-				if (!regexPassword.test(input.newPassword)) {
-					throw new TRPCError({
-						code: 'BAD_REQUEST',
-						message:
-							'New password must be at least 8 characters long and include uppercase, lowercase, number, and special character.',
-					});
-				}
+		.mutation(async ({ input, ctx }) => {
+			const previousRole = await projectQueries.getUserRoleInProject(ctx.project!.id, input.userId);
```
Standlc: you shouldn't have to do `ctx.project!.id`. The project should not be potentially undefined, and if it is, that means `projectProtectedProcedure` doesn't narrow down the type of a project

apps/backend/src/trpc/user.routes.ts
```diff
@@ -23,48 +22,36 @@ export const userRoutes = {
...
+			const previousName = (await userQueries.get({ id: input.userId }))?.name;
```
Standlc: move that line down in the function, you don't need it yet (you're throwing potential `TRPCError` before using `previousName`). Also, make it a specific query `userQueries.getName()`

apps/backend/src/trpc/user.routes.ts
```diff
@@ -98,11 +84,51 @@ export const userRoutes = {
...
+		.input(
+			z.object({
+				email: z.string().min(1),
+			}),
+		)
+		.mutation(async ({ input, ctx }) => {
+			const user = await userQueries.get({ email: input.email });
+			if (!user) {
+				return {
+					success: false,
```
Standlc: throw 404 instead?

apps/backend/src/trpc/user.routes.ts
```diff
@@ -98,11 +84,51 @@ export const userRoutes = {
...
+				role: 'user',
+			});
+
+			await emailService.sendEmail({
+				user,
+				type: 'createUser',
+				projectName: ctx.project?.name,
+			});
+
+			return { success: true };
```
Standlc: returning `success` seems useless here. Any reason to keep it?

apps/frontend/src/components/settings-add-user-form.tsx
```diff
@@ -0,0 +1,119 @@
...
+			onError: (err) => {
+				setError(err.message);
+			},
+		}),
+	);
+
+	const createUserAndAddToProject = useMutation(
+		trpc.user.createUserAndAddToProject.mutationOptions({
+			onSuccess: async (ctx) => {
+				await queryClient.invalidateQueries({
```
Standlc: it's pretty bad to invalidate the queries every time you do a mutation. That means we're refetching the data when we could simply update it locally. use `ctx.client.setQueryData()`

apps/frontend/src/components/settings-add-user-form.tsx
```diff
@@ -0,0 +1,119 @@
...
+				setError(err.message);
+			},
+		}),
+	);
+
+	const handleSubmit = async () => {
+		setError('');
+
+		if (needToCreateUser === false) {
+			await searchUserAndAddToProject.mutateAsync({ email });
```
Standlc: useless to do `mutateAsync` here. call `mutate`

apps/frontend/src/components/settings-add-user-form.tsx
```diff
@@ -0,0 +1,119 @@
...
+	const handleClose = () => {
+		setIsAddUserFormOpen(false);
+		setNeedToCreateUser(false);
+		setEmail('');
+		setName('');
+	};
+
+	const searchUserAndAddToProject = useMutation(
+		trpc.user.searchUserAndAddToProject.mutationOptions({
+			onSuccess: async (ctx) => {
```
Standlc: the first parameters here is not `ctx`, it's the data the backend sends.
Overall it doesn't seem like a clean logic.

apps/frontend/src/hooks/useEnforcePasswordReset.tsx
```diff
@@ -0,0 +1,16 @@
...
+	const navigate = useNavigate();
+	const session = useSession();
+
+	useEffect(() => {
+		if (!session.isPending && session.data?.user?.requiresPasswordReset) {
+			navigate({ to: '/modifyPassword' });
+		}
+	}, [session.isPending, session.data?.user?.requiresPasswordReset, navigate]);
+
+	return session;
```
Standlc: don't return the `session` here. If your hook is called `useEnforcePasswordReset`, then it shouldn't return a session. A better name is `useNavigateToResetPageIfNeeded`.

apps/frontend/src/hooks/useEnforcePasswordReset.tsx
Standlc, line 1: use kebab case

apps/frontend/src/contexts/user.provider.tsx
```diff
@@ -0,0 +1,73 @@
+import { createContext, useContext, useState } from 'react';
+import type { ReactNode } from 'react';
+import type { UserWithRole } from '../../../backend/src/types/project';
+
+interface UserPageContextType {
```
Standlc, line 5: that's way too much data passed in the context. This indicates you're doing waaaay too many things in the same context and that separation of concerns is weak.

.env.example
Bl3f: I add on this that this is not usual SMTP setup here.

usual SMTP looks like this:

```
[smtp]
smtp_host=smtp.sendgrid.net
smtp_starttls=False
smtp_ssl=False
smtp_port=587
smtp_mail_from=<your-from-email>```

File: apps/backend/src/services/email.service.ts (hunk 1 above)
Comment (Bl3f) [reply apps/backend/src/services/email.service.ts:21]: IMHO, email service should not be mandatory

Comment (MatLBS) [reply apps/backend/src/services/email.service.ts:21]: That is what I thought

File: .env.example (hunk 8 above)
Comment (MatLBS) [reply .env.example]: Well there is two types of setup and this one is easier : 

EMAIL_SERVICE=                  
GOOGLE_APP_PASSWORD=      
EMAIL_USER=

File: apps/frontend/src/routes/modifyPassword.tsx (hunk 7 above)
Comment (MatLBS) [reply apps/frontend/src/routes/modifyPassword.tsx]: Normally the user cannot skip the page as I enforce it in `useEnforcePasswordReset`

I already tried to rename it to '/modify-password' but had this error : 

`Argument of type '"/modify-password"' is not assignable to parameter of type 'keyof FileRoutesByPath | undefined'.`

File: apps/backend/src/db/abstractSchema.ts [hunk 21]
```diff
@@ -31,4 +31,6 @@ export type NewProjectMember = typeof sqliteSchema.projectMember.$inferInsert;
 export type DBProjectLlmConfig = typeof sqliteSchema.projectLlmConfig.$inferSelect;
 export type NewProjectLlmConfig = typeof sqliteSchema.projectLlmConfig.$inferInsert;
 
+export type UpdateUser = Partial<Omit<NewUser, 'id' | 'createdAt'>> & { id: string };
```
Bl3f: Weird that we have to omit id to add it back no?

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+class EmailService {
+	private transporter: Transporter | null = null;
+	private enabled: boolean = false;
+
+	constructor() {
+		this.initialize();
+	}
+
+	private initialize() {
+		const { EMAIL_SERVICE, EMAIL_USER, GOOGLE_APP_PASSWORD } = process.env;
```
Bl3f: As said, variables names should be consistent, GOOGLE_APP_PASSWORD feels like an impostor here.

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+
+		if (!EMAIL_USER || !GOOGLE_APP_PASSWORD) {
+			return;
+		}
+
+		try {
+			this.transporter = nodemailer.createTransport({
```
Bl3f, line 25: here it should at least support SMTP, but as I thought about it should also work with Gmail the way you implemented it.

apps/backend/src/services/email.service.ts
```diff
@@ -0,0 +1,258 @@
...
+		const { to, userName, projectName, temporaryPassword, loginUrl } = data;
+
+		const isNewUser = !!temporaryPassword;
+
+		const subject = isNewUser
+			? `You've been invited to ${projectName} on nao`
+			: `You've been added to ${projectName} on nao`;
+
+		const html = `
+<!DOCTYPE html>
```
Bl3f: Can't we get these email in JSX/TSX so it's easier to maintain? AS you have multiple time the header etc. Would be better if the parts that can be factorised can be. And if we can use some JSX or templating thing would be even better.

apps/backend/src/trpc/account.routes.ts
Bl3f: Not sure it wastes a lot of ressources, tho I agree in term of experience would be cool if the frontend does the check first. But I'd also keep it in the backend as a safe (it's an API can be called from elsewhere, we never know).

apps/backend/src/trpc/account.routes.ts
Bl3f: Not sure what can be factorised here? it's already a function?

Bl3f (commented):
few comments on backend side

PR #99, opened 2026-01-30
Fix the nao_config loading so it displays errors in the CLI

### LINKED ISSUE #97: When nao_project.yaml format is wrong, says no nao_project.yaml exist
As a user, if I broke the format of the nao_project.yaml and then launches a nao debug or nao sync command, it says no nao_project.yaml exist.

Image

Bl3f:
Fixes #97

### DIFF
--- a/cli/nao_core/commands/sync/__init__.py
+++ b/cli/nao_core/commands/sync/__init__.py
@@ -1,6 +1,5 @@
 """Sync command for synchronizing repositories and database schemas."""
 
-import sys
 from pathlib import Path
 
 from rich.console import Console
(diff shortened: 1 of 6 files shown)


cli/nao_core/commands/sync/__init__.py
```diff
@@ -38,12 +37,8 @@ def sync(
...
     console.print("\n[bold cyan]🔄 nao sync[/bold cyan]\n")
 
-    config = NaoConfig.try_load()
-    if config is None:
-        console.print("[bold red]✗[/bold red] No nao_config.yaml found in current directory")
-        console.print("[dim]Run 'nao init' to create a configuration file[/dim]")
-        sys.exit(1)
-    assert config is not None  # Help type checker after sys.exit
+    config = NaoConfig.try_load(exit_on_error=True)
+    assert config is not None  # Help type checker after exit_on_error=True
```
Standlc, line 41: instead of doing
```python
config = NaoConfig.try_load(exit_on_error=True)
assert config is not None  # Help type checker after exit_on_error=True
```
Can you do
```python
config = NaoConfig.load(exit_on_error=True)
```
and do the `assert` inside the method. In other words, throw inside the `try_load` method.

Bl3f: Hmm. Sadly I can't, if the check is done elsewhere the typechecker (pyright or ruff) can't understand it's not None. It's mainly a check for the typechecker than for the runtime itself.

PR #100, opened 2026-01-30
Reworked the `nao init` to make it friendlier

Bl3f:
https://github.com/user-attachments/assets/4b5a1638-7252-49a2-bb1a-92b1e4f573f7


cli/nao_core/commands/init.py
```diff
@@ -65,16 +41,16 @@ def setup_project_name(force: bool = False) -> tuple[str, Path]:
...
-            if force or Confirm.ask("[bold]Re-initialize this project?[/bold]", default=True):
-                return existing_config.project_name, current_dir
+            if force or ask_confirm("Update this project configuration?", default=True):
+                return existing_config.project_name, current_dir, existing_config
             else:
                 raise InitError("Initialization cancelled.")
 
     # Normal flow: prompt for project name
-    project_name = Prompt.ask("[bold]Enter your project name[/bold]")
+    project_name = ask_text("Enter your project name:", required_field=True)
```
Standlc, line 53: why not `UI.ask()`?

cli/nao_core/config/llm/__init__.py
Standlc, line 1: I thought the LLM config was not saved in the yml file anymore. What's the current state of this topic?

cli/nao_core/commands/init.py
Bl3f: It's more about concern UI is mainly about printing something without any input, whereas ask_* methods are asking something to the user, so ask are returning values and UI methods are pure display.

This is not a strong argument tho, so it could be consolidated. But as both are already defined in the same file, I'd say it's ok for now(?)

cli/nao_core/config/llm/__init__.py
Bl3f: LLM config and Slack config are still in the yml, we could remove Slack config from it but LLM no as we might use it for context and tests related tasks. 

To be honest this is not a bid burden, it either comes from the env or from the settings page UI.

PR #104, opened 2026-02-01
Use tanstack form for settings and login forms

Bl3f:
* Use tanstack form
* Structure the settings with subpages

Screenshot 2026-02-01 at 22 06 26


apps/frontend/src/routes/_sidebar-layout.settings.index.tsx
```diff
@@ -0,0 +1,7 @@
+import { createFileRoute, redirect } from '@tanstack/react-router';
+
+export const Route = createFileRoute('/_sidebar-layout/settings/')({
+	beforeLoad: () => {
+		throw redirect({ to: '/settings/profile' });
```
Standlc, line 5: instead of a redirect you could make the `_sidebar-layout.settings.index` render the profile component

apps/frontend/src/components/sidebar.tsx
```diff
@@ -64,33 +68,53 @@ export function Sidebar() {
...
+						</div>
+					</Button>
+				)}
 			</div>
 
-			<ChatList
-				chats={chats.data?.chats || []}
-				className={cn('w-72 transition-[opacity,visibility] duration-300', hideIf(isCollapsed))}
-			/>
+			{isInSettings ? (
```
Standlc, line 110: move this up in the first ternary operation?

apps/frontend/src/components/settings-create-user-form.tsx
```diff
@@ -1,100 +1,79 @@
...
 				});
 				onOpenChange(false);
-				onUserCreated(formData.email, ctx.password);
-			},
-			onError: (err) => {
-				setError(err.message);
+				onUserCreated(form.state.values.email, ctx.password);
+				form.reset();
 			},
+			onError: (err) => setServerError(err.message),
```
Standlc: can't you do `form.setError()` or something?

apps/frontend/src/components/ui/form-fields.tsx
Standlc, line 1: move to ui/

apps/frontend/src/routes/_sidebar-layout.settings.index.tsx
Bl3f: I think it can stay like this so we can easily move around which component is the default.

apps/frontend/src/components/sidebar.tsx
Bl3f: The reason it's this way is to avoid having a useless intermediary div that breaks the positioning from the flex parent.

apps/frontend/src/components/settings-create-user-form.tsx
Bl3f: yep, you're right it's `form.setErrorMap`

PR #105, opened 2026-02-01
Implement posthog tracking in frontend and CLI

Standlc:


apps/backend/src/trpc/router.ts
```diff
@@ -1,11 +1,13 @@
 import { chatRoutes } from './chat.routes';
+import { configRoutes } from './config.routes';
 import { feedbackRoutes } from './feedback.routes';
 import { googleRoutes } from './google.routes';
 import { projectRoutes } from './project.routes';
 import { router } from './trpc';
 import { userRoutes } from './user.routes';
 
 export const trpcRouter = router({
+	config: configRoutes,
```
Bl3f: Not sure we need this if we consider the Posthog key has to be embed in the code.

apps/frontend/src/contexts/posthog.provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+	return (
+		<PostHogEnabledContext.Provider value={true}>
+			<PostHogProviderOriginal
+				apiKey={config.posthog.apiKey}
+				options={{
+					api_host: config.posthog.apiHost,
+					defaults: '2025-05-24',
+					debug: import.meta.env.MODE === 'development',
+					autocapture: false,
```
Bl3f, line 38: As you remove the autocapture, what does it capture right now? The best I'd say would be to capture:
- pageviews
- when a message is sent (with the model id and the chat_id (so we know if it's a new chat))
- when a tool is executed (just the type)
- when a user like/dislike an answer
- when the backend starts (ik it's in nao chat but the best would be to have it also in the backend because it can be started from Docker)
- errors

apps/frontend/src/contexts/posthog.provider.tsx
```diff
@@ -0,0 +1,67 @@
...
+		</PostHogEnabledContext.Provider>
+	);
+}
+
+/**
+ * Safe hook to get the PostHog client.
+ * Use this instead of importing usePostHog from 'posthog-js/react' directly.
+ */
+export function usePostHog(): PostHog | undefined {
+	const isEnabled = useContext(PostHogEnabledContext);
```
Bl3f, line 59: There should be a ALLOW_PRIVATE_TRACKING variable that should activate or not the tracking for the whole parts (frontend, backend, cli) - in the case of the frontend the env variable might not be read as it's in production, so either we keep the config endpoint, either we don't do any tracking in the frontend as most of it can be done in the backend.

apps/frontend/src/hooks/use-identify-posthog.ts
```diff
@@ -0,0 +1,30 @@
...
+
+	useEffect(() => {
+		if (!posthog || session.isPending) {
+			return;
+		}
+
+		const user = session.data?.user;
+		if (user) {
+			wasConnectedRef.current = true;
+			posthog.identify(user.id, {
```
Bl3f, line 21: We should not collect users email for privacy reasons I think, maybe the email domain only.

cli/nao_core/commands/chat.py
```diff
@@ -128,7 +130,10 @@ def chat():
 
     def shutdown_servers():
         """Gracefully shut down both server processes."""
-        for name, proc in [("Chat server", chat_process), ("FastAPI server", fastapi_process)]:
+        for name, proc in [
```
Bl3f: you should fix your config somewhere because this line should not be reformatted i think, I've setup line-lenght to 120 in `pyproject.toml` it's more readable than having something on 80.

cli/nao_core/tracking.py
```diff
@@ -0,0 +1,208 @@
...
+import platform
+import uuid
+from functools import wraps
+from pathlib import Path
+from typing import Any, Callable, TypeVar
+
+import posthog
+
+# PostHog configuration from environment
+POSTHOG_API_KEY = os.environ.get("VITE_PUBLIC_POSTHOG_KEY", "")
```
Bl3f: VITE_ prefix should be removed.

cli/nao_core/tracking.py
```diff
@@ -0,0 +1,208 @@
...
+POSTHOG_HOST = os.environ.get("VITE_PUBLIC_POSTHOG_HOST", "https://us.i.posthog.com")
```
Bl3f: Should be fixed to eu I think as our project is in EU.

Bl3f (changes requested):
A few changes to be done. I think we can keep the env variable in the code, but when we build we gonna fix the env variable to our posthog key. Question is mostly about if we should keep tracking in the frontend and not only in the backend so we are sure it will work.

Lack of a variable to deactivate tracking, also we should do a privacy free tracking imho, so not taking users pii.

Issue #90, opened 2026-01-29
Add Snowflake SSO support in cli


ClaireGz:
As a user with Snowflake SSO-enabled accounts,
I want to authenticate to Snowflake using SSO from the nao CLI,
So that I can connect to my Snowflake instance without managing username/password credentials and comply with my organization's authentication policies.

NewSmoke38:
Hi @ClaireGz,

I would like to work on this issue! I am looking into the implementation now and will open a PR soon.

PR #106, opened 2026-02-01
feat(snowflake): Add snowflake SSO auth support in cli

NewSmoke38:
### Description

Implemented Snowflake SSO authentication support in the nao CLI, allowing users to authenticate to Snowflake using browser-based SSO `externalbrowser` instead of passwords or key-pairs.

### Testing

- Manual Testing
- Ran `nao init` interactively
- Selected `snowflake` database
- Tested SSO prompt: "Use SSO (external browser) for authentication?" 

SSO Flow: Answered y to SSO → Verified password prompt was skipped.
Fallback Flow: Answered n to SSO → Verified CLI correctly asked for `Key-pair/Password`.
Confirmed configuration saved successfully for both cases.

closes: #90

### DIFF
--- a/cli/nao_core/config/databases/snowflake.py
+++ b/cli/nao_core/config/databases/snowflake.py
@@ -36,7 +36,7 @@ class SnowflakeConfig(DatabaseConfig):
         default=None,
         description="Passphrase for the private key if it is encrypted",
     )
-    authenticator: str | None = Field(
+    authenticator: Literal["externalbrowser", "username_password_mfa", "jwt_token", "oauth"] | None = Field(
         default=None,
         description="Authentication method (e.g., 'externalbrowser' for SSO)",
     )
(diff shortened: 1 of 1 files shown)


cli/nao_core/config/databases/snowflake.py
```diff
@@ -35,6 +36,10 @@ class SnowflakeConfig(DatabaseConfig):
         default=None,
         description="Passphrase for the private key if it is encrypted",
     )
+    authenticator: str | None = Field(
```
Standlc: can you make it a `Literal` so that the validation checks the value of the string?

cli/nao_core/config/databases/snowflake.py
```diff
@@ -102,6 +116,12 @@ def connect(self) -> BaseBackend:
         if self.warehouse:
             kwargs["warehouse"] = self.warehouse
 
+        # Add authenticator if specified (e.g., 'externalbrowser' for SSO)
+        if self.authenticator:
+            kwargs["authenticator"] = self.authenticator
+            if self.authenticator == "externalbrowser":
```
Standlc: It looks like `authenticator` can only be "externalbrowser" or `None`. Which means you can skip this check

Standlc (changes requested):
Hey @NewSmoke38!
Thanks for your contribution. Please review the comments!

Issue #42, opened 2026-01-21
Preview jobs failed not shown in the cli logs


ClaireGz:
If one of the jobs fail (ex: create preview.md), there is no error message in cli, it just writes in preview.md ‘error’ etc.

PR #107, opened 2026-02-01
fix: Log accessor generation errors to CLI

NewSmoke38:
### Description

- Added error logging to the **Accessor Layer** (template rendering)
- Errors now display in CLI with file name, table reference, and actual error details
- Previously errors were silently written to markdown files

closes: #42 

## Testing

Two layers in sync operation:

1. **Database Provider Layer** - Connection, schema discovery
2. **Accessor Layer** - Template rendering, markdown generation (this fix)

This PR targets Layer 2. When accessor jobs fail (template errors, data transformation), users now see detailed CLI messages instead of silent failures.

Unit tests mock the database layer to test accessor error handling independently:
- `test_accessor_error_logging_to_console()` - Verifies errors print to CLI
- `test_accessor_returns_error_markdown()` - Verifies markdown error fallback

Both tests pass

### DIFF
--- a/cli/tests/nao_core/commands/sync/test_accessor_error_logging.py
+++ b/cli/tests/nao_core/commands/sync/test_accessor_error_logging.py
@@ -1,0 +1,61 @@
+"""Test that accessor errors are logged to CLI."""
+import pytest
+from unittest.mock import MagicMock, patch
+
+from nao_core.commands.sync.accessors import ColumnsAccessor
+
+
+def test_accessor_error_logging_to_console():
+    """Test that accessor errors are printed to console."""
+    accessor = ColumnsAccessor()
+    
+    # Mock connection to fail during get_context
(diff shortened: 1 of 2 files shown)


cli/tests/nao_core/commands/sync/test_accessor_error_logging.py
```diff
@@ -0,0 +1,61 @@
...
+            table="users"
+        )
+    
+    # Should return markdown with table name and error
+    assert "# users" in result
+    assert "Error generating content" in result
+    assert "Column not found" in result
+
+
+if __name__ == "__main__":
```
SIMOUNIX, line 58: this should not be needed, pytest automatically find class / functions that start with test_ as you are doing here!

Bl3f:
Checks are failing for this.

PR #109, opened 2026-02-02
feat: add project context for multi-project support

### LINKED ISSUE #72: Agent functions should be given the project they are called from
This way we can make sure everything run in isolation and we prep for the multiproject.

HammrLord:
Closes #72

### Changes
- Create `project-context.ts` with **AsyncLocalStorage** for `projectId/projectPath`.
- Wrap agent execution in `projectContext.run()` in `agent.service.ts`.
- Update `getProjectFolder()` to check context first, fallback to env var.
- Update `user-rules.ts` to use `getProjectFolder()`.

### Why
Enables tools to run in isolation per request, preparing foundation for multi-project support in the future.

### DIFF
--- a/apps/backend/src/services/agent.service.ts
+++ b/apps/backend/src/services/agent.service.ts
@@ -163,7 +163,7 @@ class AgentManager {
 				}
 				if (!project.path) {
 					throw new Error(`Project path not configured: ${this.chat.projectId}`);
-        }
+				}
 				const messages = await this._buildInitialMessages(uiMessages, this._modelSelection.provider);
 
 				result = await this._agent.stream({


apps/backend/src/utils/project-context.ts
```diff
@@ -0,0 +1,12 @@
+import { AsyncLocalStorage } from 'async_hooks';
+
+export interface ProjectContext {
+	projectId: string;
+	projectPath: string;
+}
+
+export const projectContext = new AsyncLocalStorage<ProjectContext>();
```
Standlc: Since it's used around the agent, a more appropriate name would be `agentContext`.
Also, I don't think it's necessary to export it.

apps/backend/src/utils/project-context.ts
```diff
@@ -0,0 +1,12 @@
+import { AsyncLocalStorage } from 'async_hooks';
+
+export interface ProjectContext {
+	projectId: string;
```
Standlc: we currently don't need this, let's remove it

apps/backend/src/utils/project-context.ts
```diff
@@ -0,0 +1,12 @@
...
+	projectPath: string;
+}
+
+export const projectContext = new AsyncLocalStorage<ProjectContext>();
+
+export const getProjectContext = (): ProjectContext | undefined => {
```
Standlc: instead of exporting this to context consumers, this file should export specific methods like `getProjectPath`. Then we can do:
```typescript
import * as agentContext from 'project-context'

const projectPath = agentContext.getProjectPath()
```

HammrLord:
PR #109 - Explicit Project Context

**Summary**
Pass project context explicitly via `experimental_context` instead of implicit globals.

**Changes**
- **`types/tools.ts`** — Added `ToolContext` interface + adapter in `createTool`
- **`agent.service.ts`** — Fetches project, passes `{ projectPath }` via `experimental_context`
- **Tool functions** — Accept `context?: ToolContext`, fallback to `getProjectFolder()`

apps/backend/src/agents/tools/functions/execute-sql.ts
```diff
@@ -1,8 +1,9 @@
+import type { ToolContext } from '../../../types/tools';
 import { getProjectFolder } from '../../../utils/tools';
 import type { Input, Output } from '../schema/execute-sql';
 
-export const execute = async ({ sql_query, database_id }: Input): Promise<Output> => {
-	const naoProjectFolder = getProjectFolder();
+export const execute = async ({ sql_query, database_id }: Input, context?: ToolContext): Promise<Output> => {
+	const naoProjectFolder = context?.projectPath ?? getProjectFolder();
```
Bl3f, line 6: Not sure we should keep the fallback to getProjectFolder(), why did you kept it?

apps/backend/src/services/agent.service.ts
```diff
@@ -155,11 +156,23 @@ class AgentManager {
 					});
 				}
 
+				// Fetch project path and run agent within project context
+				const project = await projectQueries.getProjectById(this.chat.projectId);
```
Bl3f, line 160: We should move the if logic elsewhere to keep this method clean.

PR #113, opened 2026-02-02
Add Anthropic prompt caching

Bl3f:
* Activate the prompt caching for Anthropic, code is a bit hard to factorise and keep clean as Anthropic caching is kinda unique and we need to rewrite each message depending if they need to be cached.

### DIFF
--- a/apps/backend/src/agents/providers.ts
+++ b/apps/backend/src/agents/providers.ts
@@ -2,52 +2,13 @@ import { type AnthropicProviderOptions, createAnthropic } from '@ai-sdk/anthropi
 import { createGoogleGenerativeAI } from '@ai-sdk/google';
 import { createMistral } from '@ai-sdk/mistral';
 import { createOpenAI } from '@ai-sdk/openai';
-import type { LanguageModel, ModelMessage } from 'ai';
+import type { LanguageModel } from 'ai';
 
 import type { LlmProvider, LlmProvidersType, ProviderConfigMap } from '../types/llm';
 
 // See: https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching
-const CACHE_1H = { type: 'ephemeral', ttl: '1h' } as const;
-const CACHE_5M = { type: 'ephemeral' } as const;
-
-/**
- * Add Anthropic cache breakpoints to messages.
- * No-op for non-Anthropic providers.
- *
- * Cache strategy:
- * - System message: 1h TTL (instructions rarely change)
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/agent.service.ts
```diff
@@ -1,22 +1,39 @@
...
 import * as chatQueries from '../queries/chat.queries';
 import * as llmConfigQueries from '../queries/project-llm-config.queries';
 import { UIChat, UIMessage } from '../types/chat';
+import type { LlmProvider } from '../types/llm';
 import { convertToTokenUsage } from '../utils/chat';
 import { getDefaultModelId, getEnvApiKey, getEnvModelSelections, ModelSelection } from '../utils/llm';
 
 export type { ModelSelection };
 
+async function buildInitialMessages(
```
Standlc: make it a private method of the AgentManager

apps/backend/src/agents/providers.ts
```diff
@@ -2,10 +2,53 @@ import { type AnthropicProviderOptions, createAnthropic } from '@ai-sdk/anthropi
...
+	provider: LlmProvider,
+	additionalBreakpoints: number[] = [],
+): ModelMessage[] {
+	if (provider !== 'anthropic' || messages.length === 0) {
+		return messages;
+	}
+
+	const breakpointSet = new Set([...additionalBreakpoints, messages.length - 1]);
+
+	return messages.map((message, index) => {
```
Standlc: instead of recreating a new array you could modify the message directly with the breakpoint indices.
```typescript
for (const breakpoint in breakpoints) {
  messages[i] = ...
}
```

apps/backend/src/agents/providers.ts
```diff
@@ -2,10 +2,53 @@ import { type AnthropicProviderOptions, createAnthropic } from '@ai-sdk/anthropi
...
+ * Cache strategy:
+ * - System message: 1h TTL (instructions rarely change)
+ * - Additional breakpoints: 5m TTL (stable conversation prefix)
+ * - Last message: 5m TTL (current step's leaf for agentic caching)
+ *
+ * @param messages - The messages array to add cache markers to
+ * @param provider - The LLM provider
+ * @param additionalBreakpoints - Extra indices to mark with 5m cache (e.g., user message as stable prefix)
+ */
+export function withAnthropicCache(
```
Standlc: this file has to do with available LLM providers and model selection. This handles messaging cache. Would be better to move as a private method in the `AgentManager`

PR #115, opened 2026-02-02
feat(cli): debug for mistralai and gemini support

SIMOUNIX:


cli/nao_core/commands/debug.py
```diff
@@ -58,6 +58,20 @@ def check_llm_connection(llm_config) -> tuple[bool, str]:
...
+            client = genai.Client(api_key=llm_config.api_key)
+            models = client.models.list()
+            model_count = sum(1 for _ in models)
+            return True, f"Connected successfully ({model_count} models available)"
+        elif llm_config.provider.value == "mistral":
+            from mistralai import Mistral
+
+            client = Mistral(api_key=llm_config.api_key)
+            models = client.models.list()
+            model_count = len(models.data)
```
MatLBS: could you reuse `model_count = sum(1 for _ in models)`

MatLBS (approved):
Left one comment, otherwise it looks good to me

MatLBS (approved):

cli/nao_core/commands/debug.py
```diff
@@ -56,6 +56,20 @@ def check_llm_connection(llm_config) -> tuple[bool, str]:
 
             models = client.models.list()
 
+            model_count = sum(1 for _ in models)
```
Bl3f: The whole if/else could be significantly factorised, because we have

```
models = client.models.list()
model_count = sum(1 for _ in models)
```

for every client, so I think we can reduce the size and improve the readability of it.

(maybe there will be a typing error tho)

PR #118, opened 2026-02-03
Add nao test

Bl3f:


apps/backend/src/routes/test.ts
```diff
@@ -0,0 +1,75 @@
...
+	 * Used for testing/evaluation purposes from the CLI.
+	 */
+	app.post(
+		'/run',
+		{
+			schema: {
+				body: z.object({
+					prompt: z.string(),
+					model: modelSelectionSchema.optional(),
+					sql: z.string().optional(),
```
MatLBS: Why are `sql` and `model` optionals ? 

If there is no sql provided we cannot compare the diff ?

Same for model ? 

If not provided `modelSelection` will be undefined and the test simply won't work

apps/backend/src/services/test-agent.service.ts
```diff
@@ -0,0 +1,126 @@
...
+		return agent.generate([userMessage]);
+	}
+
+	/**
+	 * Run a verification prompt to extract structured data from the agent's response.
+	 * Uses the responseMessages directly from the agent result to avoid double transformation.
+	 */
+	async runVerification(
+		projectId: string,
+		prompt: string,
```
MatLBS: variable unused, remove it ?

cli/nao_core/commands/test/runner.py
```diff
@@ -0,0 +1,334 @@
...
+                    break
+            else:
+                # For non-numeric columns, require exact equality
+                if not actual_series.equals(expected_series):
+                    is_close = False
+                    break
+
+        if is_close:
+            return True, "match (approximate)"
+    except Exception:
```
MatLBS, line 139: should not we log the error at least ?

cli/nao_core/commands/test/case.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@dataclass
+class TestCase:
+    """A single test case loaded from a YAML file."""
+
+    name: str
+    prompt: str
+    file_path: Path
+    sql: str | None = None
```
MatLBS: Like I said in my first comment, I think sql should be mandatory. 

Thus, we should check if the data is well retrived

apps/backend/src/routes/test.ts
Bl3f: Yes, right. Changing it.

PR #131, opened 2026-02-05
Feat/usage dashboard

Bl3f:
https://github.com/user-attachments/assets/6cb4adce-0e3b-4fd3-b808-a1260b253b27

### DIFF
--- a/apps/backend/src/utils/date.ts
+++ b/apps/backend/src/utils/date.ts
@@ -72,7 +72,7 @@ export function fillMissingDates(records: UsageRecord[], granularity: Granularit
 		(date) =>
 			dateSet.get(date) ?? {
 				date,
-				nbMessages: 0,
+				messageCount: 0,
 				inputNoCacheTokens: 0,
 				inputCacheReadTokens: 0,
 				inputCacheWriteTokens: 0,
--- a/apps/backend/src/queries/usage.queries.ts
+++ b/apps/backend/src/queries/usage.queries.ts
@@ -1,4 +1,4 @@
-import { and, count, eq, isNotNull, SQL, sql, SQLWrapper, sum } from 'drizzle-orm';
+import { and, eq, isNotNull, SQL, sql, SQLWrapper, sum } from 'drizzle-orm';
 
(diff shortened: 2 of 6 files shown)


apps/backend/src/db/db.ts
```diff
@@ -1,22 +1,25 @@
...
 import { drizzle as drizzlePostgres } from 'drizzle-orm/postgres-js';
+import { EnhancedQueryLogger } from 'drizzle-query-logger';
 import postgres from 'postgres';
 
 import { env } from '../env';
 import dbConfig, { Dialect } from './dbConfig';
 import * as pgSchema from './pgSchema';
 import * as sqliteSchema from './sqliteSchema';
 
+const logger = process.env.DB_QUERY_LOGGING === 'true' ? new EnhancedQueryLogger() : undefined;
```
Standlc: use `env`

apps/backend/src/queries/feedback.queries.ts
```diff
@@ -16,3 +19,51 @@ export const upsertFeedback = async (feedback: NewMessageFeedback): Promise<Mess
 		.execute();
 	return result;
 };
+
+export type FeedbackWithDetails = {
```
Standlc: move to types/message-feedback.ts

apps/backend/src/utils/date.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const lookbackPeriods = {
+	hour: 24,
+	day: 30,
+	month: 12,
+};
+
+export function getLookbackTimestamp(granularity: Granularity): number {
+	const now = Date.now();
+	const periods = lookbackPeriods[granularity];
```
Standlc, line 11: what's the point of using the `lookbackPeriods` if you're going to use a swtich case after? Hard code the `periods` in the switch case or make your `lookbackPeriods` useful

Standlc (approved):

apps/backend/src/utils/date.ts
Bl3f: yes agree, but periods are used twice in the file, one in the switch and one to loop over it

PR #132, opened 2026-02-05
feat(auth): added auth hook to know user roles

SIMOUNIX:
I have not found in frontend anything regarding user role so I am proposing this in order to implement the organization frontend that might render differently based on user roles. This could be use everywhere we need to disable button or whatever for example.

Tanstack query ensure that the data is cached and shared among components.

### DIFF
--- a/apps/frontend/src/hooks/useAuth.ts
+++ b/apps/frontend/src/hooks/useAuth.ts
@@ -1,9 +1,8 @@
 import { useQuery } from '@tanstack/react-query';
+import type { OrgRole } from '../../../backend/src/types/organization';
+import type { UserRole } from '../../../backend/src/types/project';
 import { useSession } from '@/lib/auth-client';
 import { trpc } from '@/main';
-
-type OrgRole = 'admin' | 'user';
-type ProjectRole = 'admin' | 'user' | 'viewer';
 
 interface User {
 	id: string;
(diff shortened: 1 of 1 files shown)


apps/backend/src/trpc/organization.routes.ts
```diff
@@ -0,0 +1,15 @@
+import * as orgQueries from '../queries/organization.queries';
+import { protectedProcedure } from './trpc';
+
+export const organizationRoutes = {
+	getCurrent: protectedProcedure.query(async ({ ctx }) => {
```
SIMOUNIX, line 5: might rename for something more descriptive since it does not return the real current organization

apps/frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,42 @@
+import { useQuery } from '@tanstack/react-query';
+import { useSession } from '@/lib/auth-client';
+import { trpc } from '@/main';
+
+type OrgRole = 'admin' | 'user';
+type ProjectRole = 'admin' | 'user' | 'viewer';
```
Standlc: isn't that declared in the backend? If so, import the types from it

apps/frontend/src/hooks/useAuth.ts
```diff
@@ -0,0 +1,42 @@
...
+		enabled: !!session?.user,
+	});
+	const org = useQuery({
+		...trpc.organization.getCurrent.queryOptions(),
+		enabled: !!session?.user,
+	});
+
+	return {
+		user: session?.user ?? null,
+		projectRole: (project.data?.userRole as ProjectRole) ?? null,
```
Standlc: Why the type casting? Instead make sure the correct type is returned from the `trpc` procedures

PR #136, opened 2026-02-06
fix(slack): correct redirect URL for chat links

### LINKED ISSUE #127: Slack bot "View full conversation" doesn't work
When in a slack bot thread, when clicking on "View full conversation", the link opens nao UI on "Not Found"

MatLBS:
Fixes #127 

Remove the /p/{projectId}/ prefix from the redirectUrl that was preventing the 'View full conversation' link from working in Slack threads.

The frontend route is /{chatId}, not /p/{projectId}/{chatId}, so the redirectUrl should just be the base URL.

### DIFF
--- a/apps/backend/src/queries/project-slack-config.queries.ts
+++ b/apps/backend/src/queries/project-slack-config.queries.ts
@@ -82,13 +82,11 @@ export async function getSlackConfig(): Promise<SlackConfig | null> {
 		return null;
 	}
 
-	const baseUrl = redirectUrl.endsWith('/') ? redirectUrl.slice(0, -1) : redirectUrl;
-
 	return {
 		projectId: project.id,
 		botToken,
 		signingSecret,
-		redirectUrl: `${baseUrl}/p/${project.id}/`,
+		redirectUrl,
 	};
 }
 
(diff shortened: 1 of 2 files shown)


apps/backend/src/queries/project-slack-config.queries.ts
```diff
@@ -88,7 +88,7 @@ export async function getSlackConfig(): Promise<SlackConfig | null> {
 		projectId: project.id,
 		botToken,
 		signingSecret,
-		redirectUrl: `${baseUrl}/p/${project.id}/`,
+		redirectUrl: `${baseUrl}/`,
```
Standlc: why the trailing slash? you can probably remove it too.

Bl3f: we should use a lib or a function that concats url so we stop asking ourselfves if it works

MatLBS: Why do you want to use a lib for that ? 

It works well with only baseUrl

Bl3f: because we never know if we concat correctly the url

Standlc: it already exists in Node. To concatenate paths, you can do
```typescript
new URL('hello', 'http://localhost').toString() // => 'http://localhost/hello'
```
`URL` will throw if not valid

Bl3f: Perfect 

(fyi in Python often I use os.path which does the same without url validation)

PR #138, opened 2026-02-06
Change deployement branch to prod, fix prod environnement detection i…

Standlc:
Minor fixes and changes:
- make `prod` the deployment branch so we can batch hotfixes and PRs into one GCP deployement
- detect Fastify environnement based on `env` variable
- correctly identify users in chat message event tracking

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,8 +2,6 @@ name: Publish to PyPI
 
 on:
     push:
-        branches:
-            - prod
         tags:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -2,6 +2,8 @@ name: Publish to PyPI
 
 on:
     push:
+        branches:
```
Bl3f: Remove this, the packages publish in only on tags.

PR #142, opened 2026-02-08
fix: add authorization check for chat deletion

Dozie2001:
Found that any logged-in user can delete someone else's chat if they know the chat ID. The delete endpoint only checks if you're logged in, not if you actually own the chat. The other endpoints like rename and get already check ownership, so delete was just missed.               
Added the same ownership check pattern to delete, so it now returns a 403 if you try to delete a chat that isn't yours

 How to reproduce

  1. Sign up two users
  2. Create a chat as User A
  3. As User B, call chat.delete with User A's chat ID
  4. Before this fix, it goes through silently. After, you get a 403.

  What I changed

 Before deleting, it now looks up who owns the chat and blocks the request if it's not you. Same approach rename already uses a few lines down.

### DIFF
--- a/apps/backend/src/trpc/chat.routes.ts
+++ b/apps/backend/src/trpc/chat.routes.ts
@@ -4,7 +4,9 @@ import { z } from 'zod/v4';
 import * as chatQueries from '../queries/chat.queries';
 import { agentService } from '../services/agent.service';
 import { type ListChatResponse, type UIChat } from '../types/chat';
-import { protectedProcedure } from './trpc';
+import { ownedResourceProcedure, protectedProcedure } from './trpc';
+
+const chatOwnerProcedure = ownedResourceProcedure(chatQueries.getChatOwnerId, 'chat');
 
 export const chatRoutes = {
 	get: protectedProcedure.input(z.object({ chatId: z.string() })).query(async ({ input, ctx }): Promise<UIChat> => {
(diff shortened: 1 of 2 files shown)


apps/backend/src/trpc/chat.routes.ts
```diff
@@ -23,9 +23,18 @@ export const chatRoutes = {
 		return chatQueries.listUserChats(ctx.user.id);
 	}),
 
-	delete: protectedProcedure.input(z.object({ chatId: z.string() })).mutation(async ({ input }): Promise<void> => {
-		await chatQueries.deleteChat(input.chatId);
-	}),
+	delete: protectedProcedure
```
Bl3f: the best would be to create a like "ownProtectedProcedure" which makes it check if it's your resource your work on, so we can re-use it elsewhere

@claude propose what's the best name for such a factorised procedure

Dozie2001: ah that makes sense ok will fix it now

Bl3f: i can understand the object entity might be different each time, so hard for the generic procedure to know on what to check ; worth to explore if there is a factorisation, if not, good to go with this

Dozie2001: "ownedResourceProcedure" isn't bad

PR #148, opened 2026-02-09
Fix Claude review workflow, create intentionnal bug for Claude to find

Standlc:


apps/backend/src/app.ts
```diff
@@ -123,6 +122,8 @@ if (staticRoot) {
 }
 
 export const startServer = async (opts: { port: number; host: string }) => {
+	throw new Error('test');
```
Bl3f, line 125: Not gonna merge this 🤔

PR #155, opened 2026-02-09
fix: add authorization check in password change mutation

Dozie2001:
Found that the modifyPassword endpoint only checks if you're logged in, not if the  password you're changing is actually yours. Any authenticated user could change another user's password by passing their userId. Added a check to make sure input.userId matches the logged-in user before allowing the
   password change.

### DIFF
--- a/apps/backend/src/trpc/account.routes.ts
+++ b/apps/backend/src/trpc/account.routes.ts
@@ -55,20 +55,12 @@ export const accountRoutes = {
 	modifyPassword: projectProtectedProcedure
 		.input(
 			z.object({
-				userId: z.string(),
 				newPassword: z.string(),
 				confirmPassword: z.string(),
 			}),
 		)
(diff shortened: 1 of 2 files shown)


apps/backend/src/trpc/account.routes.ts
```diff
@@ -60,7 +60,14 @@ export const accountRoutes = {
 				confirmPassword: z.string(),
 			}),
 		)
-		.mutation(async ({ input }) => {
+		.mutation(async ({ input, ctx }) => {
+			if (input.userId !== ctx.user.id) {
```
Standlc: I think you can remove the `input` altogether and use the `ctx.user.id` directly which safely identifies the user making the request.

Standlc (changes requested):
Hey, thanks for your contribution! Please check the comment I left.

PR #162, opened 2026-02-10
Feature/project context isolation

### LINKED ISSUE #72: Agent functions should be given the project they are called from

MatLBS:
Closes #72 

New PR based on previous #109

### DIFF
--- a/apps/backend/src/agents/tools/execute-python.ts
+++ b/apps/backend/src/agents/tools/execute-python.ts
@@ -21,7 +21,10 @@ const RESOURCE_LIMITS = {
 	maxRecursionDepth: 500,
 };
 
-async function executePython({ code, inputs }: schemas.Input, projectFolder: string): Promise<schemas.Output> {
+async function executePython(
+	{ code, inputs }: schemas.Input,
+	{ projectFolder }: { projectFolder: string },
+): Promise<schemas.Output> {
 	if (!montyModule) {
 		throw new Error('Python execution is not available on this platform');
 	}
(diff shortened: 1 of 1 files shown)


apps/backend/src/services/agent.service.ts
```diff
@@ -180,11 +180,18 @@ class AgentManager {
...
+				const project = await retrieveProjectById(this.chat.projectId);
+
 				const messages = await this._buildModelMessages(uiMessages);
 
 				result = await this._agent.stream({
 					messages,
 					abortSignal: this._abortController.signal,
+					// @ts-expect-error - experimental_context is not yet in the types
+					experimental_context: {
+						projectPath: project.path,
```
Bl3f: in order for the code to match can you either name this projectFolder or use projectPath on the other side?

apps/backend/src/agents/tools/execute-python.ts
```diff
@@ -20,9 +20,9 @@ const RESOURCE_LIMITS = {
 	maxRecursionDepth: 500,
 };
 
-export async function executePython({ code, inputs }: schemas.Input): Promise<schemas.Output> {
+async function executePython({ code, inputs }: schemas.Input, projectFolder: string): Promise<schemas.Output> {
```
Bl3f: the best here for the future would be not to unpack the projectFolder but to keep an object and do something like instead
```suggestion
async function executePython({ code, inputs }: schemas.Input, { projectFolder: string} ): Promise<schemas.Output> {
```

PR #177, opened 2026-02-11
Fix MCP settings empty state display (fixes #174)

### LINKED ISSUE #174: MCP server default explaination
If there is no mcp.json it should display a text explaining that you have to provide a mcp.json in the context folder so it works, not an empty table.

Image

MatLBS:
Capture d’écran 2026-02-11 à 16 42 18

Fixes #174

### DIFF
--- a/example/nao_config.yaml
+++ b/example/nao_config.yaml
@@ -16,4 +16,4 @@ notion:
 llm: null
 slack: null
 mcp:
-  json_file_path: {{ env('MCP_JSON_FILE_PATH') }}
+  json_file_path: ./agent/mcps/mcp.json
--- a/cli/nao_core/config/mcp/__init__.py
+++ b/cli/nao_core/config/mcp/__init__.py
@@ -21,22 +21,29 @@ class McpConfig(BaseModel):
         json_file_path = ask_text(
             "MCP JSON file path:",
             required_field=True,
-            default="agent/mcps/mcp.json",
+            default="./agent/mcps/mcp.json",
         )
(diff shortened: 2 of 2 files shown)


Bl3f (approved):
LGTM

example/nao_config.yaml
```diff
@@ -15,3 +15,5 @@ notion:
   - https://naolabs.notion.site/Jaffle-shop-information-2f8c7a70bc0680a4b7d0caf99f055360
 llm: null
 slack: null
+mcp:
+  json_file_path: {{ env('MCP_JSON_FILE_PATH') }}
```
Bl3f: As u added this, maybe you should add an empty mcp.json file in the example?

MatLBS: In example I already added a mcp.json file with the metabase setup, is it good for you ?


```
{
  "mcpServers": {
    "metabase": {
      "command": "npx",
      "args": [
        "-y",
        "@getnao/metabase-mcp-server@latest"
      ],
      "env": {
        "METABASE_URL": "${METABASE_URL}",
        "METABASE_API_KEY": "${METABASE_API_KEY}"
      }
    }
  }
}

```

Bl3f: Sorry, I was looking at the example folder, it was in a subfolder. Then in this case you don't need to put an env variable for the mcp path right? Can't you use relative path?

MatLBS: Well `/agents/mcps` is the default mcp file location but what if the user change it ?

PR #178, opened 2026-02-11
fix(backend): return the message with cache control logic

SIMOUNIX:
was not using the message built with _withCache for anthropic provider


Bl3f:
Are you sure it was not working before?

PR #179, opened 2026-02-11
tool outputs versionning and rendering to markdown

Standlc:


apps/backend/scripts/test-markdown-rendering.tsx
```diff
@@ -0,0 +1,155 @@
+import React from 'react';
```
Bl3f: Write real unit tests instead of doing this.

PR #196, opened 2026-02-12
feat: add partition/clustering column detection

### LINKED ISSUE #158: Teach agent to query partitioned / clustered table
As a user, I want the agent to query a partitioned table always indicating a partition

Dozie2001:
closes #158


cli/nao_core/config/databases/bigquery.py
```diff
@@ -139,6 +168,23 @@ def fetch_column_descriptions(self, conn: BaseBackend, schema: str, table_name:
         except Exception:
             return {}
 
+    def create_context(self, conn: BaseBackend, schema: str, table_name: str) -> "DatabaseContext":
```
Bl3f: It feels a bit weird that we have to do it this way. Why can we add this to the provider.py and eventually pass it to the DatabaseContext. I think there might be a design issue around the fetch_table_description and fetch_column_descriptions for the Context, would be better if we can just implement this and inherits context from sub database types.

Wdyt?

Bl3f (changes requested):
Thank you so much for this, there is a design choice that bugs me a bit.

cli/nao_core/config/databases/bigquery.py
Dozie2001: fixed

cli/nao_core/config/databases/context.py
```diff
@@ -0,0 +1,77 @@
+"""Base database context exposing methods available in templates during sync."""
```
Bl3f, line 1: This should not be at the root level of the CLI but if possible in providers/databases folder or config/databases.

Dozie2001: Fixed this. @Bl3f was wondering if. There is an opportunity to hire engineers remotely at nao

Bl3f (approved):
Perfect, LGTM

PR #209, opened 2026-02-16
feat: add dbt project indexer for fast model lookup

### LINKED ISSUE #210: feat: dbt project indexer for fast model lookup
## Problem

When users ask the agent lineage or model questions (e.g., _"What is the lineage of BASE_EMAIL__BREAD_SAVINGS?"_ or _"What dbt model creates the fact_page_impression table?"_), the agent has to **grep through 694+ raw SQL files** in `repos/`. This causes:

1. **Slow and unreliable searches** — scanning hundreds of files often returns incomplete results or times out
2. **Missed downstream dependencies** — the agent can find a model's `ref()` calls (upstream), but has no efficient way to find which other models reference it (downstream)
3. **No source-to-database mapping** — users asking _"What Snowflake table does `source('core', 'dim_offers')` point to?"_ have to manually correlate YAML source definitions with database schemas
4. **Repeated work** — every question re-scans the same files with no cached representation of the project structure

## Proposed Solution

A **backend-side dbt indexer** that parses all dbt SQL/YAML files in `repos/` and generates searchable markdown index files:

- **`manifest.md`** — one entry per model with path, materialization, refs (upstream), sources, and description. Sorted alphabetically for consistent grep results.
- **`sources.md`** — source-to-database mapping with schema and table names.

The agent reads a single `manifest.md` (~4000 lines for 695 models) instead of grepping through hundreds of individual SQL files.

### Use cases

| Use case | Before | After |
|----------|--------|-------|
| "What is the lineage of model X?" | Grep 694 SQL files, often miss downstream refs | Grep 1 file, find upstream + downstream in one search |
| "What creates table Y?" | Grep all files hoping the model name matches | Search manifest for model name, get path + materialization |
| "What source feeds into model Z?" | Read SQL, then find YAML source def | `sources` field inline in manifest |
| "Map source to Snowflake database" | Manually find YAML, parse database/schema | Grep `sources.md`, get database + schema + tables |
| "Show all incremental models" | Impossible without scanning every file | Grep manifest for `materialized: incremental` |

### Trigger points

| Trigger | When |
|---------|------|
| `nao sync` | After repos are cloned/pulled |
| FastAPI startup | When `nao chat` starts |
| `POST /api/refresh` | Manual or webhook trigger |
| Cron scheduler | `NAO_REFRESH_SCHEDULE` env var |

## Implementation

See PR #209 for the complete implementation.

ealexisaraujo:
## Summary

- **New dbt indexer** parses all SQL/YAML files in `repos/` and generates searchable markdown index files (`manifest.md` + `sources.md`) per dbt project
- **Hooks into `nao sync`** as a post-sync step, plus FastAPI startup, `/api/refresh`, and optional cron scheduler
- **Updates system prompt** to instruct the agent to search `dbt-index/` first before grepping raw SQL files
- **32 unit tests** covering SQL parsing, YAML parsing, project discovery, and error handling

## Motivation

When users asked the agent lineage or model questions (e.g., _"What is the lineage of stg_orders?"_ or _"What dbt model creates the fact_revenue table?"_), the agent had to **grep through hundreds of raw SQL files** in `repos/`. This caused several problems:

1. **Slow and unreliable** — searching hundreds of files often timed out or returned incomplete results
2. **Missed downstream dependencies** — the agent could find a model's `ref()` calls (upstream), but had no efficient way to find which other models reference it (downstream). It would have to grep every single SQL file for the model name
3. **No source-to-database mapping** — users asking _"What Snowflake table does `source('core', 'dim_offers')` point to?"_ had to manually correlate YAML source definitions with database schemas
4. **Repeated work** — every question re-scanned the same files. There was no cached representation of the project structure

### Use cases this enables

| Use case | Before | After |
|----------|--------|-------|
| "What is the lineage of model X?" | Grep hundreds of SQL files, often miss downstream refs | Grep 1 file (`manifest.md`), find upstream `refs` + downstream in one search |
| "What creates table Y?" | Grep all files hoping the model name matches | Search `manifest.md` for the model name, immediately get path + materialization |
| "What source feeds into model Z?" | Read the SQL file, then find the YAML source def | `sources` field shows `source_name.table_name` inline in manifest |
| "Map this source to a database table" | Manually find YAML, parse database/schema | Grep `sources.md` for the source name, get database + schema + tables |
| "Show me all incremental models" | Impossible without scanning every file | Grep `manifest.md` for `materialized: incremental` |

### Why index at the Python (FastAPI/CLI) layer

- FastAPI already owns the context lifecycle (refresh/startup)
- Python has mature YAML and regex parsing (`pyyaml` already a dependency)
- Output is file-based, consistent with the existing `databases/` pattern
- Both `nao sync` (CLI) and FastAPI (server) can import from `nao_core.dbt_indexer`

### Data flow

```
nao sync -p repositories
  │
  ├─ git clone/pull → repos/<name>/
  │
  └─ _index_dbt_projects()
       │
       ├─ find_dbt_projects()        ← scans repos/ for dbt_project.yml
       ├─ read_project_config()      ← reads project name, default materializations
       ├─ index_dbt_project()
       │    ├─ parse_yaml_sources()      ← YAML: source definitions
       │    ├─ parse_yaml_descriptions() ← YAML: model descriptions
       │    ├─ parse_sql_dependencies()  ← SQL: ref() and source() calls
       │    └─ parse_sql_config()        ← SQL: materialization config
       ├─ generate_manifest_md()     ← sorted alphabetically for grep
       └─ generate_sources_md()      ← source-to-database mapping
```

### Output (generated in user's project folder)

```
<project>/
├── dbt-index/
│   └── <repo-name>/
│       ├── manifest.md    (grep-friendly, one entry per model)
│       └── sources.md     (source-to-database mapping)
├── repos/                 (raw git repos — still needed for full SQL)
└── databases/             (database schema docs)
```

### Trigger points (4 ways to run)

| Trigger | When | Where |
|---------|------|-------|
| `nao sync` | After repos are cloned/pulled | `cli/nao_core/commands/sync/__init__.py` |
| FastAPI startup | When `nao chat` starts | `main.py` lifespan |
| `POST /api/refresh` | Manual or webhook trigger | `main.py` refresh endpoint |
| Cron scheduler | `NAO_REFRESH_SCHEDULE` env var | `main.py` APScheduler |

### manifest.md (grep-friendly)

```markdown
### stg_core__dim_products
- **path:** models/staging/core/stg_core__dim_products.sql
- **materialized:** view
- **refs:** —
- **sources:** core.dim_products

### int_orders__enriched
- **path:** models/transform/orders/int_orders__enriched.sql
- **materialized:** incremental
- **refs:** stg_core__dim_products, stg_core__fact_orders
- **sources:** —
```

The agent greps `manifest.md` for a model name and instantly finds:
- **Upstream:** `sources` and `refs` fields
- **Downstream:** grep for the model name in `refs:` lines of other models

## Files changed

| File | Change |
|------|--------|
| `cli/nao_core/dbt_indexer.py` | **New** — Core indexer implementation (single source of truth): SQL/YAML parsing, manifest/sources generation |
| `apps/backend/fastapi/dbt_indexer.py` | **New** — Re-export shim so FastAPI imports work |
| `apps/backend/fastapi/main.py` | Hook indexer into startup, refresh endpoint, and scheduler |
| `apps/backend/src/agents/user-rules.ts` | Add `indexed` field to Repository type, detect `dbt-index/` |
| `apps/backend/src/components/system-prompt.tsx` | Tell agent to search `dbt-index/` first; show indexed status |
| `cli/nao_core/commands/sync/__init__.py` | Run indexer after `nao sync -p repositories` |
| `apps/backend/fastapi/test_dbt_indexer.py` | **New** — 32 unit tests |

## Performance

- Tested on a real dbt project with **694 models + 129 YAML files** — indexed in **0.83 seconds**
- Handles edge cases: directories named `.sql`, binary files, Jinja in YAML, malformed YAML

## Test plan

- [x] `python -m pytest apps/backend/fastapi/test_dbt_indexer.py -v` — 32 tests pass
- [x] `make lint` (cli/) — ty, ruff check, ruff format all pass
- [x] `npm run lint` — TypeScript lint passes (0 errors)
- [x] Run `nao sync -p repositories` on real project — `dbt-index/` generated correctly
- [x] Run `nao chat` — agent searches `dbt-index/**/manifest.md` first (verified)
- [x] Test lineage question — agent finds upstream + downstream from manifest
- [x] `POST /api/refresh` — returns 200, triggers re-indexing
- [x] Verify `sources.md` contains correct source-to-database mappings

Fixes #210

### DIFF
--- a/apps/backend/fastapi/main.py
+++ b/apps/backend/fastapi/main.py
@@ -11,8 +11,6 @@ from fastapi import FastAPI, HTTPException
 from fastapi.middleware.cors import CORSMiddleware
 from pydantic import BaseModel
 
-from nao_core.dbt_indexer import index_all_projects
-
 load_dotenv()
 
 cli_path = Path(__file__).parent.parent.parent / "cli"
@@ -25,19 +23,6 @@ port = int(os.environ.get("PORT", 8005))
 
 # Global scheduler instance
 scheduler = None
-
-
-def _run_dbt_indexing(source: str) -> None:
-    """Run dbt indexing with error handling."""
-    try:
-        project_path = Path(
-            os.environ.get("NAO_DEFAULT_PROJECT_PATH", "/app/context")
-        )
-        if project_path.exists():
-            index_all_projects(project_path)
-            print(f"[{source}] dbt projects indexed at {datetime.now().isoformat()}")
-    except Exception as e:
-        print(f"[{source}] Failed to index dbt projects: {e}")
 
 
 @asynccontextmanager
@@ -66,8 +51,6 @@ async def lifespan(app: FastAPI):
         except ValueError as e:
             print(f"[Scheduler] Invalid cron expression '{refresh_schedule}': {e}")
 
-    _run_dbt_indexing("Startup")
-
     yield
 
     # Shutdown scheduler
@@ -82,7 +65,6 @@ async def _refresh_context_task():
         updated = provider.refresh()
         if updated:
             print(f"[Scheduler] Context refreshed at {datetime.now().isoformat()}")
-            _run_dbt_indexing("Scheduler")
         else:
             print(
                 f"[Scheduler] Context already up-to-date at {datetime.now().isoformat()}"
@@ -171,7 +153,6 @@ async def refresh_context():
         updated = provider.refresh()
 
         if updated:
-            _run_dbt_indexing("Refresh")
             return RefreshResponse(
                 status="ok",
                 updated=True,
--- a/apps/backend/src/components/system-prompt.tsx
+++ b/apps/backend/src/components/system-prompt.tsx
@@ -1,4 +1,4 @@
-import { getConnections, getUserRules } from '../agents/user-rules';
+import { getConnections, getRepositories, getUserRules } from '../agents/user-rules';
 import { Block, Bold, Br, Italic, Link, List, ListItem, Location, Span, Title } from '../lib/markdown';
 import { skillService } from '../services/skill.service';
 
@@ -6,6 +6,8 @@ export function SystemPrompt() {
 	const userRules = getUserRules();
 	const connections = getConnections();
 	const skills = skillService.getSkills();
+	const repositories = getRepositories();
+	const hasDbtProjects = repositories?.some((r) => r.hasDbtProject) ?? false;
 
 	return (
 		<Block>
@@ -71,17 +73,22 @@ export function SystemPrompt() {
 				</ListItem>
 				<ListItem>
 					The <Italic>repos</Italic> folder contains synced git repositories with source code, data
-					transformation logic (dbt models), and documentation.
+					transformation logic, and documentation.
 				</ListItem>
-				<ListItem>
-					The <Italic>dbt-index</Italic> folder contains pre-built indexes of dbt models for fast lookup. Use{' '}
-					<Italic>grep</Italic> on manifest.md to find models by name, and sources.md for source-to-database
-					mappings.
-				</ListItem>
-				<ListItem>
-					When asked about a database table, first search <Italic>dbt-index/</Italic> manifest.md for the
-					model name, then read the actual SQL file from repos/ for full business logic and transformations.
-				</ListItem>
+				{hasDbtProjects && (
+					<>
+						<ListItem>
+							The <Italic>dbt-index</Italic> folder contains pre-built indexes of dbt models for fast
+							lookup. Use <Italic>grep</Italic> on manifest.md to find models by name, and sources.md for
+							source-to-database mappings.
+						</ListItem>
+						<ListItem>
+							When asked about a database table, first search <Italic>dbt-index/</Italic> manifest.md for
+							the model name, then read the actual SQL file from repos/ for full business logic and
+							transformations.
+						</ListItem>
+					</>
+				)}
 			</List>
 			<Title level={2}>SQL Query Rules</Title>
 			<List>
--- a/cli/nao_core/commands/sync/__init__.py
+++ b/cli/nao_core/commands/sync/__init__.py
@@ -8,7 +8,6 @@ from cyclopts import Parameter
 from rich.console import Console
 
 from nao_core.config import NaoConfig
-from nao_core.dbt_indexer import index_all_projects
 from nao_core.templates.render import render_all_templates
 from nao_core.tracking import track_command
 
@@ -21,21 +20,6 @@ from .providers import (
 )
 
 console = Console()
-
-
-def _index_dbt_projects(project_path: Path, results: list[SyncResult]) -> None:
-    """Index dbt projects in repos/ after repository sync."""
-    repos_synced = any(r.provider_name == "Repositories" and r.success and r.items_synced > 0 for r in results)
-    repos_dir = project_path / "repos"
-
-    if not repos_synced and not repos_dir.is_dir():
-        return
-
-    try:
-        console.print("\n[bold cyan]📇 Indexing dbt projects[/bold cyan]\n")
-        index_all_projects(project_path)
-    except Exception as e:
-        console.print(f"  [yellow]⚠[/yellow] Failed to index dbt projects: {e}")
 
 
 @track_command("sync")
@@ -118,9 +102,6 @@ def sync(
             # Capture error but continue with other providers
             results.append(SyncResult.from_error(sync_provider.name, e))
             console.print(f"  [yellow]⚠[/yellow] {sync_provider.emoji} {sync_provider.name}: [red]{e}[/red]")
-
-    # Index dbt projects if repos were synced
-    _index_dbt_projects(project_path, results)
 
     # Render user Jinja templates
     template_result = None
--- a/cli/nao_core/dbt_indexer.py
+++ b/cli/nao_core/dbt_indexer.py
@@ -7,14 +7,13 @@ the dbt-index/ folder.
 
 import re
 from datetime import datetime, timezone
+from itertools import chain
 from pathlib import Path
 from typing import NamedTuple
 
 import yaml
 
-# ---------------------------------------------------------------------------
-# Data structures
-# ---------------------------------------------------------------------------
+from nao_core.ui import UI
 
 
 class ModelInfo(NamedTuple):
@@ -33,10 +32,6 @@ class SourceInfo(NamedTuple):
     tables: list[str]
 
 
-# ---------------------------------------------------------------------------
-# Compiled regex patterns
-# ---------------------------------------------------------------------------
-
 RE_REF = re.compile(
     r"""\{\{\s*ref\(\s*['"]([^'"]+)['"]\s*\)\s*\}\}""",
 )
@@ -49,12 +44,10 @@ RE_CONFIG_MATERIALIZED = re.compile(
 )
 RE_JINJA_BLOCK = re.compile(r"\{%.*?%\}", re.DOTALL)
 RE_JINJA_EXPR = re.compile(r"\{\{.*?\}\}", re.DOTALL)
+RE_DBT_PROJECT_NAME = re.compile(r"""^name:\s*['"]?([^'"#\n]+)""", re.MULTILINE)
 
 SKIP_DIRS = {"dbt_packages", "dbt_modules", "target", "logs", ".git"}
-
-# ---------------------------------------------------------------------------
-# Entry point
-# ---------------------------------------------------------------------------
+MAX_PROJECT_DEPTH = 3
 
 
 def index_all_projects(project_folder: Path) -> None:
@@ -74,7 +67,7 @@ def index_all_projects(project_folder: Path) -> None:
         try:
             _index_single_project(repo_name, project_path, index_root, project_folder)
         except Exception as e:
-            print(f"[dbt-indexer] Error indexing {repo_name}: {e}")
+            UI.error(f"Error indexing {repo_name}: {e}")
 
 
 def _index_single_project(
@@ -99,12 +92,18 @@ def _index_single_project(
     sources_md = generate_sources_md(sources, repo_name)
     (out_dir / "sources.md").write_text(sources_md, encoding="utf-8")
 
-    print(f"[dbt-indexer] Indexed {repo_name}: {len(models)} models, {len(sources)} sources")
+    UI.success(f"Indexed {repo_name}: {len(models)} models, {len(sources)} sources")
 
 
-# ---------------------------------------------------------------------------
-# Project discovery
-# ---------------------------------------------------------------------------
+def _find_dbt_project_yml(repo_dir: Path) -> Path | None:
+    """Search for dbt_project.yml within a repo up to MAX_PROJECT_DEPTH."""
+    for depth in range(MAX_PROJECT_DEPTH + 1):
+        pattern = "/".join(["*"] * depth) + "/dbt_project.yml" if depth else "dbt_project.yml"
+        matches = list(repo_dir.glob(pattern))
+        for match in matches:
+            if match.is_file() and not _should_skip(match):
+                return match
+    return None
 
 
 def find_dbt_projects(repos_path: Path) -> list[tuple[str, Path]]:
@@ -115,20 +114,11 @@ def find_dbt_projects(repos_path: Path) -> list[tuple[str, Path]]:
         if not entry.is_dir():
             continue
 
-        root_project = entry / "dbt_project.yml"
-        nested_project = entry / "dbt" / "dbt_project.yml"
-
-        if root_project.is_file():
-            results.append((entry.name, entry))
-        elif nested_project.is_file():
-            results.append((entry.name, entry / "dbt"))
+        project_yml = _find_dbt_project_yml(entry)
+        if project_yml:
+            results.append((entry.name, project_yml.parent))
 
     return results
-
-
-# ---------------------------------------------------------------------------
-# Project indexing
-# ---------------------------------------------------------------------------
 
 
 def index_dbt_project(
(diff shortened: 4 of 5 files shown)


Bl3f (commented):
Hey, thank you so much for the contribution, to be honest this is great. There are a few things to change I think, a few part to refacto a bit or remove and the main question is around the manifest.md that might be a large file and how we should mitigate this in order to not overflow users context windows.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+    database: str | None
+    schema_name: str
+    tables: list[str]
+
+
+# ---------------------------------------------------------------------------
+# Compiled regex patterns
+# ---------------------------------------------------------------------------
+
+RE_REF = re.compile(
```
Bl3f, line 35: As you tried it on a real dbt project I'm eager to see if this combination of regular expressions worked well, I'm a bit unsure to do it with regexp as dbt jinja syntax could be in some wild projects not compatible with this and we could miss a lot of lineage information this way (like generated join from macros, etc.)

cli/nao_core/commands/sync/__init__.py
```diff
@@ -103,6 +119,9 @@ def sync(
             results.append(SyncResult.from_error(sync_provider.name, e))
             console.print(f"  [yellow]⚠[/yellow] {sync_provider.emoji} {sync_provider.name}: [red]{e}[/red]")
 
+    # Index dbt projects if repos were synced
+    _index_dbt_projects(project_path, results)
```
Bl3f: I don't think this should be called from there, I'd say as it's related to repos provider that it's called from there as the sync method here is kinda generic.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+    models_dir = project_path / "models"
+    if not models_dir.is_dir():
+        return [], []
+
+    models: list[ModelInfo] = []
+    sources: list[SourceInfo] = []
+    descriptions: dict[str, str] = {}
+
+    # First pass: collect YAML descriptions and sources
+    for yaml_path in models_dir.rglob("*.yml"):
```
Bl3f: Feel like this code is repeated twice.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+def find_dbt_projects(repos_path: Path) -> list[tuple[str, Path]]:
+    """Return (repo_name, dbt_project_dir) for each repo with a dbt project."""
+    results: list[tuple[str, Path]] = []
+
+    for entry in sorted(repos_path.iterdir()):
+        if not entry.is_dir():
+            continue
+
+        root_project = entry / "dbt_project.yml"
+        nested_project = entry / "dbt" / "dbt_project.yml"
```
Bl3f: There are cases I think where the nested folder is not named dbt

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+            descriptions.update(parse_yaml_descriptions(yaml_path))
+        except Exception as e:
+            print(f"[dbt-indexer] Warning: failed to parse descriptions from {yaml_path}: {e}")
+        try:
+            sources.extend(parse_yaml_sources(yaml_path))
+        except Exception as e:
+            print(f"[dbt-indexer] Warning: failed to parse sources from {yaml_path}: {e}")
+
+    # Second pass: parse SQL models
+    for sql_path in models_dir.rglob("*.sql"):
```
Bl3f, line 150: I think the 3 forloops here could be factorised and make this code way easier to read.

apps/backend/src/agents/user-rules.ts
```diff
@@ -28,6 +28,58 @@ export function getUserRules(): string | null {
...
+	try {
+		const entries = readdirSync(reposPath, { withFileTypes: true });
+		const repositories: Repository[] = [];
+
+		for (const entry of entries) {
+			if (!entry.isDirectory()) {
+				continue;
+			}
+
+			const rootDbtProject = join(reposPath, entry.name, 'dbt_project.yml');
```
Bl3f, line 59: The whole search dbt_project.yml should be factorised in a search function (maybe we can use the search tool from the backend).

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+    manifest = generate_manifest_md(models, repo_name, rel_project_path, project_name)
+    (out_dir / "manifest.md").write_text(manifest, encoding="utf-8")
+
+    sources_md = generate_sources_md(sources, repo_name)
+    (out_dir / "sources.md").write_text(sources_md, encoding="utf-8")
+
+    print(f"[dbt-indexer] Indexed {repo_name}: {len(models)} models, {len(sources)} sources")
+
+
+# ---------------------------------------------------------------------------
```
Bl3f: All these comments can be removed, the function names are enough.

apps/backend/src/agents/user-rules.ts
```diff
@@ -28,6 +28,58 @@ export function getUserRules(): string | null {
...
 }
 
+type Repository = {
+	name: string;
+	hasDbtProject: boolean;
+	dbtProjectPath?: string;
+};
+
+export function getRepositories(): Repository[] | null {
+	const projectFolder = env.NAO_DEFAULT_PROJECT_PATH;
```
Bl3f, line 38: This is legacy and should not be used (I know the getConnections does it but it shouldn't), we should be given the project information from the agent when building the system-prompt - this way we are more functional and creating the needed project isolation.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+    descriptions: dict[str, str] = {}
+
+    # First pass: collect YAML descriptions and sources
+    for yaml_path in models_dir.rglob("*.yml"):
+        if _should_skip(yaml_path) or not yaml_path.is_file():
+            continue
+        try:
+            descriptions.update(parse_yaml_descriptions(yaml_path))
+        except Exception as e:
+            print(f"[dbt-indexer] Warning: failed to parse descriptions from {yaml_path}: {e}")
```
Bl3f: Use UI. (ui.py) methods to print in the console.

cli/nao_core/commands/sync/__init__.py
```diff
@@ -8,6 +8,7 @@
 from rich.console import Console
 
 from nao_core.config import NaoConfig
+from nao_core.dbt_indexer import index_all_projects
```
Bl3f: Maybe we should add in the sync command an option to deactivate this indexing behaviour (or might it be an option on the repo configuration in nao_config.yaml. Like:

is_dbt_indexed: True | False

apps/backend/src/components/system-prompt.tsx
```diff
@@ -69,6 +69,19 @@ export function SystemPrompt() {
 					Each table have files describing the table schema and the data in the table (like columns.md,
 					preview.md, etc.)
 				</ListItem>
+				<ListItem>
+					The <Italic>repos</Italic> folder contains synced git repositories with source code, data
```
Bl3f, line 75: dbt mention here should be conditional to the fact that we have a dbt project because if not it will make the agent believe we are in a dbt context.

apps/backend/src/agents/user-rules.ts
```diff
@@ -28,6 +28,58 @@ export function getUserRules(): string | null {
...
+
+			let hasDbtProject = false;
+			let dbtProjectPath: string | undefined;
+
+			if (existsSync(rootDbtProject)) {
+				hasDbtProject = true;
+				dbtProjectPath = `repos/${entry.name}`;
+			} else if (existsSync(subDbtProject)) {
+				hasDbtProject = true;
+				dbtProjectPath = `repos/${entry.name}/dbt`;
```
Bl3f, line 70: Not enough generic here as well.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+        "",
+        f"- **Project name:** {project_name}",
+        f"- **dbt project path:** {project_path}",
+        f"- **Total models:** {len(models)}",
+        f"- **Indexed at:** {now}",
+        "",
+        "---",
+    ]
+
+    for model in models:
```
Bl3f, line 373: I'm a bit concerned about the size of the generated manifest file in the end. You said you ran it on a ~700 models dbt project, the issue is that each model entry is roughly ~60 tokens, which in the end lead to a file with 37k characters - if the LLM decide to read it and add it to the context it will overflow the context window very fast.

Maybe we can think of a better structure (and not a single file to avoid this case), on the other we also have to work on a better system prompt for this, the fact that files can get out of the context window and read using a range (#193 PR does it)

apps/backend/fastapi/main.py
```diff
@@ -153,6 +171,7 @@ async def refresh_context():
         updated = provider.refresh()
 
         if updated:
+            _run_dbt_indexing("Refresh")
```
Bl3f: As you move this to the repo provider I think it's not needed anymore to have the refresh done here in this file.

cli/nao_core/dbt_indexer.py
```diff
@@ -0,0 +1,473 @@
...
+    """Scan repos/, find dbt projects, index each, write to dbt-index/."""
+    repos_path = project_folder / "repos"
+    if not repos_path.is_dir():
+        return
+
+    projects = find_dbt_projects(repos_path)
+    if not projects:
+        return
+
+    index_root = project_folder / "dbt-index"
```
Bl3f, line 63: I don't know if dbt-index is a great name, what about dbt-projects, dbt-compiled or even without an hyphen which is usually not so good for folder names?

apps/backend/src/components/system-prompt.tsx
```diff
@@ -69,6 +69,19 @@ export function SystemPrompt() {
...
+					transformation logic (dbt models), and documentation.
+				</ListItem>
+				<ListItem>
```
Bl3f: This should be conditioned because some users will not have the dbt-index, nor any dbt project.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Good point. Tested on a real 694-model project and it correctly extracted all standard `ref()` and `source()` calls. You're right that macro-generated refs (`{{ generate_join(...) }}`) won't be captured — that's an inherent limitation of static analysis without running dbt. For those cases, PR #213's compiled approach would give complete lineage. The regex covers the ~95% common case and works without requiring dbt setup.

Open to ideas on how to handle the gap — maybe a note in the manifest header flagging that macro-generated refs aren't captured?

cli/nao_core/commands/sync/__init__.py
ealexisaraujo: Done — moved to `RepositorySyncProvider.sync()` as a post-sync step. The generic sync no longer knows about dbt indexing.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Done — factorized into a single loop using `itertools.chain(models_dir.rglob("*.yml"), models_dir.rglob("*.yaml"))`.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Good catch — replaced with `_find_dbt_project_yml()` that searches up to 3 levels deep via glob patterns, finding `dbt_project.yml` regardless of the parent directory name.

cli/nao_core/dbt_indexer.py
ealexisaraujo: The YAML loops are now 1 loop (via `chain`). The SQL loop stays separate since it depends on descriptions collected from the YAML pass — these are sequential by design (YAML first collects descriptions, then SQL uses them).

apps/backend/src/agents/user-rules.ts
ealexisaraujo: Acknowledged — this ties into the broader refactor of `getRepositories()` toward injected project context. Happy to extract a shared search function in a follow-up once the architecture direction is clearer.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Done — removed all separator blocks and section headers.

apps/backend/src/agents/user-rules.ts
ealexisaraujo: Agreed — both `getRepositories()` and `getConnections()` read the filesystem directly from env vars instead of receiving project context. Kept `getRepositories()` minimally for the `hasDbtProjects` conditional in the system prompt, but the broader refactor toward injected project isolation makes total sense. Happy to contribute to that in a follow-up — would be great to align on the target architecture for how project context flows into the system prompt.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Done — all `print()` calls replaced with `UI.success()`, `UI.warn()`, `UI.error()`.

cli/nao_core/commands/sync/__init__.py
ealexisaraujo: Good idea. I think this fits well as a follow-up once the core indexer is stable — an `index_dbt: bool` flag in repo config or a CLI flag like `nao sync --skip-indexing`. It would also pair nicely with PR #213's `compile_dbt_docs` flag, giving users fine-grained control over what happens during sync.

Open to implementing this in a follow-up PR if you'd like.

apps/backend/src/components/system-prompt.tsx
ealexisaraujo: Done — dbt-specific instructions are now wrapped in `{hasDbtProjects && (...)}`, computed from `getRepositories()`. Users without dbt projects won't see these instructions, so the agent won't incorrectly assume dbt context.

apps/backend/src/agents/user-rules.ts
ealexisaraujo: Agreed — this ties into the broader `getRepositories()` refactor toward injected project context. Deferring this to a follow-up since it affects other consumers beyond our PR scope. Happy to contribute to that refactor when the direction is decided.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Great point — this is the key design question. For our test project (694 models), the manifest is ~37k chars (~42k tokens). A few options I see:

1. **Single file + grep + range reads** — if PR #193 lands, the agent can grep for a model name then read just that section. The current H3 headers (`### model_name`) are already grep-friendly. This keeps things simple and composable.
2. **Split by directory** — one manifest file per `models/` subdirectory (e.g., `staging/manifest.md`, `marts/manifest.md`). Smaller files, but more to grep across.
3. **Table-of-contents header** — a lightweight index at the top mapping model names to line numbers, so the agent reads only the TOC first and targets specific ranges.

I lean toward option 1 since it's the simplest and leverages existing infrastructure, but open to whatever works best for the overall system. What's your preferred direction?

apps/backend/fastapi/main.py
ealexisaraujo: Removed entirely. No dbt indexing hooks in FastAPI anymore — indexing happens exclusively via `nao sync` through the `RepositorySyncProvider`.

cli/nao_core/dbt_indexer.py
ealexisaraujo: Open to renaming. `dbt_projects` (underscore, no hyphen) would be more consistent with Python conventions. `dbt_index` or `dbt_compiled` are also options. What's your preference? Happy to adjust to whatever naming convention fits the project best.

apps/backend/src/components/system-prompt.tsx
ealexisaraujo: Done — this is now conditional on `hasDbtProjects`, same fix as the other dbt mention. Users without dbt projects won't see dbt-index related instructions.

cli/nao_core/dbt_indexer.py
Bl3f: Yeah, mainly good for now to do it static this way.

cli/nao_core/commands/sync/__init__.py
Bl3f: I think the best would be an option in the yaml like the compile_dbt_docs, tho naming should be uniform between the 2 flags. I don't think it should be in the CLI sync option at main level nonetheless.

Bl3f:
Last 2 changes (if I'm not mistaken) to me are:
- add the config in the nao_config.yaml, I want users to be able to deactivate this behaviour
- name the created folder `dbt/`

PR #213, opened 2026-02-16
feat(sync): pre-compile dbt docs for repo sources to improve centralized docs context

### LINKED ISSUE #190: Pre-compile dbt docs when centralized docs
As a user, if I have a centralized docs, with subdocs referencing this centralize docs in yaml, the raw files might be hard to understand for the agent. So instead, I would want to give it a compiled version of the docs - maybe catalog.json

Flamki:
## Summary
Adds optional dbt docs pre-compilation during nao sync for repository sources, so the agent can use compiled dbt artifacts (catalog.json, manifest.json) when projects use centralized docs references in YAML.

Closes #190.

## Changes
• Added repo config options:
  • compile_dbt_docs: bool (default false)
  • dbt_profiles_dir: Optional[str]

• Updated repository sync flow:
  • after clone/pull, when enabled, run:
    dbt docs generate --project-dir <repo> [--profiles-dir ...]

• Added best-effort behavior:
  • if repo is not dbt, dbt is missing, or generation fails, sync continues with warning logs

• Added unit tests for:
  • skip when disabled
  • skip when non-dbt repo
  • skip when dbt binary missing
  • execute command path with profiles dir
  • integration into repo sync success flow

• Updated docs:
  • README.md (usage + config example + behavior notes)
  • README.md (“Recent Updates” note)
  • nao_config.yaml enabling this for example dbt repo

## Why
Raw dbt YAML with centralized doc() references can be hard for the agent to interpret directly. Compiled docs artifacts provide richer, resolved metadata.

## Backward compatibility
• No behavior change unless compile_dbt_docs: true is explicitly set.
• Existing configs continue to work unchanged.

## Validation
• test_repository_provider.py
• Result: 17 passed

### DIFF
--- a/cli/nao_core/commands/sync/providers/repositories/provider.py
+++ b/cli/nao_core/commands/sync/providers/repositories/provider.py
@@ -108,9 +108,7 @@ def pre_compile_dbt_docs(repo: RepoConfig, base_path: Path) -> bool:
         result = subprocess.run(cmd, capture_output=True, text=True, check=False, timeout=300)
 
         if result.returncode != 0:
-            UI.warn(
-                f"{repo.name}: dbt docs generate failed: {result.stderr.strip() or result.stdout.strip()}"
-            )
+            UI.warn(f"{repo.name}: dbt docs generate failed: {result.stderr.strip() or result.stdout.strip()}")
             return True
 
         catalog_path = repo_path / "target" / "catalog.json"
@@ -140,7 +138,7 @@ class RepositorySyncProvider(SyncProvider):
 
     @property
     def emoji(self) -> str:
-        return "\U0001F4E6"
+        return "\U0001f4e6"
 
     @property
     def default_output_dir(self) -> str:


cli/nao_core/config/databases/__init__.py
```diff
@@ -10,6 +10,7 @@
 from .postgres import PostgresConfig
 from .redshift import RedshiftConfig
 from .snowflake import SnowflakeConfig
+from .trino import TrinoConfig
```
Bl3f: To be removed from this PR.

cli/nao_core/commands/sync/providers/repositories/provider.py
```diff
@@ -81,6 +82,64 @@ def clone_or_pull_repo(repo: RepoConfig, base_path: Path) -> bool:
...
+    """
+    try:
+        if not repo.compile_dbt_docs:
+            return True
+
+        repo_path = base_path / repo.name
+        dbt_project_file = repo_path / "dbt_project.yml"
+
+        if not dbt_project_file.exists():
+            console.print(
```
Bl3f: Use the `ui.py` methods to display stuff in the console.

cli/nao_core/commands/sync/providers/repositories/provider.py
```diff
@@ -81,6 +82,64 @@ def clone_or_pull_repo(repo: RepoConfig, base_path: Path) -> bool:
...
+                f"  [yellow]⚠[/yellow] {repo.name}: dbt not found in PATH, cannot pre-compile docs (install dbt-core)"
+            )
+            return True
+
+        cmd = [dbt_binary, "docs", "generate", "--project-dir", str(repo_path)]
+        if repo.dbt_profiles_dir:
+            cmd.extend(["--profiles-dir", repo.dbt_profiles_dir])
+
+        console.print(f"  [dim]Pre-compiling dbt docs for[/dim] {repo.name}")
+        result = subprocess.run(cmd, capture_output=True, text=True, check=False, timeout=300)
```
Bl3f, line 108: The thing I'm the most worried about is that a user will have to forward all the necessary dbt env variables to compile correctly the dbt docs (esp. if we want it in the "production" context). It adds a small complexity, but the feature is interesting, so we can add it, but might be complex for users to make it work.

README.md
```diff
@@ -183,6 +183,11 @@ For end-to-end self-hosted deployment (for example on Cloud Run with PostgreSQL)
 
 See [CONTRIBUTING.md](CONTRIBUTING.md) for development setup, commands, and guidelines.
 
+## Recent Updates
+
+- `nao sync` now supports optional dbt docs pre-compilation for repository sources (`compile_dbt_docs: true`), which
+  generates compiled artifacts such as `target/catalog.json` and `target/manifest.json` for richer agent context.
```
Bl3f, line 189: One last thing I'm afraid about is the size of the manifest.json for large project that might make it unusable and pollute the context window. Nonetheless, as it generate also the compile version of each query in the folder structure that's interesting.

PR #218, opened 2026-02-17
Feat: support Athena

vincentpwls:
Hi, I am very interested in your project. Here is a contribution to support Athena engine.
I am using Athena and Github Copilot at the moment so I plan to add Github copilot as available LLM.

### DIFF
--- a/cli/nao_core/config/databases/athena.py
+++ b/cli/nao_core/config/databases/athena.py
@@ -66,8 +66,10 @@ class AthenaConfig(DatabaseConfig):
             "s3_staging_dir": self.s3_staging_dir,
             "region_name": self.region_name,
             "schema_name": self.schema_name or "default",
-            "work_group": self.work_group,
         }
+
+        if self.work_group:
+            kwargs["work_group"] = self.work_group
 
         if self.profile_name:
             kwargs["profile_name"] = self.profile_name
--- a/cli/tests/nao_core/commands/sync/integration/test_athena.py
+++ b/cli/tests/nao_core/commands/sync/integration/test_athena.py
@@ -32,16 +32,22 @@ pytestmark = pytest.mark.skipif(
 def temp_databases():
     """Create temporary databases and populate them with test data, then clean up."""
     # Ensure S3 staging dir is available (fixture scope ensures env var is checked already via skipif)
-    db_name = f"nao_unit_tests_{uuid.uuid4().hex[:8]}"
-    another_db_name = f"nao_unit_tests_another_{uuid.uuid4().hex[:8]}"
+    db_name = f"nao_integration_tests_{uuid.uuid4().hex[:8]}"
+    another_db_name = f"nao_integration_tests_another_{uuid.uuid4().hex[:8]}"
     kwargs = {
         "s3_staging_dir": ATHENA_S3_STAGING_DIR,
         "region_name": ATHENA_REGION,
     }
-    if ATHENA_PROFILE:
-        kwargs["profile_name"] = ATHENA_PROFILE
     if ATHENA_WORKGROUP:
         kwargs["work_group"] = ATHENA_WORKGROUP
+
+    if ATHENA_PROFILE:
+        kwargs["profile_name"] = ATHENA_PROFILE
+    elif os.environ.get("ATHENA_ACCESS_KEY_ID") and os.environ.get("ATHENA_SECRET_ACCESS_KEY"):
+        kwargs["aws_access_key_id"] = os.environ.get("ATHENA_ACCESS_KEY_ID")
+        kwargs["aws_secret_access_key"] = os.environ.get("ATHENA_SECRET_ACCESS_KEY")
+        if os.environ.get("ATHENA_SESSION_TOKEN"):
+            kwargs["aws_session_token"] = os.environ.get("ATHENA_SESSION_TOKEN")
 
     conn = ibis.athena.connect(**kwargs)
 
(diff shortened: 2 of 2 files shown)


cli/nao_core/config/databases/athena.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def get_database_name(self) -> str:
+        return self.schema_name or "default"
+
+    def check_connection(self) -> tuple[bool, str]:
+        """Test connectivity to Athena."""
+        try:
+            conn = self.connect()
+            tables = conn.list_tables()
+            return True, f"Connected successfully ({len(tables)} tables found in {self.schema_name})"
```
MatLBS: Close the connection after testing

cli/nao_core/config/databases/athena.py
```diff
@@ -0,0 +1,85 @@
...
+        schema_name = ask_text("Schema Name:", default="default")
+        work_group = ask_text("Workgroup:", default="primary")
+
+        profile_name = ask_text("AWS Profile Name (optional):")
+        aws_access_key_id = None
+        aws_secret_access_key = None
+        aws_session_token = None
+
+        if not profile_name:
+            aws_access_key_id = ask_text("AWS Access Key ID (optional):")
```
MatLBS: `aws_access_key_id` and `aws_secret_access_key` should have `password=True`

See how it is done in snowflake.py for correction

MatLBS (changes requested):
Hi @sanga8 , thank you so much for contributing to the project, 

I saw you had lots of fields,
Once, I had set up an Athena connection and here were my fields:

```
- connection name (required)
- region (required)
- access key id (required)
- secret access key (required)
- session token (optionnal)
- database (optionnal)
```

Then, I tried to connect to my Athena and did not suceed...

Although, the error message I got is:
`Connection failed: Backend.do_connect() missing 1 required keyword-only argument: 's3_staging_dir'`, so ok to keep this field. 

But why `profile_name, schema_name , work_group `?

Did you suceed to connect to athena on your side ?

Thank you again !

cli/nao_core/config/databases/athena.py
vincentpwls: Sure, I couldn't find an example of best practice implementation for that on other connectors, do you mind sharing one? Thank you

vincentpwls: The Ibis Athena (stateless & REST) backend uses PyAthena under the hood so there shouldn't be a persistent connection that stays open.

cli/nao_core/config/databases/athena.py
```diff
@@ -0,0 +1,93 @@
...
+            aws_secret_access_key = ask_text("AWS Secret Access Key (optional):", password=True)
+            aws_session_token = ask_text("AWS Session Token (optional):", password=True)
```
MatLBS: My bad, only `aws_secret_access_key` should have `password=True`. 

Remove it for `aws_session_token`

vincentpwls: Are you sure about removing it for `aws_session_token` ? I believe it  is a sensitive credentials.

Adding aws_access_key_id and aws_secret_access_key as required when no profile provided 👍

MatLBS: OK to keep password for `aws_session_token`

cli/nao_core/config/databases/athena.py
```diff
@@ -0,0 +1,93 @@
...
+    @classmethod
+    def promptConfig(cls) -> "AthenaConfig":
+        """Interactively prompt the user for Athena configuration."""
+        name = ask_text("Connection name:", default="athena-prod") or "athena-prod"
+        region_name = ask_text("AWS Region:", default="us-east-1") or "us-east-1"
+        s3_staging_dir = ask_text("S3 Staging Directory (s3://...):", required_field=True) or ""
+        schema_name = ask_text("Default schema (optional):") or None
+        work_group = ask_text("Workgroup (optional):", default="primary")
+
+        profile_name = ask_text("AWS Profile Name (optional):")
```
Bl3f: The best UI here would be to use a ask_select (https://github.com/getnao/nao/blob/main/cli/nao_core/ui.py#L130) and ask for the user to either use a profile or a Access Key/Secret Access. Would make it clearer.

cli/nao_core/config/databases/athena.py
```diff
@@ -0,0 +1,93 @@
...
+    profile_name: str | None = Field(default=None, description="AWS profile name")
+    schema_name: str | None = Field(default=None, description="Athena schema name")
+    work_group: str | None = Field(default="primary", description="Athena workgroup")
+
+    @classmethod
+    def promptConfig(cls) -> "AthenaConfig":
+        """Interactively prompt the user for Athena configuration."""
+        name = ask_text("Connection name:", default="athena-prod") or "athena-prod"
+        region_name = ask_text("AWS Region:", default="us-east-1") or "us-east-1"
+        s3_staging_dir = ask_text("S3 Staging Directory (s3://...):", required_field=True) or ""
```
Bl3f, line 30: Weird that this is a required field in the prompt but with a `or ""` — also is it really required?

vincentpwls: It is required if workgroup is not provided.

But it can also be used to override the workgroup location, based on the boolean "Override client-side settings" configuration on AWS account side.

In the end both `workgroup` and `s3_staging_dir` should be configurable.

vincentpwls: I contributed a bit to dbt-athena, where we choose to have workgroup as optional and s3_staging_dir as required (but truth is s3_staging_dir is optional when you provide a workgroup)

See the inputs profile config:
https://github.com/dbt-labs/dbt-adapters/tree/main/dbt-athena#configuring-your-profile

Bl3f: Ok super clear, thank you for the details. Then I'm good to go with the same behaviour as dbt-athena.

Bl3f:
I'm not able to find a way to launch the test on my side with any AWS configuration I dont know why. I will find a way later, we can merge from now! Thank you so much and welcome as a contributor 🤗

PR #219, opened 2026-02-17
Fix/context management limits

Bl3f:
* Fix context management for Anthropic and OpenAI
* For anthropic
  * remove the thinking after 2 message
  * clear the tool output if it goes > 180k tokens but only for MCP tools, internal tools aren't clearer
* For OpenAI let the SDK do the truncation

next step is summary


apps/backend/src/agents/providers.ts
```diff
@@ -192,9 +192,26 @@ export function getProviderModelConfig<P extends LlmProvider>(provider: P, model
...
+					keep: { type: 'thinking_turns', value: 2 },
+				},
+				{
+					type: 'clear_tool_uses_20250919',
+					trigger: {
+						type: 'input_tokens',
+						value: 180_000,
+					},
+					clearToolInputs: false,
+					excludeTools: ['display_chart', 'execute_python', 'execute_sql', 'grep', 'list', 'read', 'search'],
```
Standlc, line 208: did you exclude all tools from being cleared?

Bl3f: It clears MCP tools actually, but not internal tools because as we already format them we can keep them in the context.

PR #220, opened 2026-02-17
Posthog tracking: add nao_core version and project_id to events

Standlc:


apps/backend/src/trpc/project.routes.ts
```diff
@@ -217,8 +217,8 @@ export const projectRoutes = {
 				prompt: input.prompt,
 			});
 			posthog.capture(ctx.user.id, PostHogEvent.SavedPromptCreated, {
-				projectId: ctx.project.id,
-				savedPromptId: saved.id,
+				project_id: ctx.project.id,
```
MatLBS, line 220: You should keep `projectId` to be consistent with line 215

apps/backend/src/trpc/project.routes.ts
```diff
@@ -238,8 +238,8 @@ export const projectRoutes = {
 				throw new TRPCError({ code: 'INTERNAL_SERVER_ERROR', message: 'Failed to update saved prompt' });
 			}
 			posthog.capture(ctx.user.id, PostHogEvent.SavedPromptUpdated, {
-				projectId: ctx.project.id,
-				savedPromptId: promptId,
+				project_id: ctx.project.id,
```
MatLBS, line 241: Same

apps/backend/src/services/posthog.service.ts
```diff
@@ -38,22 +39,40 @@ export class PostHogService {
...
 			});
 		} catch {
 			// Tracking should never break the backend
 		}
 	}
 
+	/** Add properties that will be shown on the PostHog person's profile. */
+	private _personProperties(): Record<string, unknown> {
+		return {
+			nao_core_version: env.NAO_CORE_VERSION, // Set `nao_core_version` in event and person properties for convenience
```
MatLBS, line 64: What if `env.NAO_CORE_VERSION` is `undefined` ? 

Will you send `undefined` three times every time ?

apps/backend/src/trpc/chat.routes.ts
```diff
@@ -48,7 +48,10 @@ export const chatRoutes = {
 		if (!agent.checkIsUserOwner(ctx.user.id)) {
 			throw new TRPCError({ code: 'FORBIDDEN', message: 'You are not allowed to stop this agent.' });
 		}
+
 		agent.stop();
+
+		posthog.capture(ctx.user.id, PostHogEvent.AgentStopped, { chat_id: input.chatId });
```
MatLBS: You should send the `project_id` here

MatLBS (commented):
Few comments, otherwise LGTM

apps/backend/src/trpc/project.routes.ts
Standlc: line 215 has nothing to do with this

apps/backend/src/services/posthog.service.ts
Standlc: `undefined` values get removed when serialized, so we just send an empty object. It's fine

MatLBS: Nice

PR #221, opened 2026-02-17
add user to project when login with social provider like Google

### LINKED ISSUE #217: bug: No project configured. Set NAO_DEFAULT_PROJECT_PATH environment variable.
## One-liner
`Project → Settings` shows a `No project configured. Set NAO_DEFAULT_PROJECT_PATH environment variable.` error for certain users.

Image

## Context
I successfully deployed an instance of `nao` chat using Google OAuth. I first signed up with Google OAuth and became the admin. Then, a colleague signed up (OAuth) and couldn't use the chat because of this error. The error was shown on the chat interface and in the settings.

Image

This doesn't happen if I create a new user (non-OAuth) as an admin. It looks like the problem is appearing only for users that signed up on their own.

MatLBS:
Fix #217 

When a user login with Google, he will be automatically added to the project and could start instantly chatting with the agent.


apps/backend/src/auth.ts
```diff
@@ -39,6 +39,8 @@ export const auth = betterAuth({
 				async after(user) {
 					// Handle first user signup: create default org and project in a single transaction
 					await orgQueries.initializeDefaultOrganizationForFirstUser(user.id);
+					// Add subsequent users to the default project
+					await orgQueries.addUserToDefaultProjectIfExists(user.id);
```
Bl3f: How can we be sure this is called only when someone signup from Google SSO? I know we deactivated the global signup, but can't we check if the signup comes from Google tho?

MatLBS:
What would be the point of deactivating the auto-join behaviour? 

We would have the same issue as currently where a user thinks he can chat with the agent but ends up getting errors.

The issue I see is that we currently have only one org and project. As Google credentials are at org scope, how will we know in which project to add the new signed-up user?

Maybe we should add it only in the organization and the admin would have to add him to a project?

apps/backend/src/auth.ts
MatLBS: Added `if (isGoogle) ...` before  `addUserToDefaultProjectIfExists`

Bl3f:
@MatLBS yes, I had in mind that you select in which project you auto-add the newcomers? Maybe this is for a later PR.

PR #223, opened 2026-02-17
close connection after checking in check_connection()

MatLBS:


MatLBS:
When I reviewed PR #218 , I remarked that we never close the connection when running a `nao debug`. 

I didn't get your point as to why it would be useless?

PR #289, opened 2026-02-18
admin can override google env variables

MatLBS:


apps/backend/src/middleware/auth.ts
```diff
@@ -16,6 +16,7 @@ declare module 'fastify' {
 
 export async function authMiddleware(request: FastifyRequest, reply: FastifyReply) {
 	const headers = convertHeaders(request.headers);
+	const auth = await authPromise;
```
Standlc: That will throw if `authPromise` throws an error. Make sure this is ok in this context

apps/backend/src/auth.ts
```diff
@@ -7,48 +7,58 @@ import { env } from './env';
...
+			additionalFields: {
+				requiresPasswordReset: { type: 'boolean', default: false, input: false },
+			},
 		},
-	},
-});
+	});
+}
+
+export let authPromise = orgQueries.getGoogleConfig().then(createAuthInstance);
```
Standlc: why didn't you leave `getAuth()`? It was cleaner than to export a promise like this.
```typescript
const getAuth = () => authPromise;
```

apps/backend/src/middleware/auth.ts
MatLBS: I think fastify will catch the error so the server won't crash

PR #290, opened 2026-02-18
Admin can choose which MCP tools for each server are available to users

### LINKED ISSUE #230: Tools choice for MCP servers
## User story
✨ As an admin, I want to choose which MCP tools for each server are available to users

## Expected behaviour
- In Settings > Project > MCP, on each MCP server entry, admin can enable/disable tools individually.
- By default, all tools are enabled.
- UI shows an active count (for example: `12/18 active`).
- Changes apply to conversations for users in that admin scope/project.
- This setting is admin-web only (not CLI) for v1.

## UI changes
- Add toggle per tool in MCP tool list, in each server MCP tool list
- Add header/status summary: `X/Y active`.
- Add lightweight token estimate indicator:
  - estimated **input tokens contributed by enabled tool definitions** (name + description/schema),
  - updates live when toggling tools.
 - Option to unselect all tools

## Technical implementation
- WIP

## Edge cases
- Newly discovered tools on a server should default to disabled
- If a disabled tool is requested indirectly, model should not see/call it.
- Handle MCP server refresh without losing saved enable/disable choices.

## Checks to validate
- Admin can toggle tools and see state persist after reload.
- Conversations only receive enabled tools in context.
- Active counter and “all active” label are accurate.
- Token estimate updates correctly when toggles change.
- Non-admin users cannot modify tool availability settings.

MatLBS:
Fix #230 

Following PR #285 made by @Flamki 

- Admin can now enable / disable tools in mcp servers
- Show how many tools are activated with an token estimation 
- Persisted project-level MCP tool settings in DB

https://github.com/user-attachments/assets/8631f21d-f6e4-41cc-98c2-ca2b85666464

### DIFF
--- a/apps/backend/src/services/mcp.service.ts
+++ b/apps/backend/src/services/mcp.service.ts
@@ -5,6 +5,7 @@ import { existsSync, readFileSync, watch } from 'fs';
 import { createRuntime, type Runtime, ServerDefinition, ServerToolInfo } from 'mcporter';
 import { join } from 'path';
 
+import * as mcpConfigQueries from '../queries/project.queries';
 import { mcpJsonSchema, McpServerConfig, McpServerState } from '../types/mcp';
 import { retrieveProjectById } from '../utils/ai';
 import { prefixToolName, removePrefixToolName, sanitizeTools } from '../utils/tools';
@@ -20,6 +21,7 @@ export class McpService {
 	private _runtime: Runtime | null = null;
 	private _failedConnections: Record<string, string> = {};
 	private _toolsToServer: Map<string, string> = new Map();
+	private _projectId: string | null = null;
 	public cachedMcpState: Record<string, McpServerState> = {};
 
 	constructor() {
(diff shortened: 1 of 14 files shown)


apps/backend/src/db/pgSchema.ts
```diff
@@ -136,6 +136,7 @@ export const project = pgTable(
 		slackBotToken: text('slack_bot_token'),
 		slackSigningSecret: text('slack_signing_secret'),
 		agentSettings: jsonb('agent_settings').$type<AgentSettings>(),
+		disabledMcpTools: text('disabled_tools').array().notNull().default([]),
```
Bl3f: doing it this way make every new tool added by the server activated by default right?

apps/backend/src/services/mcp.service.ts
```diff
@@ -64,38 +73,51 @@ export class McpService {
...
+			Object.values(this.cachedMcpState)
+				.flatMap((server) => server.tools)
+				.filter((tool) => !tool.disabled)
+				.map((tool) => tool.name),
+		);
 
-				// If it's an AI SDK schema wrapper with jsonSchema getter
-				if (inputSchema && typeof inputSchema === 'object' && 'jsonSchema' in inputSchema) {
-					const originalJsonSchema = inputSchema.jsonSchema;
+		const sanitizedMcpTools = Object.fromEntries(
```
Bl3f: Can we move this to a method so it's easier to read?

PR #294, opened 2026-02-18
feat(chat): add conversation compaction/summary on context overflow

### LINKED ISSUE #225: Compaction / summary
## User story

✨ As a user I want to chat indefinitely regardless of the max context possible by the underlying LLM.

## Expected behaviour

When we reach maximum context window a compaction mecanism should summarise the chat so the chat can continue. The UI should change and display that a summary was generated (in this first version we don't care about showing the summary).

When a summary has been generated we then send only what comes after the last summary generated (incl. the summary generated).

## Technical implementation

* It seems that Vercel AI SDK already contains a summary mechanism for Anthropic, for other providers we might need to do it differently (up to discussion)
* We might test a few prompts to converge on the best solution, see attached the one that we use in nao IDE.
* The summary messages might be a new type of messages in the messages/messagePart tables

## UI changes

* You can reuse the Simple tool to display that a summary have been generated.
Image

## Edge cases
* If a summary fails we should retry once and then raise and display a nice error to the user (saying sorry obv) - and fallback to the current behaviour where conversation is truncated (instead of compacted)

## Checks to validate

- [ ]  When reaching max context window on a chat it should trigger the summary
- [ ] P1: works with Anthropic, OpenAI and Gemini
- [ ] P2: works with Mistral and Openrouter

## Attached
Prompt we already used in the past
```
You have been working on the task described above but have not yet completed it. Write a continuation summary that will allow you (or another instance of yourself) to resume work efficiently in a future context window where the conversation history will be replaced with this summary. Your summary should be structured, concise, and actionable. Include:

1. Task Overview
The user's core request and success criteria
Any clarifications or constraints they specified

2. Current State
What has been completed so far
Files or tables created, modified, or analyzed (with paths if relevant)
Key outputs or artifacts produced

3. Important Discoveries
Technical constraints or requirements uncovered
Decisions made and their rationale
Errors encountered and how they were resolved
What approaches were tried that didn't work (and why)

4. Next Steps
Specific actions needed to complete the task
Any blockers or open questions to resolve
Priority order if multiple steps remain

5. Context to Preserve
User preferences or style requirements
Domain-specific details that aren't obvious
Any promises made to the user

Be concise but complete—err on the side of including information that would prevent duplicate work or repeated mistakes. Write in a way that enables immediate resumption of the task.
```

Flamki:
Closes #225

## Summary
Implements chat compaction when context limits are reached so conversations can continue without hard failure.

## What’s included
- Detect context-window overflow during agent streaming.
- On overflow, generate a compact summary of prior conversation state.
- Retry once with a smaller summary window if summary generation fails.
- If summary still fails, fallback to truncated-context behavior (non-blocking).
- Persist generated summary as a synthetic assistant message.
- For subsequent turns, only include messages from the last summary onward.
- Emit a stream data event to notify frontend when compaction occurred.
- Show a UI notice in chat when a summary is generated.

## Files changed
- `apps/backend/src/services/agent.service.ts`
- `apps/backend/src/types/chat.ts`
- `apps/frontend/src/hooks/use-agent.ts`
- `apps/frontend/src/components/chat-messages.tsx`

## Validation
- `npm run lint -w @nao/backend`
- `npm run lint -w @nao/frontend`

Both pass locally.


apps/frontend/src/hooks/use-agent.ts
```diff
@@ -78,7 +80,13 @@ export const useAgent = (): AgentHelpers => {
 					};
 				},
 			}),
-			onData: ({ data: newChat }) => {
+			onData: ({ data }) => {
```
Standlc: use the type property to differentiate between the two events and remove the typecasting

Standlc (changes requested):
Hi, thanks for this PR.

1) When it comes to the code, I think it could be cleaner and better factorized. The `AgentManager` is starting to get large and difficult to follow without getting in the details of every method. To split the code, the conversation summary should definitely be implemented as another service.

2) When it comes to the compaction trigger, we had previously  implemented memory compaction based on token count (could be approximated to `number_of_totol_characters / 4`) which I find cleaner than through error detection and retries. Memory compaction should not **solely** be triggered by the context window being full, but should be configurable so we can pick a specific threshold before conversation compaction.

 3) Placing the summary content as a `assistant` message is not only a "trick" to bypass the ai sdk's messages' types, but it might also not be the best block to use. We could instead pass it as a `system` or `user` message which might yield better results. Maybe we could find a way to expand the possible message types, at least in the database.

4) The compaction could be needed while the agent loop is running, not necessarily when the user sends a message. From what I read, that was not implemented. 

5) The frontend should be able to show the summary not only when it is triggers and the backend streams it as an event, but also when the chat is reloaded (on page refresh).

In conclusion:
- better factorization, code splitting and less repetition (e.g. `this._agent.stream()`)
- trigger compaction based on a configurable token count and while the agent's loop is running
- add some JSDocs to some high level functions (no params or output) to explain the overall strategy for compaction)

Thanks again!

PR #323, opened 2026-02-21
Add memory opt-out option enforced at user and project level, setting…

Standlc:
### Agent Memory Settings
- user can opt-out of memory
- admin can opt-out of memory for the entire project and its users
- user can see their memories, edit or delete them
- refactoring of settings + improved UI & routing logic (split the project's settings into sub routes (`/settings/project/agent`) instead of query `/settings/project?view=agent`)
- move settings components under `components/settings/`

https://github.com/user-attachments/assets/c9cc11fb-f797-4255-9f39-4175b534e962

### DIFF
--- a/apps/backend/src/queries/user.queries.ts
+++ b/apps/backend/src/queries/user.queries.ts
@@ -2,6 +2,7 @@ import { count, eq } from 'drizzle-orm';
 
 import s, { NewAccount, NewProjectMember, NewUser, User } from '../db/abstractSchema';
 import { db } from '../db/db';
+import { takeFirstOrThrow } from '../utils/queries';
 
 export const get = async (identifier: { id: string } | { email: string }): Promise<User | null> => {
 	const condition = 'id' in identifier ? eq(s.user.id, identifier.id) : eq(s.user.email, identifier.email);
@@ -21,12 +22,11 @@ export const modify = async (id: string, name: string): Promise<void> => {
 };
 
 export const getMemoryEnabled = async (userId: string): Promise<boolean> => {
-	const [user] = await db
-		.select({ memoryEnabled: s.user.memoryEnabled })
-		.from(s.user)
-		.where(eq(s.user.id, userId))
-		.execute();
-	return user?.memoryEnabled ?? true;
+	const user = await takeFirstOrThrow(
+		db.select({ memoryEnabled: s.user.memoryEnabled }).from(s.user).where(eq(s.user.id, userId)).execute(),
+	);
+
+	return user.memoryEnabled;
 };
 
 export const setMemoryEnabled = async (userId: string, memoryEnabled: boolean): Promise<void> => {
--- a/apps/backend/src/utils/queries.ts
+++ b/apps/backend/src/utils/queries.ts
@@ -10,3 +10,17 @@ export function conflictUpdateSet<TTable extends SQLiteTable>(
 		columns.map((k) => [k, sql.raw(`excluded.${(table[k] as Column).name}`)]),
 	) as SQLiteUpdateSetSource<TTable>;
 }
+
+/**
+ * Returns the first row of a query promise or throws an error if nothing is returned.
+ */
+export async function takeFirstOrThrow<T>(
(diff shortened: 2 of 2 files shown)


apps/frontend/src/queries/use-memories.ts
```diff
@@ -0,0 +1,47 @@
+import { useMutation, useQuery } from '@tanstack/react-query';
```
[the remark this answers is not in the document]
Standlc: false

apps/frontend/src/components/sidebar-settings-nav.tsx
```diff
@@ -14,34 +18,29 @@ const settingsNavItems = [
...
-							isActive
-								? 'bg-sidebar-accent text-foreground font-medium'
-								: 'text-muted-foreground hover:bg-sidebar-accent hover:text-foreground',
+							'flex items-center gap-3 px-3 py-2 text-sm rounded-md transition-colors whitespace-nowrap',
 						)}
+						activeProps={{
+							className: cn('bg-sidebar-accent text-foreground font-medium'),
+						}}
+						inactiveProps={{
+							className: cn('hover:bg-sidebar-accent hover:text-foreground'),
```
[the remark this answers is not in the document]
Standlc: that's on purpose

apps/backend/src/queries/memory.ts
```diff
@@ -28,6 +28,25 @@ export const getUserMemories = async (userId: string, excludeChatId?: string): P
 	return memories;
 };
 
+export const getIsMemoryEnabledForUserAndProject = async (userId: string, projectId: string): Promise<boolean> => {
```
Bl3f, line 31: Should this file named memory.queries.ts?

apps/backend/src/queries/user.queries.ts
```diff
@@ -20,6 +20,19 @@ export const modify = async (id: string, name: string): Promise<void> => {
 	await db.update(s.user).set({ name }).where(eq(s.user.id, id)).execute();
 };
 
+export const getMemoryEnabled = async (userId: string): Promise<boolean> => {
+	const [user] = await db
+		.select({ memoryEnabled: s.user.memoryEnabled })
+		.from(s.user)
+		.where(eq(s.user.id, userId))
+		.execute();
+	return user?.memoryEnabled ?? true;
```
Bl3f: As we already have a default in db schema, why do we need to default here inline?

apps/backend/src/trpc/project.routes.ts
```diff
@@ -283,4 +283,16 @@ export const projectRoutes = {
...
 			return projectQueries.updateAgentSettings(ctx.project.id, input);
 		}),
+
+	getMemorySettings: projectProtectedProcedure.query(async ({ ctx }) => {
+		const memoryEnabled = await projectQueries.getProjectMemoryEnabled(ctx.project.id);
+		return { memoryEnabled };
+	}),
+
+	updateMemorySettings: adminProtectedProcedure
+		.input(z.object({ memoryEnabled: z.boolean() }))
```
Bl3f, line 293: Question - we never type the output of trpc methods, shouldn't we do it?

apps/backend/src/queries/memory.ts
Standlc: I'm starting to rollback the `<domain>.<role>` file naming convention. I initially thought there would be benefits to suffixing files with their `<role>` for search and organization but I'm realizing it's useless when they're all under the `<role>/` directory already.

apps/backend/src/queries/user.queries.ts
Standlc: this query might return no rows if the user doesn't exist. I've fixed by adding a utility function that throws an error if that's the case.

apps/backend/src/trpc/project.routes.ts
Standlc: Yes, but at the same time creating endpoints that are implicitly typed is what makes tRPC convenient.

PR #329, opened 2026-02-21
Edit and resend user message

Standlc:


apps/backend/src/db/pg-schema.ts
```diff
@@ -197,6 +197,7 @@ export const chatMessage = pgTable(
 		errorMessage: text('error_message'),
 		llmProvider: text('llm_provider').$type<LlmProvider>(),
 		llmModelId: text('llm_model_id'),
+		supersededAt: timestamp('superseded_at'),
```
Bl3f, line 200: Why did you decided to call it this way? What about updated_at in this case? 

Just in term of behaviour, did you consider storing also a parent id when superseded so we can loop in the frontend over all the version (and responses)? Because here if i'm not mistaken we loose the track of the original message

apps/backend/src/handlers/agent.ts
```diff
@@ -0,0 +1,105 @@
...
+}
+
+interface HandleAgentMessageResult {
+	chatId: string;
+	isNewChat: boolean;
+	modelId: string;
+	stream: ReadableStream;
+}
+
+export const handleAgentRoute = async (opts: HandleAgentMessageInput): Promise<HandleAgentMessageResult> => {
```
Bl3f, line 21: Handler vs. service?

apps/backend/src/queries/chat.queries.ts
```diff
@@ -117,72 +117,93 @@ export const getChatOwnerId = async (chatId: string): Promise<string | undefined
 	return result?.userId;
 };
 
-export const createChat = async (newChat: NewChat, message: UIMessage): Promise<UIChat> => {
-	return db.transaction(async (t): Promise<UIChat> => {
+/** Suppress all messages from a given message id onwards (won't be used in the conversation anymore). */
```
Bl3f: It does really suppress, it's more a soft delete, but update should be better than say suppress

apps/backend/src/handlers/agent.ts
Standlc: `handlers` are for handling specific http routes (except for Posthog tracking).
`services` are standalone objects that provide something (agent, memory, email, etc).

apps/backend/src/db/pg-schema.ts
Standlc: I wanted to implement the "Is a message still in the conversation?" logic in the most straightforward way possible. `superseeded_at` is added to all messages after the edited message and including it.
Later, if we wanted to let the user switch branches in a chat, I think we'd need to store a `parent_id` on each message, but that's out of scope for this PR.

PR #345, opened 2026-02-23
Fix #343: save step-start message parts to convert UI messages to mod…

### LINKED ISSUE #343: Agent tool calls not persisted in context window

Standlc:
…el messages correctly


apps/backend/src/utils/chat-message-part-mappings.ts
```diff
@@ -77,7 +83,7 @@ export const convertDBPartToUIPart = (part: DBMessagePart, provider?: LlmProvide
 			rawInput: part.toolRawInput as any,
 			output: part.toolOutput as any,
 			errorText: part.toolErrorText as any,
-			providerExecuted: provider === 'anthropic',
+			providerExecuted: false,
```
Bl3f, line 86: should even remove it from here?

PR #360, opened 2026-02-25
Improve user instructions and profile memory system

Standlc:
- add recent conversation messages to the extractor LLM for better contextualization
- improve extractor LLM output schema to generate less tokens
- prevent memory deletion and instead force replacements with new memories
- minor improvements and fixes in the markdown components
- add llm_inference table to track AI usage

https://github.com/user-attachments/assets/f62ca244-e52a-4f3e-ace0-91421da4a65c

### DIFF
--- a/apps/backend/src/agents/memory/memory-extractor-llm.ts
+++ b/apps/backend/src/agents/memory/memory-extractor-llm.ts
@@ -1,5 +1,4 @@
 import { generateText, LanguageModelUsage, ModelMessage, Output } from 'ai';
-import { inspect } from 'util';
 
 import { renderMemoryExtractionUserMessage } from '../../components/ai';
 import { MEMORY_EXTRACTION_SYSTEM_PROMPT } from '../../components/ai/memory-system-prompt';
@@ -8,6 +7,7 @@ import { UIMessage } from '../../types/chat';
 import type { ExtractorLLMOutput } from '../../types/memory';
 import { ExtractorOutputSchema } from '../../types/memory';
 import { findLastUserMessage, getLastUserMessageText, joinAllTextParts } from '../../utils/ai';
+import { debugMemory } from '../../utils/debug';
 import { truncateMiddle } from '../../utils/utils';
 import { type ProviderModelResult } from '../providers';
 
@@ -15,8 +15,6 @@ interface MemoryExtractorResult {
 	output: ExtractorLLMOutput;
 	usage: LanguageModelUsage;
 }
-
-const DEBUG = false;
 
 const CONVERSATION_MESSAGE_LIMIT = 17;
 const MESSAGE_CHAR_LIMIT = 1_250;
@@ -37,7 +35,7 @@ export class MemoryExtractorLLM {
 
 		const modelMessages = this._buildModelMessages(memories, uiMessages);
 
-		this._printDebug('--- modelMessages ---', modelMessages);
+		debugMemory('modelMessages', modelMessages);
 
 		const { output, usage } = await generateText({
 			...this.model,
@@ -46,7 +44,7 @@ export class MemoryExtractorLLM {
 			maxOutputTokens: 4000,
 		});
 
-		this._printDebug('--- output ---', output);
+		debugMemory('output', output);
 
 		return { output, usage };
 	}
@@ -82,11 +80,5 @@ export class MemoryExtractorLLM {
 
 	private _buildUserMemoryMessage(memories: DBMemory[]): ModelMessage {
 		return { role: 'user', content: renderMemoryExtractionUserMessage(memories) };
-	}
-
-	private _printDebug(message: string, data: unknown): void {
-		if (DEBUG) {
-			console.log(message, inspect(data, { showHidden: false, depth: null, colors: true }), message);
-		}
 	}
 }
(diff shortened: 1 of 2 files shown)


apps/backend/src/agents/memory/output-schema.ts
```diff
@@ -0,0 +1,37 @@
+import { z } from 'zod/v4';
```
MatLBS: move this file into `types` folder ?

apps/backend/src/agents/memory/memory-extractor-llm.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	private _buildConversationMessages(uiMessages: UIMessage[]): ModelMessage[] {
+		const recent = uiMessages.slice(-CONVERSATION_MESSAGE_LIMIT);
+		const modelMessages: ModelMessage[] = [];
+
+		for (const message of recent) {
+			if (message.role !== 'user' && message.role !== 'assistant') {
+				continue;
+			}
+			const text = extractAllTextParts(message).slice(0, MESSAGE_CHAR_LIMIT);
```
MatLBS: `.slice(0, MESSAGE_CHAR_LIMIT)` silently truncates messages. 

Could we avoid truncating in the middle of a sentence for example ?

apps/backend/src/queries/memory.ts
MatLBS: Looks like this function is unused, remove it ?

apps/backend/src/queries/memory.ts
```diff
@@ -62,12 +64,38 @@ export const upsertMemories = async (memories: DBNewMemory[]): Promise<void> =>
 		.execute();
 };
 
-export const deleteMemories = async (memoryIds: string[]): Promise<void> => {
-	if (memoryIds.length === 0) {
+export const upsertAndSupersedeMemories = async (
+	memories: (DBNewMemory & { supersedesId?: string | null })[],
+): Promise<void> => {
+	if (memories.length === 0) {
```
MatLBS: You should check for this before, in `/services/memory.ts`

apps/backend/src/types/memory.ts
```diff
@@ -1,7 +1,14 @@
+import { z } from 'zod/v4';
+
 import { DBMemory } from '../db/abstractSchema';
-import { MemoryCategory } from '../utils/memory';
+import { UIMessage } from './chat';
 import { LlmProvider } from './llm';
 
+/** Categories of memories that can be extracted from user messages. Ordered by priority. */
+export const MEMORY_CATEGORIES = ['global_rule', 'personal_fact'] as const;
```
Bl3f, line 8: memory categories could be directly mapped to ExtractorOutputSchema?

so we see more directly what maps global_rule > user_instructions and personal_fact -> user_profile

```
export const ExtractorOutputSchema = z.object({
	user_instructions: z.array(UserInstructionSchema).nullable(),
	user_profile: z.array(UserProfileSchema).nullable(),
});
```

or rename the categories?

apps/backend/src/agents/memory/memory-extractor-llm.ts
```diff
@@ -0,0 +1,91 @@
...
+import { getLastUserMessageText, joinAllTextParts } from '../../utils/ai';
+import { truncateMiddle } from '../../utils/utils';
+import { type ProviderModelResult } from '../providers';
+
+interface MemoryExtractorResult {
+	output: ExtractorLLMOutput;
+	usage: LanguageModelUsage;
+}
+
+const DEBUG = false;
```
Bl3f: use an env variable? The only annoying part of this is that if we forget to put it back to false after debugging.

apps/backend/src/agents/memory/memory-extractor-llm.ts
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		return modelMessages;
+	}
+
+	private _buildUserMemoryMessage(memories: DBMemory[]): ModelMessage {
+		return { role: 'user', content: renderMemoryExtractionUserMessage(memories) };
+	}
+
+	private _printDebug(message: string, data: unknown): void {
```
Bl3f: Could move to a util.

apps/backend/src/agents/memory/memory-extractor-llm.ts
```diff
@@ -0,0 +1,91 @@
...
+		}
+
+		const modelMessages = this._buildModelMessages(memories, uiMessages);
+
+		this._printDebug('--- modelMessages ---', modelMessages);
+
+		const { output, usage } = await generateText({
+			...this.model,
+			output: Output.object({ schema: ExtractorOutputSchema }),
+			messages: modelMessages,
```
Bl3f, line 43: We can eventually cache for Anthropic - but for Haiku caching works only if token count > 4096 - not sure we pass the threshold often?

Bl3f (approved):
LGTM, small comment about typing

apps/backend/src/types/memory.ts
Standlc: Memory categories are for the database, if we rename them we either need to do a migration or keep the old types to be able to manipulate previous memories.

PR #362, opened 2026-02-25
Feat/ssr charts for slack

### LINKED ISSUE #199: Send charts to Slack thread
As a user, when I ask a question to nao bot in slack, I want to see any chart that was sent in the conversation

### LINKED ISSUE #356: Add a download button
When a chart is displayed, I would like to download it as a png file

Image

MatLBS:
Fix #199 

https://github.com/user-attachments/assets/938343e2-0529-476c-8454-dbcfae1ebf0f

Also fix #356 

-> implemented a download button 

Capture d’écran 2026-02-26 à 09 08 19

### DIFF
--- a/apps/backend/src/db/pg-schema.ts
+++ b/apps/backend/src/db/pg-schema.ts
@@ -408,7 +408,7 @@ export const message_part_chart_image = pgTable('chart_image', {
 	id: text('id')
 		.$defaultFn(() => crypto.randomUUID())
 		.primaryKey(),
-	toolCallId: text('tool_call_id').unique(),
+	toolCallId: text('tool_call_id').notNull().unique(),
 	data: text('data').notNull(),
 	createdAt: timestamp('created_at').defaultNow().notNull(),
 });
--- a/apps/backend/src/routes/chart.ts
+++ b/apps/backend/src/routes/chart.ts
@@ -9,7 +9,7 @@ const paramsSchema = z.object({
 });
 
 export const chartRoutes = async (app: App) => {
-	app.get('/:chatId/:chartId.png', { schema: { params: paramsSchema } }, async (request, reply) => {
+	app.get('/:chatId/:chartid.png', { schema: { params: paramsSchema } }, async (request, reply) => {
 		const { chartid } = request.params;
 
 		const imageData = await getChartById(chartid);
--- a/apps/backend/migrations-postgres/0018_added_message_part_chart_image.sql
+++ b/apps/backend/migrations-postgres/0018_added_message_part_chart_image.sql
@@ -1,6 +1,6 @@
 CREATE TABLE "chart_image" (
 	"id" text PRIMARY KEY NOT NULL,
-	"tool_call_id" text,
+	"tool_call_id" text NOT NULL,
 	"data" text NOT NULL,
 	"created_at" timestamp DEFAULT now() NOT NULL,
 	CONSTRAINT "chart_image_tool_call_id_unique" UNIQUE("tool_call_id")
--- a/apps/backend/migrations-postgres/meta/0018_snapshot.json
+++ b/apps/backend/migrations-postgres/meta/0018_snapshot.json
@@ -1,5 +1,5 @@
 {
-  "id": "f352422f-a7e9-473d-8469-5d825693d28f",
+  "id": "7f6c4561-06ae-4a2b-bb30-0073a40f8ca7",
   "prevId": "2efd7482-4f01-48b1-a00a-2c8127fd35f3",
   "version": "7",
   "dialect": "postgresql",
@@ -1023,7 +1023,7 @@
           "name": "tool_call_id",
           "type": "text",
           "primaryKey": false,
-          "notNull": false
+          "notNull": true
         },
         "data": {
           "name": "data",
--- a/apps/backend/migrations-postgres/meta/_journal.json
+++ b/apps/backend/migrations-postgres/meta/_journal.json
@@ -131,7 +131,7 @@
     {
       "idx": 18,
       "version": "7",
-      "when": 1772122539765,
+      "when": 1772124793182,
       "tag": "0018_added_message_part_chart_image",
       "breakpoints": true
     }
--- a/apps/backend/migrations-sqlite/0018_added_message_part_chart_image.sql
+++ b/apps/backend/migrations-sqlite/0018_added_message_part_chart_image.sql
@@ -1,6 +1,6 @@
 CREATE TABLE `chart_image` (
 	`id` text PRIMARY KEY NOT NULL,
-	`tool_call_id` text,
+	`tool_call_id` text NOT NULL,
 	`data` text NOT NULL,
 	`created_at` integer DEFAULT (cast(unixepoch('subsecond') * 1000 as integer)) NOT NULL
 );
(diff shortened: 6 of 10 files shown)


apps/shared/package.json
```diff
@@ -16,6 +17,18 @@
 	"dependencies": {
 		"zod": "^4.2.1"
 	},
+	"peerDependencies": {
```
Standlc: These shouldn't be in the peer dependencies, remove them and install react and recharts with `npm i` again

apps/shared/package.json
```diff
@@ -5,7 +5,8 @@
 	"exports": {
 		".": "./src/index.ts",
 		"./posthog": "./src/posthog.ts",
-		"./tools": "./src/tools/index.ts"
+		"./tools": "./src/tools/index.ts",
+		"./chart-builder": "./src/chart-builder.tsx"
```
Standlc: remove that and `export * from './chart-builder'` in the `index.ts` instead

apps/backend/src/queries/chat.queries.ts
Standlc, line 1: move these queries to `chart-image.ts`

apps/backend/src/queries/chat.queries.ts
```diff
@@ -324,6 +324,23 @@ const caseInsensitiveLike = (column: Parameters<typeof like>[0], pattern: string
...
+export const saveChart = async (toolCallId: string, data: string): Promise<void> => {
+	await db
+		.insert(s.chartImage)
+		.values({ toolCallId, data })
+		.onConflictDoUpdate({ target: s.chartImage.toolCallId, set: { data } })
+		.execute();
+};
+
+export const getChart = async (toolCallId: string): Promise<string | null> => {
+	const [result] = await db
```
Standlc: You can wrap your query in `takeFirstOrThrow()` to not return a nullable value

apps/backend/src/routes/chart.ts
```diff
@@ -0,0 +1,17 @@
+import type { App } from '../app';
+import * as chatQueries from '../queries/chat.queries';
+import { HandlerError } from '../utils/error';
+
+export const chartRoutes = async (app: App) => {
+	app.get('/:toolCallId', async (request, reply) => {
```
Standlc: validate your input!

apps/backend/src/services/slack.service.ts
```diff
@@ -126,79 +166,124 @@ export class SlackService {
...
+		if (part.state !== 'output-available' || state.renderedChartIds.has(part.toolCallId)) {
+			return;
+		}
+		const data = state.sqlOutputs.get(part.input.query_id);
+		if (!data) {
+			return;
+		}
+		try {
+			const png = generateChartImage({ config: part.input, data });
+			await chatQueries.saveChart(part.toolCallId, png.toString('base64'));
```
Standlc: Not sure you need to store the image in base 64. Try and see if it still works when you store the raw image and send it as is.

apps/backend/src/services/slack.service.ts
```diff
@@ -126,79 +166,124 @@ export class SlackService {
...
-				{
-					text: this._text,
-				},
-			);
-			return createdChat.id;
+			this._currentConv.blocks[this._textBlockIndex] = block;
 		}
 	}
+
+	private async _postChartImageBlock(toolCallId: string): Promise<void> {
```
Standlc: Why don't you use a uuid? Tool call ids are very large thus not the safest when it comes to uniqueness collisions. Use the `id` of the newly inserted chart image instead.

apps/backend/src/utils/slack.ts
```diff
@@ -15,3 +17,14 @@ export const addButtonStopBlock = () => {
 		],
 	};
 };
+
+export const addImageBlock = (imageUrl: string): ImageBlock => ({
```
Standlc: `createImageBlock` is more consistent

apps/frontend/src/components/tool-calls/display-chart.tsx
```diff
@@ -22,6 +25,25 @@ export const DisplayChartToolCall = ({
...
+			},
+		}),
+	);
+
+	const handleDownload = async () => {
+		if (!config) {
+			return;
+		}
+		const image = await downloadImage.mutateAsync({ config, data: filteredData });
+		const link = document.createElement('a');
```
Standlc: kinda hacky, why don't you have the backend send the correct headers so that the browser will automatically download the image?
```ts
res.setHeader('Content-Disposition', 'attachment; filename="yourfile.csv"');
res.setHeader('Content-Type', 'application/octet-stream'); // or a specific MIME type
```

apps/backend/src/db/pg-schema.ts
```diff
@@ -362,3 +362,9 @@ export const memories = pgTable(
 	},
 	(t) => [index('memories_userId_idx').on(t.userId), index('memories_chatId_idx').on(t.chatId)],
 );
+
+export const chartImage = pgTable('chart_image', {
```
Bl3f: would call the table message_part_chart_image

apps/backend/src/trpc/chart.routes.ts
```diff
@@ -0,0 +1,27 @@
+import { z } from 'zod/v4';
+
+import { generateChartImage } from '../components/generate-chart';
+import { projectProtectedProcedure } from './trpc';
+
+export const chartRoutes = {
+	download: projectProtectedProcedure
+		.input(
+			z.object({
```
Bl3f, line 10: why do we need this to download a chart, this way it requires the client to send the data and the whole configuration. Can't we just use the tool call id to download the image? And in this case we need to make sure the user has the rights on the tool call.

apps/backend/src/routes/chart.ts
```diff
@@ -0,0 +1,17 @@
+import type { App } from '../app';
+import * as chatQueries from '../queries/chat.queries';
+import { HandlerError } from '../utils/error';
+
+export const chartRoutes = async (app: App) => {
```
Bl3f, line 11: should be named differently I think because it's a bit confusing that we have two chartRoutes (one trpc and the other fastify).

Also the endpoint should not be /api/chart as it's not an API, could be something like `/c/:chatId/:chartId` so we hide it more in a non english word

apps/backend/src/routes/chart.ts
```diff
@@ -0,0 +1,17 @@
...
+	app.get('/:toolCallId', async (request, reply) => {
+		const { toolCallId } = request.params as { toolCallId: string };
+
+		const imageData = await chatQueries.getChart(toolCallId);
+		if (!imageData) {
+			throw new HandlerError('NOT_FOUND', 'Chart image not found');
```
Bl3f, line 17: Should throw and less leaky error like 'NOT_FOUND' generic or even just 'ERROR'

apps/backend/src/trpc/chart.routes.ts
Bl3f: Also can't we also then re-use the fastify url so we don't manage 2 endpoints.

Bl3f (changes requested):
A few things that we could change

apps/frontend/src/components/tool-calls/display-chart.tsx
MatLBS: I tried but the request has to be a GET Method. 

As I need to send the config to create the image, it cannot work

PR #377, opened 2026-02-28
Feature/agent conversation compaction

Standlc:
## Conversation Compaction

Adds automatic conversation compaction to prevent the agent from running out of context on long conversations.

### Strategy

When estimated token usage (messages + tools + max output) exceeds **75% of the model's context window**, the agent summarizes the conversation history before sending the next request to the LLM.

Only the history *before the current user turn* is compacted — the current turn is always kept verbatim. This gives the model a clean handoff: it receives a summary of prior context followed by the exact message it's responding to.

### How it works

1. **Threshold check** — before each agent step, token usage is estimated. If below the threshold, nothing happens.
2. **Summarization** — messages between the first non-system message and the last user message are extracted and sent to a cheap extractor model (Haiku, Flash, etc.) with a compaction prompt.
3. **In-place replacement** — the summarized messages are replaced with a single assistant message containing the summary wrapped in `<conversation-summary>` XML tags, which the agent's system prompt teaches it to interpret.
4. **Persistence** — a `data-compaction` part is appended to the assistant message so the compaction is stored in the DB and surfaced in the UI.
5. **Rehydration** — on subsequent turns, `useLastCompaction` reconstructs context from the stored compaction marker: `[summary message, ...messages from last user turn onward]`.

### Context window

Each model in `LLM_PROVIDERS` now declares its own `contextWindow`. Both the compaction threshold and the extractor model's input budget use the actual model context window rather than a hardcoded 200k.

### DIFF
--- a/apps/backend/src/components/ai/compaction-system-prompt.tsx
+++ b/apps/backend/src/components/ai/compaction-system-prompt.tsx
@@ -12,8 +12,8 @@ export const COMPACTION_SYSTEM_PROMPT = renderToMarkdown(
 			This summary will replace older messages to keep the conversation within context limits while preserving all
 			important information to provide continuity for future context.
 			<Br />
-			The purpose of this summary is in a future context Write down anything that would be helpful, including the
-			state, next steps, learnings etc. Emphasize the most recent messages and assistant tool calls.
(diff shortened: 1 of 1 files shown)


Bl3f:
Thanks! I'll have a look 😊

PR #379, opened 2026-03-01
Kpi cards

cainemerrick98:
# Summary
- Added a display kpi tool to the agent. 
- Implementation similar to dipslay chart (the agent provides a query id and a display name for the api). The agent may pass more than one kpi to the tool and these kpis will be rendered in one kpi banner.
- Ensured stories also support kpis.

# Examples
## KPI Cards in Chat
image

## KPI Cards in Story Side Panel
image

## KPI Cards in Story Side Panel Editor
image

## KPI Cards in Story Thumbnail
image

## KPI Cards In Story Viewer
image

### DIFF
--- a/apps/shared/src/chart-builder.tsx
+++ b/apps/shared/src/chart-builder.tsx
@@ -95,7 +95,7 @@ function buildKpiCard(props: ResolvedProps) {
 	const { data, series } = props;
 
 	const kpis = series.map((s) => {
-		const value = data[0][s.data_key];
+		const value = data[0]?.[s.data_key];
 		return { value, displayName: s.label ?? s.data_key };
 	});
 


Bl3f:
Hello! Thank you for your first contribution!

KPI card should not be a tool but more a chart subtype of the current charts. 

If we create tools for different kind of visualisation it gonna add so much noise to the context windows. 

Second point is: I think the style of the card should be improved, card looks like way too much like the default vibe coded dashboards cards we have on every Lovable apps.

apps/backend/src/agents/tools/display-chart.ts
```diff
@@ -16,7 +16,7 @@ export default tool<displayChart.Input, displayChart.Output>({
 
 		// Validate pie charts have exactly one series
 		if (chartType === 'pie' && series.length !== 1) {
-			return { _version: '1', success: false, error: 'Pie charts require exactly one series.' };
+			return { _version: '1', success: false, error: `${chartType} charts require exactly one series.` };
```
Bl3f: Change not needed as the if is chartType === 'pie'

apps/frontend/src/components/tool-calls/display-chart.tsx
```diff
@@ -170,7 +170,10 @@ export const DisplayChartToolCall = ({
 	};
 
 	return (
-		<div className='flex flex-col items-center my-4 gap-2 aspect-3/2'>
+		// aspect-3/2
```
Bl3f: remove this

PR #381, opened 2026-03-01
feat(frontend): add SQL pretty print formatting

Balurc:
## Summary

Add automatic SQL formatting to the "View SQL" panel for better readability.

## Implementation Details

- Uses `sql-formatter` library for base formatting
- Uses `sql-formatter` library for base formatting
- Applies regex post-processing ONLY for SELECT statements. Regex splits SELECT columns to put each column on its own line
- Intentionally keeps FROM, WHERE, ORDER BY, etc on single lines (simpler, less risky?)
- Falls back to the original unformatted SQL if formatting fails

## Before

SQL displayed as a single line:

```sql
SELECT customer_id, first_name, last_name FROM customers WHERE number_of_orders > 2
```

## After

SQL with each clause on its own line:

```sql
SELECT
  customer_id,
  first_name,
  last_name
FROM customers
WHERE number_of_orders > 2
```

## Changes

- Install `sql-formatter` package
- Create `formatSQL` utility function in `apps/frontend/src/lib/sql-formatter.ts`
- Apply formatting to SQL display in `apps/frontend/src/components/side-panel/sql-editor.tsx`

Screenshot 2026-03-01 at 17 33 23

### DIFF
--- a/apps/frontend/src/lib/sql-formatter.ts
+++ b/apps/frontend/src/lib/sql-formatter.ts
@@ -1,15 +1,33 @@
-import { format } from 'sql-formatter';
+import { format, supportedDialects } from 'sql-formatter';
 
-export function formatSQL(sql: string): string {
+// Only map where nao db name differs from sql-formatter name
+const dialectMap: Record<string, string> = {
+	athena: 'trino',
+	databricks: 'spark',
+	mssql: 'tsql',
+	postgres: 'postgresql',
+};
+
+export function formatSQL(sql: string, dialect: string = 'sql'): string {
 	try {
+		// Normalize to lowercase for case-insensitive lookup
+		const normalizedDialect = dialect.toLowerCase();
+
+		// Check map first. If not in map, use the original dialect name.
+		const formatterDialect = dialectMap[normalizedDialect] || normalizedDialect;
+
+		// Validate it's supported by sql-formatter
+		const safeDialect = supportedDialects.includes(formatterDialect) ? formatterDialect : 'sql';
+
 		return format(sql, {
-			language: 'postgresql',
+			language: safeDialect as any,
 			keywordCase: 'upper',
 			indentStyle: 'standard',
 			tabWidth: 2,
 			linesBetweenQueries: 2,
 		});
 	} catch (error) {
+		// Fallback to original SQL if formatting fails
 		return sql;
 	}
 }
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
 	},
 	"dependencies": {
 		"@pydantic/monty": "^0.0.7",
-		"depd": "^2.0.0"
+		"depd": "^2.0.0",
+		"sql-formatter": "^15.7.2"
```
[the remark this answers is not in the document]
Balurc: Addressed in commit [2c20c7d](https://github.com/getnao/nao/commit/2c20c7d8ac03b37c76750aa08ec4beb2ffedd4fe)

Moved sql-formatter from root package.json to apps/frontend/package.json.

apps/frontend/src/lib/sql-formatter.ts
```diff
@@ -0,0 +1,15 @@
+import { format } from 'sql-formatter';
+
+export function formatSQL(sql: string): string {
+	try {
+		return format(sql, {
+			language: 'postgresql',
```
Bl3f: Can we pass the dialect the query is using so we make sure it works with most of it?

Balurc: Yes we can. I have implemented dialect-aware SQL formatting:

#### Changes Made
1. Backend - `apps/backend/fastapi/main.py`
- Added `dialect: str | None = None` to ExecuteSQLResponse model
- Returns `db_config.type `in the response (e.g., "duckdb", "postgres", "bigquery")
2. Shared Types - `apps/shared/src/tools/execute-sql.ts`
- Added `dialect: z.string().optional() `to OutputSchema
3. Frontend - `apps/frontend/src/lib/sql-formatter.ts`
- Maps nao database types to sql-formatter dialects:
```
    const dialectMap = {
    athena: 'trino',
    databricks: 'spark',
    mssql: 'tsql',
    postgres: 'postgresql',
    // ... others pass through as-is
  };
```
  - Validates against `supportedDialects` from `sql-formatter`
- Falls back to generic 'sql' formatter for unsupported types
- Handles case-insensitive lookup with `.toLowerCase()`
4. Frontend Component - `apps/frontend/src/components/side-panel/sql-editor.tsx`
- Uses `output.dialect` when calling `formatSQL()`

PR #382, opened 2026-03-01
feat(cli): support AI annotations in templates (prompt + ai_summary)

### LINKED ISSUE #226: AI annotations in the templating system
## User story
✨ As a user, I want to use AI to generate context on existing context assets

## Expected behaviour
- Templating supports an AI helper function, e.g. `prompt(...)`, callable inside Jinja templates.
- `prompt(string)` takes: a single string which concatenates prompt + given assets (table name, columns, descriptions, other template outputs) 
- If AI annotations are used without a configured LLM key, generation fails with a clear actionable error.
- Config allows choosing which model is used for AI annotations (with sensible provider defaults).
- New template available: `ai_summary.md.j2`.
- When `ai_summary` is enabled, system generates `ai_summary.md` per table using AI (from info of columns + description)

## CLI changes
- Add `ai_summary` option in template selection.
- Add AI-annotation model setting in config UX (or config schema/docs if file-driven only).
- Show clear error surface when AI annotation cannot run due to missing credentials/config.

## Edge cases
- Provider/model unavailable: fallback or explicit failure (as configured).

Flamki:
Closes #226

This PR adds AI annotations support to the CLI templating flow.

You can now call `prompt("...")` inside Jinja templates, and there is a new optional default template, `ai_summary.md.j2`, for table-level AI summaries.

What's included:

- Added `ai_summary` as a database accessor (opt-in)
- Added `llm.annotation_model` to config with provider defaults
- Added `prompt(...)` helper to the template engine
- Wired sync to pass LLM config so `prompt(...)` works during `nao sync`
- Added provider paths for OpenAI/OpenRouter, Anthropic, Mistral, Gemini, and Ollama
- Added explicit actionable errors when LLM config/credentials are missing
- Added default `databases/ai_summary.md.j2`
- Updated CLI README docs
- Added tests for:
  - annotation model defaults/validation
  - prompt helper behavior
  - template engine cache invalidation for LLM config changes
  - presence of ai_summary default template

Local validation (CLI scope):

- `ruff check` (including import checks): pass
- `ruff format --check`: pass
- `pytest` on touched CLI test sets: pass (37 passed)

Notes:

- `ai_summary` is intentionally opt-in, so existing users do not trigger AI calls unless they enable it.
- If enabled, table metadata/preview are sent to the configured LLM by design.

### DIFF
--- a/cli/nao_core/config/llm/__init__.py
+++ b/cli/nao_core/config/llm/__init__.py
@@ -146,7 +146,7 @@ class LLMConfig(BaseModel):
         if prompt_annotation_model:
             annotation_model = ask_text(
                 "Model to use for ai_summary generation (prompt helper):",
-                default=DEFAULT_ANNOTATION_MODELS.get(provider),
+                default=DEFAULT_ANNOTATION_MODELS[provider],
             )
 
         config = LLMConfig(


cli/nao_core/templates/engine.py
```diff
@@ -138,18 +301,37 @@ def is_user_override(self, template_name: str) -> bool:
 
 # Global template engine instance (lazily initialized)
 _engine: TemplateEngine | None = None
+_engine_signature: tuple[str | None, str | None, str | None, str | None, int | None] | None = None
+
+
+def _llm_signature(llm_config: LLMConfig | None) -> tuple[str | None, str | None, str | None, int | None]:
```
MatLBS: @Flamki, Claude told me something interesting, tell me if it's true.

```
id() returns the memory address of a Python object. Every time nao sync creates a new LLMConfig object (even with identical values), it gets a different address.

So the signature always changes → the cache always misses → a new TemplateEngine is constructed on every call. The cache does nothing.

Fix: remove id(llm_config) from the signature and add api_key instead (which is the field that actually affects behavior and was missing):


return (
    llm_config.provider.value,
    llm_config.annotation_model,
    llm_config.base_url,
    llm_config.api_key,   # ← replaces id()
)
```

Also, you should change the doc string -> `"""Return a tuple of LLM config values used as a cache key for the template engine."""`

Flamki:
> Hi @Flamki, thanks for the PR,
> 
> One thing that I find weird is that event if `llm-key` and `model` are setup for ai_annotations, the nao_config.yml does not include it :
> 
> ```
>   accessors:
>   - columns
>   - description
>   - preview
> ```
> 
> Should we should put it by default ?
> 
> Also, when running `nao init`, we set up `ai annotations` but in the config file we have to put `ai_summary`.
> 
> It would be nice to change everything to `ai_summary` for better consistency

Thanks @MatLBS — agreed, that inconsistency was real.

I addressed it in the latest update:

- During `nao init`, if an LLM is configured and there are database connections, we now ask:
  - `Enable ai_summary accessor for all configured databases?` (default: yes)
- If confirmed, `ai_summary` is automatically added to each DB’s `accessors` list (no duplicates).

So after init, users don’t need to manually edit `nao_config.yaml` to enable AI summaries.

I also aligned wording toward `ai_summary` for consistency (instead of mixed `ai_annotations` naming), including prompt/help text around the feature.

cli/nao_core/templates/engine.py
Flamki: Thanks @MatLBS — yes, that analysis is correct.

Using `id(llm_config)` in the cache signature made the key object-identity-based, so equivalent `LLMConfig` instances could still miss the cache.

I updated this in `templates/engine.py`:

- removed `id(llm_config)` from `_llm_signature(...)`
- switched to value-based fields:
  - `provider`
  - `annotation_model`
  - `base_url`
  - `api_key`
- updated the docstring to:
  - `Return a tuple of LLM config values used as a cache key for the template engine.`

This keeps cache reuse for equivalent config values while still invalidating when effective LLM settings change.

MatLBS:
Capture d’écran 2026-03-05 à 17 24 08

You should ask for the model only if user enable `ai_summary`...

Also do a quick rebase plz 🙏🏼

Issue #108, opened 2026-02-02
Support AWS Bedrocks


ClaireGz:
As a user, I want to use keys from AWS bedrocks

Bl3f:
When supporting Bedrock it should support 2 ways of connecting to it:
* Using the API_KEY
* Using the ACCESS_KEY_ID, and SECRET_ACCESS_KEY

With the API_KEY it's pretty straightforward to add with the current code, with ACCESS/SECRET keys it adds a new kind of way to auth to LLM APIs, it's feasible but will add a few ifs at few places.

Here the AI SDK doc for it (cf. https://ai-sdk.dev/providers/ai-sdk-providers/amazon-bedrock#using-iam-access-key-and-secret-key)

PR #383, opened 2026-03-01
Add AWS Bedrock as LLM Provider

rsivapr:
Adds support for AWS Bedrock as an LLM provider, with Claude Sonnet 4.6 and Claude Opus 4.6 available by default.

### Files touched
- `providers.ts` adds Bedrock provider config and model creator with support for both bearer token and IAM credential auth
- `llm.ts` adds `bedrock` to the `LlmProvider` type
- `usage-filters.tsx`adds 'AWS Bedrock' display label
- `.env.example`  required env vars for both auth methods
- `debug.py` / `__init__.py` CLI support for Bedrock

### Authentication
Bedrock supports two authentication methods as discussed in #108 
- Bearer token via `AWS_BEARER_TOKEN_BEDROCK`
- IAM credentials via `AWS_ACCESS_KEY_ID` + `AWS_SECRET_ACCESS_KEY`

Region defaults to `us-east-1` if `AWS_REGION` is not set.

### Note on missing `contextWindow` and `costPerM`
I tested this at work where we have custom pricing. Unlike other providers, Bedrock models do not have `contextWindow` or `costPerM` fields populated. AWS Bedrock pricing also varies by region, availability zone, and enterprise agreement, making it impractical to hardcode. Since I see examples of providers with no value in these fields, I decided to omit. But happy to consider any modifications here. 

### Note 
I've tested this only with bearer token. I don't have access to access key/secret mode of authentication, but boto should handle it well. 

image


Bl3f:
Hey @rsivapr I took the time to properly add Bedrock, there were some details to be added so the integration works in many cases everywhere in the UI! Thank you for starting this.

PR #392, opened 2026-03-02
[New Database Connector] Microsoft Fabric Data Warehouse / SQL Analytics Endpoint

d-axel-b:
## Summary

Adds Microsoft Fabric Data Warehouse / SQL Analytics Endpoint as a supported database connector.

### Authentication modes

Four auth modes are supported, each with a distinct connection strategy:

- **`sql_password`** — SQL username + password embedded in the ODBC string
- **`azure_interactive`** — Browser-based Azure AD login via `InteractiveBrowserCredential` from `azure-identity`. Uses Azure AD token injection (`SQL_COPT_SS_ACCESS_TOKEN`). I excluded ODBC driver's `ActiveDirectoryInteractive` mode, which does not reliably open a browser on macOS.
- **`azure_cli`** — Azure AD token from `az login` via `AzureCliCredential`, injected the same way
- **`azure_service_principal`** — Client ID + secret via `Authentication=ActiveDirectoryServicePrincipal` in the ODBC string, with optional `tenant_id`

### Key technical decisions

- **ibis 12 compatibility**: `ibis.mssql.connect(con=engine)` silently ignores the `con=` kwarg in ibis 12. The fix is `MSSQLBackend.from_connection(pyodbc_conn)`, which accepts a raw pyodbc connection directly.
- **pandas 2.0+ compatibility**: `pyodbc.Row` C extension objects are not treated as sequences by pandas 2.0+, causing shape errors on SQL results. Fixed in `DatabaseConfig.execute_sql()` by converting rows with `[list(row) for row in cursor.fetchall()]`. This fix applies to all connectors using the ODBC/pyodbc path.
- **Azure AD token encoding**: The ODBC driver requires the token encoded as UTF-16-LE with a 4-byte little-endian length prefix, passed via `attrs_before` before the session opens.

### Changes

- `cli/nao_core/config/databases/fabric.py` — new connector
- `cli/nao_core/config/databases/base.py` — pandas 2.0+ fix in `execute_sql()`, added `FABRIC` to `DatabaseType`
- `cli/pyproject.toml` — added `azure-identity>=1.19.0`
- `cli/tests/nao_core/config/test_fabric.py` — 39 unit tests (all passing)

## Test plan

- [x] `uv run pytest tests/nao_core/config/test_fabric.py` — 39 tests pass
- [x] `make lint` in `cli/` passes
- [x] `nao sync` with `azure_cli` auth mode connects and lists tables
- [x] `nao sync` with `azure_interactive` auth mode opens a browser and connects
- [x] SQL queries executed from the UI return correct results for Fabric

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -676,7 +676,7 @@
 			"version": "0.9.31",
 			"resolved": "https://registry.npmjs.org/@acemir/cssom/-/cssom-0.9.31.tgz",
 			"integrity": "sha512-ZnR3GSaH+/vJ0YlHau21FjfLYjMpYVIzTD8M8vIEQvIGxeOXyXdzCI140rrCY862p/C/BbzWsjc1dgnM9mkoTA==",
-			"devOptional": true,
+			"dev": true,
 			"license": "MIT"
 		},
 		"node_modules/@ai-sdk/amazon-bedrock": {
@@ -899,7 +899,7 @@
 			"version": "4.1.2",
 			"resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-4.1.2.tgz",
 			"integrity": "sha512-NfBUvBaYgKIuq6E/RBLY1m0IohzNHAYyaJGuTK79Z23uNwmz2jl1mPsC5ZxCCxylinKhT1Amn5oNTlx1wN8cQg==",
-			"devOptional": true,
+			"dev": true,
 			"license": "MIT",
 			"dependencies": {
 				"@csstools/css-calc": "^3.0.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
MatLBS, line 1: Remove all the changes from this file

MatLBS (changes requested):
Thanks for the PR @d-axel-b ,

Just remove your `package-lock.json` from the PR and we will be good to merge!

package-lock.json
Bl3f: The reason the package-lock is that big is because you are pbly using a node version to recent. You should nvm use in the project root so you use the version that we recommend

Issue #211, opened 2026-02-16
How to configure SQL Server or MYSQL databases?


raulcigil:
¿SQL Server and MYSQL are supported?
https://docs.getnao.io/nao-agent/context-builder/databases#supported-databases

```
Database Configuration

Select database type [bigquery/duckdb/databricks/snowflake/postgres] (bigquery): sqlserver
Please select one of the available options
Select database type [bigquery/duckdb/databricks/snowflake/postgres] (bigquery):
```

Please help for configure connections of this types.
Thanks.

PR #393, opened 2026-03-03
add mysql backend support

teoria:
#211 
add mysql backend support


docker-compose.mysql.yml
```diff
@@ -0,0 +1,20 @@
+services:
```
Bl3f: instead of creating this, the best would be to merge into a docker-compose.test.yml every database that can be embed in a Docker component, so we don't finish with an endless list of docker-compose files.

PR #399, opened 2026-03-03
Feat/227 db profiling context

socallmebertille:
## Summary

- Add a profiling accessor to the contextual database.
- Override the profiling method based on warehouse usage.
- Add integration tests for each.
- Add a profiling category to choose the refresh policy and its optional interval in days.
- Add unit tests to verify correct implementation.

## Related issue

Closing #227

### DIFF
--- a/cli/nao_core/config/databases/base.py
+++ b/cli/nao_core/config/databases/base.py
@@ -79,11 +79,12 @@ class DatabaseConfig(BaseModel, ABC):
             DatabaseAccessor.COLUMNS,
             DatabaseAccessor.DESCRIPTION,
             DatabaseAccessor.PREVIEW,
+            DatabaseAccessor.PROFILING,
         ],
         description=(
             "Which default templates to render per table "
             "(e.g., ['columns', 'description', 'ai_summary']). "
-            "Defaults to ['columns', 'description', 'preview']."
+            "Defaults to ['columns', 'description', 'preview', 'profiling']."
         ),
     )
     profiling: ProfilingConfig = Field(


cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -44,6 +45,41 @@ def _fmt_duration(seconds: float) -> str:
...
+    output_file: Path,
+    profiling_config,
+) -> bool:
+    """Decide whether profiling should be recomputed based on refresh policy."""
+
+    policy = profiling_config.refresh_policy
+    if policy == ProfilingRefreshPolicy.ALWAYS:
+        return True
+
+    if not output_file.exists():
```
Bl3f: This if condition should be first to appear no? You can even combine it with the ALWAYS condition

Also try to be consistent in the same function with the newline returns, either you add one either you don't, but don't mix it it makes it harder to read.

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -44,6 +45,41 @@ def _fmt_duration(seconds: float) -> str:
...
+        return True
+    if policy == ProfilingRefreshPolicy.ONCE:
+        return False
+
+    if policy == ProfilingRefreshPolicy.INTERVAL:
+        try:
+            content = output_file.read_text()
+            for line in content.splitlines():
```
Bl3f: Instead of split lines and doing a forloop, can't a regexp works for this?

cli/nao_core/config/databases/athena.py
```diff
@@ -7,6 +8,141 @@
 from nao_core.ui import ask_select, ask_text
 
 from .base import DatabaseConfig
+from .context import DatabaseContext
+
+
+class AthenaDatabaseContext(DatabaseContext):
+    def profiling(self) -> dict[str, Any] | None:
```
Bl3f: The way the SQL query is generated is not easy to maintain and/or read. Can't we find a better way where we have a base SQL query with string formatting injecting values in it?

I'm a bit afraid that it becomes a bit hard to maintain also because we have different SQL queries per providers, so finding bugs might be super hard because of this as well. 

A solution might be to have unit tests to check the SQL profiling query is correctly generated.

Let's talk about this 👍

PR #403, opened 2026-03-04
feat: send current datetime to LLM without invalidating cache

### LINKED ISSUE #372: Send the current datetime to the LLM to ground it
As a user i want the LLM to know about the current datetime. 

## Technical implementation
* We should send the current datetime to the LLM
* We should not send it in the system prompt because it invalidates eventual cache

Fizza-Mukhtar:
Closes #372

## What
Injects the current datetime as a `user`/`assistant` message pair immediately
after the system prompt in `_buildModelMessages()`.

## Why this approach
- System prompt has `CACHE_1H` — adding datetime there would invalidate
  cache on every request.
- Messages are injected in `UIMessage` format (matching the rest of the
  codebase) before `convertToModelMessages()` is called.
- A `user` + `assistant` pair maintains valid conversation structure.

## Message order after change
[0] system    ← CACHE_1H (untouched ✅)
[1] user      ← "Current datetime: 2026-03-04T..."
[2] assistant ← "Understood."
[3+] actual conversation messages

### DIFF
--- a/apps/backend/src/services/agent.ts
+++ b/apps/backend/src/services/agent.ts
@@ -285,13 +285,14 @@ class AgentManager {
 			parts: [{ type: 'text', text: `Current datetime: ${new Date().toISOString()}` }],
 		};
 
-		const datetimeAssistantMessage: Omit<UIMessage, 'id'> = {
-			role: 'assistant',
-			parts: [{ type: 'text', text: 'Understood.' }],
-		};
+		const lastUserIndex = [...uiMessagesWithCompaction].map(m => m.role).lastIndexOf('user');
+		const messagesWithDatetime = [...uiMessagesWithCompaction];
+		if (lastUserIndex !== -1) {
+			messagesWithDatetime.splice(lastUserIndex, 0, datetimeUserMessage);
+		}
 
 		const modelMessages = await convertToModelMessages<UIMessage>(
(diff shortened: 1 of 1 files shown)


apps/backend/src/services/agent.ts
```diff
@@ -280,9 +280,20 @@ class AgentManager {
...
-			tools: this._agentTools,
-		});
+		const datetimeUserMessage: Omit<UIMessage, 'id'> = {
+			role: 'user',
+			parts: [{ type: 'text', text: `Current datetime: ${new Date().toISOString()}` }],
+		};
+
+		const datetimeAssistantMessage: Omit<UIMessage, 'id'> = {
+			role: 'assistant',
+			parts: [{ type: 'text', text: 'Understood.' }],
```
Bl3f: Why do we need this?

apps/backend/src/services/agent.ts
```diff
@@ -280,9 +280,20 @@ class AgentManager {
...
+		};
+
+		const modelMessages = await convertToModelMessages<UIMessage>(
+			[systemMessage, datetimeUserMessage, datetimeAssistantMessage, ...uiMessagesWithCompaction],
```
Bl3f: I'm not sure it should be done this way to avoid invalidating the cache. Because here it's almost as the same as putting the date time in the system prompt. I think the date time should be added just before the last user message in the current agent loop.

PR #404, opened 2026-03-05
Fix : Feat/266 current context usage

socallmebertille:
## Summary

- Merge the tokenCounter estimation token logic into the context window usage calculation
- Correct the context window value update logic


apps/frontend/src/components/ui/chat-input-context-window-ring.tsx
```diff
@@ -66,6 +75,42 @@ export function ContextWindowRing({ className }: ContextWindowRingProps) {
 			},
 		),
 	);
+	const { refetch: refetchContextUsage, isFetching, isFetched } = contextUsage;
```
Bl3f: I feel we don't need to do this! Let's have a look together

PR #417, opened 2026-03-09
Added stacked bar validation to display chart

### LINKED ISSUE #414: Agent incorrectly uses the display chart tool for stacked bar charts
## Description
I have noticed that sometimes the agent intends to display a stacked bar chart but the query result is not in the correct format to support a stacked bar and we end up with just a bar chart that isn't very helpful to the user and even can be confusing. 

The agent needs to have a column for each stack in the bar chart however the agent often generates queries that group by the stack category rather than pivoting the data into separate series for each stack. As a result, it attempts to render a stacked bar chart with only a single series

## Example
In the example below I wanted the agent to breakdown sales by new vs existing customers. 

The agent generated a query where it grouped revenue by customer status (New, Returning) and then tried to create a stacked bar chart with month as the x-axis and revenue as the series. This resulted in a bar chart with 6 columns (each of the three months repeated for each customer status). 

In this case the query would need a column which summed revenue if customer was new and a column which summed revenue if the customer was returning and then the agent would successfully create the stacked bar chart.

### Query Output
Image

### Tool Input
```json
{
  "query_id": "query_37e410e6",
  "chart_type": "stacked_bar",
  "x_axis_key": "month",
  "x_axis_type": "date",
  "series": [
    {
      "data_key": "revenue",
      "color": "",
      "label": "Revenue"
    }
  ],
  "title": "Q4 2025 revenue by month: new vs returning"
}
```

### Chart displayed
Image

## Solution
I think this will be a rather common occurence as its more natural to write the query with the stack category as a seperate column and groupby this column than writing sumifs for each unique value in the stack category.

Add in validation to the stacked bar chart to check that more than one series is present. If there is one series then do not display the chart, mark it as a failure and send a message back to the agentic loop informing the agent that it probably wanted pivot and to try to do this now for the stacked bar chart it intended to display. So something like the below,

```ts
export default tool<displayChart.Input, displayChart.Output>({
	description: 'Display a chart visualization of the data from a previous `execute_sql` tool call.',
	inputSchema: displayChart.InputSchema,
	outputSchema: displayChart.OutputSchema,

	execute: async ({ chart_type: chartType, x_axis_key: xAxisKey, series }) => {
		// Validate xAxisKey is provided for bar/area charts
		if ((chartType === 'bar' || chartType === 'line') && !xAxisKey) {
			return { _version: '1', success: false, error: `xAxisKey is required for ${chartType} charts.` };
		}

		// Validate pie charts have exactly one series
		if (chartType === 'pie' && series.length !== 1) {
			return { _version: '1', success: false, error: 'Pie charts require exactly one series.' };
		}

		// Validate series is not empty
		if (series.length === 0) {
			return { _version: '1', success: false, error: 'At least one series is required.' };
		}

		// New Code
		if(chartType === 'stacked_bar' && series.length < 2){
			return { _version: '1', success: false, error: 'Stacked bar charts require at least two series. Perhaps you need to pivot the data' };
		}

		// TODO: check that the chart is displayable and that the data is valid

		return { _version: '1', success: true };
	},

	toModelOutput: ({ output }) => renderToModelOutput(DisplayChartOutput({ output }), output),
});
```

### Future Improvement (Optional)

An alternative approach would be to support a `split` or `breakdown` field in the chart schema that allows the renderer to dynamically generate series from a categorical column. However, enforcing a wide-format query keeps the chart renderer simpler and more deterministic.

cainemerrick98:
Fixes #414

Ensures that stacked bar has at least two series present. If not then a message is sent to the agent infomring it that it probably wants to pivot the data to display the chart. Example below

## Without fix
image

## With fix
image

### DIFF
--- a/apps/backend/src/agents/tools/display-chart.ts
+++ b/apps/backend/src/agents/tools/display-chart.ts
@@ -29,7 +29,7 @@ export default tool<displayChart.Input, displayChart.Output>({
 			return {
 				_version: '1',
 				success: false,
-				error: 'Stacked bar requires at least two series you may need to pivot the data to create a series for each stack',
+				error: 'Stacked bar chart requires at least two series. You may need to pivot the data to create a series for each stack.',
 			};
 		}
 


package.json
```diff
@@ -39,6 +39,8 @@
 	"dependencies": {
 		"@boxlite-ai/boxlite": "^0.3.0",
 		"@pydantic/monty": "^0.0.7",
-		"depd": "^2.0.0"
+		"depd": "^2.0.0",
+		"jsonrepair": "^3.13.2",
```
Bl3f: Why do you added these 2 packages?

cainemerrick98: Hey yeh my bad - I assumed these came from the main as I synced my fork with main and ran npm install and then tried to run npm run dev and was prompted that these packages were not installed - if its a version error I will fix and update the PR

Bl3f: No worries! Tell me if you need help

apps/backend/src/utils/slack.ts
```diff
@@ -75,7 +75,9 @@ function mdToMrkdwn(text: string): string {
 	const parts = text.split(/(```[\s\S]*?```|~~~[\s\S]*?~~~|`[^`\n]+`)/);
 	return parts
 		.map((part, i) => {
-			if (i % 2 === 1) return part;
+			if (i % 2 === 1) {
```
cainemerrick98, line 78: This was the linter

package.json
cainemerrick98: Thanks - think I managed to sort it tho!

PR #423, opened 2026-03-10
Feat/microsoft teams bot

### LINKED ISSUE #235: Queue next prompts
## User story
✨ As a user, I want to submit prompts while an agent is still generating, so I can stack work without waiting.

## Expected behaviour
- While generation is running, pressing Enter does not send immediately; it adds the prompt to a queue.
- The queue is visible in the UI with pending prompts.
- Multiple prompts can be queued (1, 2, 3, ... no fixed small limit in v1).
- When current generation ends, the next queued prompt is sent automatically, then the next, etc.
- User can click **Send immediately** on a queued prompt:
  - Cancel/interrupt current run.
  - Send that queued prompt right away.
  - It takes priority over the current execution.

## Technical implementation
- Queueing can be handled frontend-first (session/UI state + dispatch orchestration).
- Backend must support safe interruption + immediate handoff to a queued prompt.

## UI changes
- Show a queue panel/list of pending prompts.
- Actions per queued prompt: `Send immediately`, `Remove`
- Cursor-like UX is a good reference direction.

## Edge cases
- If interruption fails, keep queued items intact and show clear error.
- If queued send fails, allow retry without losing queue order.
- Navigation/reload behavior should be defined (v1 can be ephemeral if documented).

## Checks to validate
- Can queue prompts while a run is active.
- Multiple queued prompts are dispatched in order after run completion.
- `Send immediately` interrupts active run and dispatches selected queued prompt.

### LINKED ISSUE #416: Slack documentation lists incorrect OAuth scopes
## Problem

The [Slack integration documentation](https://docs.getnao.io/nao-agent/chat/slack#step-2-configure-bot-scope) lists incorrect OAuth scopes, causing `missing_scope` errors when using the Slack bot.

### 1. `users.profile:read` should be `users:read`

The docs say to add `users.profile:read`, but the code calls `users.info` API which requires `users:read`.

**Location in code:** `@chat-adapter/slack` package, `lookupUser()` method:
```typescript
const result = await this.client.users.info(this.withToken({ user: userId }));
```

**Error seen:**
```
[chat-sdk:slack] Could not fetch user info {
  error: warn: An API error occurred: missing_scope
  needed: "users:read"
  provided: "...,users.profile:read,..."
}
```

### 2. `files:write` is not documented but required

The nao backend uploads CSV query results to Slack using `files.uploadV2`, which requires `files:write` scope.

**Location in code:** `apps/backend/src/services/slack.ts`, `_uploadLastSqlResultAsCsv()` method:
```typescript
await this._slackClient?.files.uploadV2({
  channel_id: channelId,
  thread_ts: threadTs,
  filename,
  content: csv,
});
```

## Suggested Fix

Update the documentation to list the correct scopes:

```diff
- users.profile:read
+ users:read
+ files:write
```

## Environment

- nao version: v0.0.54+ (Docker image `e06ed70`)
- `@chat-adapter/slack`: v4.15.0

MatLBS:
Fix #235 
Fix #416 

teams

### DIFF
--- a/apps/backend/src/services/slack.ts
+++ b/apps/backend/src/services/slack.ts
@@ -43,38 +43,39 @@ class SlackService {
 
 	constructor() {}
 
-	public getWebhooks(slackConfig: SlackConfig) {
-		if (this._configChanged(slackConfig)) {
-			this._initialize(slackConfig);
+	public getWebhooks(config: SlackConfig) {
+		if (this._configChanged(config)) {
+			this._initialize(config);
 		}
 		return this._bot?.webhooks;
 	}
 
-	private _configChanged(slackConfig: SlackConfig): boolean {
+	private _configChanged(config: SlackConfig): boolean {
 		return (
-			this._currentBotToken !== slackConfig.botToken ||
-			this._currentSigningSecret !== slackConfig.signingSecret ||
-			this._projectId !== slackConfig.projectId ||
-			this._redirectUrl !== slackConfig.redirectUrl ||
-			this._modelSelection !== slackConfig.modelSelection
+			this._currentBotToken !== config.botToken ||
+			this._currentSigningSecret !== config.signingSecret ||
+			this._projectId !== config.projectId ||
+			this._redirectUrl !== config.redirectUrl ||
+			this._modelSelection?.provider !== config.modelSelection?.provider ||
+			this._modelSelection?.modelId !== config.modelSelection?.modelId
 		);
 	}
 
-	private _initialize(slackConfig: SlackConfig): void {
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/settings/slack-form.tsx
```diff
@@ -45,8 +45,9 @@ function buildSlackManifest(webhookUrl: string) {
 					'reactions:read',
 					'reactions:write',
 					'app_mentions:read',
-					'users.profile:read',
+					'users:read',
```
Bl3f, line 49: we need to add `users:read.email`

PR #427, opened 2026-03-10
Feat/add date timezone to system prompt

### LINKED ISSUE #372: Send the current datetime to the LLM to ground it

Bl3f:
closes #372 and #403 

Decided to only send the date instead of the datetime, it's way simpler and invalidate the cache only once a day.

### DIFF
--- a/apps/backend/src/components/ai/system-prompt.tsx
+++ b/apps/backend/src/components/ai/system-prompt.tsx
@@ -3,6 +3,7 @@ import type { Skill } from '../../services/skill';
 import { tokenCounter } from '../../services/token-counter';
 import type { UserMemory } from '../../types/memory';
 import { MEMORY_CATEGORIES, MemoryCategory } from '../../types/memory';
+import { formatCurrentDate } from '../../utils/date';
 import { groupBy } from '../../utils/utils';
 
 type Connection = {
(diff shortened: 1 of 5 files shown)


apps/backend/src/components/ai/system-prompt.tsx
```diff
@@ -161,6 +164,30 @@ function getMemoriesInTokenRange(memories: UserMemory[], limit: number): UserMem
 	return visible;
 }
 
+export function formatDate(timezone?: string): string {
```
MatLBS: Move this to a utils file ?

apps/frontend/src/hooks/use-agent.ts
```diff
@@ -108,6 +108,7 @@ export const useAgent = (): AgentHelpers => {
 							},
 							model: selectedModelRef.current ?? undefined,
 							mentions: mentions.length > 0 ? mentions : undefined,
+							timezone: Intl.DateTimeFormat().resolvedOptions().timeZone,
```
MatLBS, line 111: You should also add this in slack.ts, function `_createAgentStream` so that the agent on Slack as the same context 👌

Bl3f: Actually, the date is already added for Slack (in UTC via the system prompt) — i'm looking if it's possible to get user timezone

PR #445, opened 2026-03-12
fix: link project_id and domain_host across tracking events

### LINKED ISSUE #441: Add project_id in all tracking events
Currently we track: 
- domain_host on native posthog events ($pageview, $identify), but not the nao project_id
- project_id in message_sent events, but no domain_host.

-> makes it impossible to link a project_id to the corresponding host.

Expected behaviour
- Track project_id in all native posthog events
- OR add domain_host tracking in message_sent events
- OR add an event project_started at the first start of a project, where we track the project_id, and domain

paulpriyanshu:
Fixes #441

## Summary
Frontend PostHog events captured the host but lacked `project_id`.
Backend events captured `project_id` but lacked host context.
Slack and Teams events lacked complete telemetry, such as `domain_host`, `modelId`, and `isNewChat`.

This PR links everything by:
- registering `project_id` as a PostHog super property on the frontend
- adding `domain_host` from request headers to backend `message_sent` events
- correctly populating `domain_host`, `modelId`, and `isNewChat` in Slack and Teams events before capture

## Changes
- fetch and register `project_id` in `use-identify-posthog.ts`
- add `domain_host` to PostHog capture in `agent.ts`
- add `domain_host`, `modelId`, and missing captures to `slack.ts` and `teams.ts`

This allows correlating frontend (`$pageview`, `$identify`) and backend (`message_sent`) events in analytics across Web, Slack, and Teams.

### DIFF
--- a/apps/frontend/src/hooks/use-identify-posthog.ts
+++ b/apps/frontend/src/hooks/use-identify-posthog.ts
@@ -31,12 +31,9 @@ export const useIdentifyPostHog = () => {
 			posthog.identify(user.id, {
 				email_domain: user.email.split('@').at(1),
 				name: user.name,
-				project_id: project?.id,
+				project_id: project.id,
 			});
-
-			if (project?.id) {
-				posthog.register({ project_id: project.id });
-			}
+			posthog.register({ project_id: project.id });
 		} else if (wasConnectedRef.current) {
 			wasConnectedRef.current = false;
 			posthog.reset();


apps/frontend/src/hooks/use-identify-posthog.ts
MatLBS, line 26: add this : 

```
 if (!project?.id) {
    return;
} 
```

So that we wait until project is loaded

apps/backend/src/services/teams.ts
```diff
@@ -252,6 +256,15 @@ class TeamsService {
 		this._lastCompletionCard.set(ctx.thread.id, { card, chatUrl });
 
 		ctx.assistantMessage = state.lastMessage;
```
MatLBS: I think you can remove all occurrences to `assistantMessage` in both `slack and teams` as this variable is never used

apps/frontend/src/hooks/use-identify-posthog.ts
paulpriyanshu: Thanks @MatLBS  for you feedback. 
Sure, I will be adding this check.

apps/backend/src/services/teams.ts
paulpriyanshu: Yes, you right i will remove assistantMessage as it is not being used.

PR #446, opened 2026-03-13
WIP : System wide logs support

### LINKED ISSUE #432: Add a debug page in the admin to see backend logs
As a nao admin i want to be able to see backend logs and specifically errors in a admin panel page.

Rish-it:
## Summary

Adds a full logging capture layer that persists backend logs to the database and exposes them in a terminal-style admin settings page with real-time polling.

Closes #432

## What's included

**Backend — Capture layer**
- New `log` table (Postgres + SQLite) with level, message, source, context, and project association
- `logger` utility that writes to console AND persists to DB asynchronously via `scheduleTask`
- 4 log sources: `http`, `agent`, `tool`, `system`
- 4 severity levels: `error`, `warn`, `info`, `debug`
- Wired into all backend error paths: HTTP errors, tRPC errors, agent stream/title failures, memory injection/extraction, MCP connections/config/tools/watchers, skills loading, chart generation (Slack + Teams), and email sending
- HTTP request logging via Fastify `onResponse` hook (skips log-polling to avoid noise)
- 7-day retention cleanup routine (runs on startup + every 24h)
- Admin-only tRPC endpoint with level/source filtering

**Frontend — Terminal UI**
- New `/settings/logs` page (admin-only) under Settings nav
- Terminal-style view with monospace font, color-coded levels and sources
- Auto-scroll to bottom (like tailing server logs)
- Polls every 5 seconds for near real-time updates
- Level and source filter dropdowns
- Copy logs to clipboard + manual refresh

## Design decisions

- **Polling over WebSocket** — avoids infra changes for self-hosted deployments; 5s interval is good enough for v1
- **DB storage over files** — simpler to query/filter, works across all deployment targets
- **`scheduleTask` for writes** — log persistence is fire-and-forget, never blocks request handling
- **No pagination** — terminal-style scroll with 200-entry limit; keeps the UI simple for v1

## Test plan

- [ ] Verify logs page is only visible to admin users
- [ ] Trigger errors (bad MCP config, agent failure) and confirm they appear in the logs UI
- [ ] Verify level and source filters work correctly
- [ ] Verify auto-scroll behavior (scrolls to bottom on new logs, pauses when user scrolls up)
- [ ] Verify copy button copies formatted log text
- [ ] Confirm log cleanup removes entries older than 7 days
- [ ] Test on both Postgres and SQLite

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -10,7 +10,7 @@
 		"./saved-prompts": "./src/types/saved-prompt.ts",
 		"./tools": "./src/agents/tools/index.ts",
 		"./providers": "./src/agents/providers.ts",
-		"./log": "./src/types/log.ts"
+		"./log": "./src/types/log.ts",
 		"./provider-meta": "./src/agents/provider-meta.ts"
 	},
 	"scripts": {


apps/backend/src/services/skill.ts
```diff
@@ -43,21 +44,24 @@ class SkillService {
...
 				this._skills = [];
 				return;
 			}
 
 			const files = readdirSync(this._skillsFolderPath).filter((f) => f.endsWith('.md'));
 			this._readSkills(files);
 		} catch (error) {
-			console.error('[skills] Failed to load skills:', error);
+			logger.error('Failed to load skills', {
+				source: 'system',
```
Bl3f: you can categorise skills issues as agent

apps/backend/src/utils/logger.ts
```diff
@@ -0,0 +1,56 @@
...
+export function serializeError(error: unknown): Record<string, unknown> {
+	if (error instanceof Error) {
+		return { name: error.name, message: error.message, stack: error.stack };
+	}
+	return { value: String(error) };
+}
+
+const SENSITIVE_KEYS = new Set(['password', 'token', 'secret', 'authorization', 'cookie', 'apikey', 'api_key']);
+
+function redactContext(ctx: Record<string, unknown>): Record<string, unknown> {
```
Bl3f, line 21: maybe there is a pattern recognition in text also to redact keys that might be slipping through this—tho i think it's a good first start

apps/backend/src/app.ts
```diff
@@ -77,7 +104,10 @@ app.register(fastifyTRPCPlugin, {
 		router: trpcRouter,
 		createContext,
 		onError({ path, error }) {
-			console.error(`Error in tRPC handler on path '${path}':\n`, error);
+			logger.error(`tRPC error on ${path}: ${error.message}`, {
+				source: 'system',
```
Bl3f:
```suggestion
				source: 'http',
```

trpc is actually the api so http would be a better category

Rish-it:
@Bl3f Rebased on the latest `main` and updated the migration numbering from `0025` → `0026`, including corresponding snapshot and journal updates for both Postgres and SQLite. happy to increment it to `0027` if another PR is ready.

Issue #233, opened 2026-02-17
Share a chat experience


Bl3f:
## User story
✨ As a user, I want to share a chat with teammates so they can read my analysis without editing the conversation.

## Expected behaviour
- Chat owner can click `Share` from a conversation.
- Owner can grant read-only access:
  - to specific users (by email, existing users in same project),
  - or to all users in the same project.
- Shared users can open the chat via URL and also see it in their sidebar.
- Shared chat is strictly read-only
- Chat ownership does not change: one chat still has one owner.

## Technical implementation
- WIP

## UI changes
- Add `Share` action in the chat submenu in the left side bar
- Share modal:
  - add viewer by email,
  - toggle/share with “Everyone in this project,”
  - list/remove existing viewers.
- Sidebar:
  - show shared chats with distinct icon/style,
  - support filtering (`My chats` vs `Shared with me`).
- Chat screen in shared mode:
  - visible `Read-only` indicator,
  - composer hidden/disabled.

## Checks to validate
- Owner can share/unshare with specific project users.
- Owner can enable/disable project-wide read-only sharing.
- Shared user sees chat in sidebar and can open via URL.
- Shared user cannot send/edit in that chat.
- Non-authorized users cannot access shared chat.

PR #486, opened 2026-03-19
Feat/233 share chat experience

socallmebertille:
## Summary

- Add share option on each chats
image

image

image

- Sharing a chat to all project members in one click or to selected members
image

image

- Each sharing send an email to viewers with a link
image

- Add stories view read only on shared chats AND chats replay for admin
image

image

- "Shared with me" chats on the chat list sidebar AND on the filter button next to the chat list header folder
image

image

## Linked issue

#233

### DIFF
--- a/apps/frontend/src/components/share-dialog.chat.tsx
+++ b/apps/frontend/src/components/share-dialog.chat.tsx
@@ -1,17 +1,18 @@
 import { useState, useCallback, useRef, useEffect, useMemo } from 'react';
 import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
-import { Check, Globe, Loader2, Users, Link as LinkIcon, Unlink } from 'lucide-react';
+import { Check, Link as LinkIcon, Loader2 } from 'lucide-react';
 import type { Visibility } from '@/components/share-dialog';
-import { hasAccessChanges, MemberPicker, VisibilityOption } from '@/components/share-dialog';
-import { Button } from '@/components/ui/button';
 import {
-	Dialog,
-	DialogContent,
-	DialogDescription,
-	DialogFooter,
-	DialogHeader,
-	DialogTitle,
-} from '@/components/ui/dialog';
+	hasAccessChanges,
+	ManageShareFooter,
+	MemberPicker,
+	ShareErrorDialog,
+	ShareLoadingDialog,
+	VisibilityPicker,
+	VisibilitySummary,
+} from '@/components/share-dialog';
+import { Button } from '@/components/ui/button';
+import { Dialog, DialogContent, DialogDescription, DialogHeader, DialogTitle } from '@/components/ui/dialog';
 import { useSession } from '@/lib/auth-client';
 import { trpc } from '@/main';
 import { useMemberPicker, useCopyWithFeedback } from '@/hooks/use-share-dialog';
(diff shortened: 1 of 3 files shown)


apps/backend/src/components/email/shared-chat.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { EmailLayout } from './email-layout';
+
+interface SharedChatProps {
+	userName: string;
+	sharerName: string;
+	chatTitle: string;
+	chatUrl: string;
+}
+
+export function SharedChat({ userName, sharerName, chatTitle, chatUrl }: SharedChatProps) {
```
Bl3f: I think it could be factorise with shared-story as the template is almost the same (except the name of the object)

apps/backend/src/db/pg-schema.ts
```diff
@@ -331,7 +331,41 @@ export const projectLlmConfig = pgTable(
...
-export const STORY_VISIBILITY = ['project', 'specific'] as const;
+export const SHARE_VISIBILITY = ['project', 'specific'] as const;
+
+export const sharedChat = pgTable(
+	'shared_chat',
+	{
+		id: text('id')
+			.$defaultFn(() => crypto.randomUUID())
+			.primaryKey(),
+		projectId: text('project_id')
```
Bl3f: Do we really need to copy the projectId and the userId in this table as it's already in chat table?

apps/backend/src/queries/shared-chat.queries.ts
```diff
@@ -0,0 +1,138 @@
...
+			return true;
+		}
+		if (chat.userId === userId) {
+			return true;
+		}
+		return accessibleIds.has(chat.id);
+	});
+}
+
+export async function findByChat(chatId: string, userId: string): Promise<{ id: string; visibility: string } | null> {
```
Bl3f: Not sure this function is well named

apps/backend/src/services/email.ts
```diff
@@ -52,6 +53,8 @@ class EmailService {
 			email = this._createResetPasswordEmail(params);
 		} else if (params.type === 'sharedStory') {
 			email = this._createSharedStoryEmail(params);
+		} else if (params.type === 'sharedChat') {
```
Bl3f: It's more a design issue from the EmailService, but sendEmail should not contain this if, the method sendEmail should receive an email to send (already created and send it) also the method to create the email should be outside of a generic email service which role is to send email not to know what are all the different emails possible.

apps/backend/src/trpc/router.ts
```diff
@@ -21,6 +22,7 @@ import { userRoutes } from './user.routes';
 
 export const trpcRouter = router({
 	chart: chartRoutes,
+	sharedChat: sharedChatRoutes,
```
Bl3f: put sharedChat after chat

apps/backend/src/trpc/shared-chat.routes.ts
```diff
@@ -0,0 +1,162 @@
...
+				shareId: created.id,
+				chatTitle: chatInfo.title,
+				visibility: input.visibility,
+				allowedUserIds: input.allowedUserIds,
+			}).catch((err) => console.error('Failed to notify shared chat recipients', err));
+
+			return created;
+		}),
+
+	getChat: protectedProcedure
```
Bl3f: It's not easy to understand why we have a getChat, a get and a findByChat endpoints and to what each correspond. Especially getChat which actually could be a route from the chat.routes.ts

(didn't read the frontend part yet) but maybe in the frontend you should just use chat ids instead of sharedId and add the permissions to get a chat info in the chat.routes instead?

apps/backend/src/types/project.ts
```diff
@@ -1,4 +1,6 @@
-import type { UserRole } from '@nao/shared';
+export type UserRole = 'admin' | 'user' | 'viewer';
```
Bl3f: can't we still import this from shared?

apps/backend/src/utils/email.ts
```diff
@@ -41,3 +41,40 @@ export async function notifySharedStoryRecipients({
...
+}: {
+	projectId: string;
+	sharerId: string;
+	sharerName: string;
+	shareId: string;
+	chatTitle: string;
+	visibility: 'project' | 'specific';
+	allowedUserIds?: string[];
+}): Promise<void> {
+	const chatUrl = `${env.BETTER_AUTH_URL || 'http://localhost:3000'}/shared-chat/${shareId}`;
```
Bl3f: don't do the ||, there is already a default on BETTER_AUTH_URL in env.ts

apps/frontend/src/components/side-panel/share-story-dialog.tsx
```diff
@@ -1,9 +1,9 @@
-import { useState, useMemo, useCallback, useRef, useEffect } from 'react';
```
Bl3f, line 1: move this close to share-dialog.chat.tsx, it's ok if not in the side-panel folder

apps/frontend/src/components/share-dialog.chat.tsx
```diff
@@ -0,0 +1,360 @@
+import { useState, useCallback, useRef, useEffect, useMemo } from 'react';
```
Bl3f, line 1: Not easy to read with all the changes, but make sure to factorise what can be factorised with the share-dialog.story.tsx it seems there is text duplicated still?

apps/frontend/src/hooks/use-side-panel.ts
```diff
@@ -13,9 +13,13 @@ import {
 export const useSidePanel = ({
 	containerRef,
 	sidePanelRef,
+	defaultWidthRatio,
+	collapsesSidebar = true,
```
Bl3f:
```suggestion
	shouldCollapseSidebar = true,
```

Issue #316, opened 2026-02-20
In admin, see token usage by action (chat, test, memory, voice)


ClaireGz:
## User Story
✨ As an admin, I want to monitor precisely my LLM costs

## Expected behaviour
In the admin / Usage & costs panel, show the costs of LLM split between these actions: 
- nao test
- nao chats (by file, tool, user, chat)
- memory
- voice

## Technical implementation
- Track for every LLM usage the input / output token, models used and associated costs
- Surface in admin panel

## UI changes
- Add filter / categories on usage chart

PR #489, opened 2026-03-20
usage by action + ux sanding

Rish-it:
## Summary

Implements #316 — surfaces LLM cost breakdown by action type (chat, test, memory, voice) in the admin Usage & Costs panel.

### Backend

- **Schema**: Added `estimated_cost` column to `llm_inference` table (both SQLite and PostgreSQL) for voice transcription costs that use per-minute pricing instead of per-token
- **Inference tracking**: Every LLM call now saves a record to `llm_inference` via `scheduleSaveLlmInferenceRecord`:
  - `chat` — streaming responses in `AgentManager.stream()`
  - `title_generation` — chat title generation in `AgentManager._generateTitle()`
  - `test` — test runs and verifications in `TestAgentService`
  - `voice` — audio transcription in `transcribeAudio()` with duration-based cost
- **Usage query**: New `getActionCostsByDate()` query aggregates costs from `llm_inference` grouped by action type, merged with existing message usage data. Chat cost includes `chat`, `compaction`, and `title_generation` types. Voice cost uses the pre-computed `estimated_cost` column instead of token math.
- **Date alignment fix**: Usage records now merge dates from both message and action queries (previously action-only dates were silently dropped)

### Frontend

- Added **"Actions"** chart view to the usage page dropdown — stacked bar chart showing Chat, Test, Memory, and Voice costs over time
- Granularity and provider filters apply to the actions view

### UX sanding

- Added back button on story preview page to return to `/stories`

## Test plan

- [x] Run `npm run db:push` to apply schema changes
- [x] Send chat messages and verify `llm_inference` records are created with `type = 'chat'`
- [x] Run `nao test` and verify `type = 'test'` records appear
- [x] Use voice transcription and verify `type = 'voice'` records with `estimated_cost`
- [x] Open admin Usage & Costs → select "Actions" view → verify stacked bar chart renders with 4 series
- [x] Toggle granularity (hour/day/month) and provider filter — chart updates correctly
- [x] Verify existing chart views (Messages, Tokens, Cost) are unaffected
- [x] Verify story preview back button navigates to `/stories`
- [x] `npm run lint` passes


Rish-it:
hey @Bl3f

while review, I noticed a couple of non-blocking improvements worth considering:

* **Duplicate cost table construction**
  `buildCostValuesTable` is currently invoked twice per request (`getMessagesUsage` and `getActionCostsByDate`), rebuilding the same `VALUES` table each time. We could pass the computed table as a parameter to avoid redundant work.

* **Hardcoded `userId` in test flow**
  In `TestAgentService.runTest`, the temporary chat uses `userId: 'test'`, while inference records use the actual `userId`. If compaction runs during a test, it logs against the wrong user. This seems to be a pre-existing issue, not introduced by this PR.

Bl3f:
We will change the admin panel a bit, this PR will be redone later on.

PR #504, opened 2026-03-22
Breakdown key added to series config schema to allow for multi series charts

cainemerrick98:
## Summary
Implements idea from #414 
* Added optional breakdown key to series config schema 
* Updated validation in backend
* Added pivot logic and series expansion to the display chart component

### The Query
image

### Stacked Bar
image

### Line
image

### Bar
image

### Tool Call Examples 
```json
"{""title"": ""Sales by month by payment method"", ""series"": [{""color"": """", ""label"": ""Sales"", ""data_key"": ""sales_amount"", ""breakdown_key"": ""payment_method""}], ""query_id"": ""query_91cdb958"", ""chart_type"": ""stacked_bar"", ""x_axis_key"": ""order_month"", ""x_axis_type"": ""date""}"
```
```json
"{""title"": ""Sales by month by payment method"", ""series"": [{""color"": """", ""label"": ""Sales"", ""data_key"": ""sales_amount"", ""breakdown_key"": ""payment_method""}], ""query_id"": ""query_91cdb958"", ""chart_type"": ""bar"", ""x_axis_key"": ""order_month"", ""x_axis_type"": ""date""}"
```

```json
"{""title"": ""Sales by month by payment method"", ""series"": [{""color"": """", ""label"": ""Sales"", ""data_key"": ""sales_amount"", ""breakdown_key"": ""payment_method""}], ""query_id"": ""query_91cdb958"", ""chart_type"": ""line"", ""x_axis_key"": ""order_month"", ""x_axis_type"": ""date""}"
```

### DIFF
--- a/apps/backend/src/agents/tools/display-chart.ts
+++ b/apps/backend/src/agents/tools/display-chart.ts
@@ -24,13 +24,23 @@ export default tool<displayChart.Input, displayChart.Output>({
 			return { _version: '1', success: false, error: 'At least one series is required.' };
 		}
 
-		// Stacked bar requires at least two series
-		if (chartType === 'stacked_bar' && series.length < 2) {
+		if (series.some((s) => s.breakdown_key) && series.some((s) => s.breakdown_key === undefined)) {
 			return {
 				_version: '1',
 				success: false,
-				error: 'Stacked bar chart requires at least two series. You may need to pivot the data to create a series for each stack.',
+				error: 'Cannot combine breakdown series and none breakdown series',
 			};
+		}
+
+		// Stacked bar requires at least two series or a breakdown
+		if (chartType === 'stacked_bar' && series.length < 2) {
+			if (!series[0]?.breakdown_key) {
+				return {
+					_version: '1',
+					success: false,
+					error: 'Stacked bar chart requires at least two series or a breakdown key. Pivot the data to create a series for each stack or provide a breakdown key.',
+				};
+			}
 		}
 
 		// TODO: check that the chart is displayable and that the data is valid
--- a/apps/backend/src/agents/tools/display-chart.ts
+++ b/apps/backend/src/agents/tools/display-chart.ts
@@ -19,16 +19,31 @@ export default tool<displayChart.Input, displayChart.Output>({
 			return { _version: '1', success: false, error: 'Pie charts require exactly one series.' };
 		}
 
+		// Validate breakdown key not set when for pie charts
+		if (chartType === 'pie' && series[0].breakdown_key) {
+			return { _version: '1', success: false, error: 'Pie charts do not accept a breakdown key.' };
+		}
+
 		// Validate series is not empty
 		if (series.length === 0) {
 			return { _version: '1', success: false, error: 'At least one series is required.' };
 		}
 
+		// Validates that breakdown series are not combined with non-breakdown series
 		if (series.some((s) => s.breakdown_key) && series.some((s) => s.breakdown_key === undefined)) {
 			return {
 				_version: '1',
 				success: false,
-				error: 'Cannot combine breakdown series and none breakdown series',
+				error: 'Cannot combine breakdown and non-breakdown series.',
+			};
+		}
+
+		// Validates that only one breakdown series is passed
+		if (series.filter((s) => s.breakdown_key).length > 1) {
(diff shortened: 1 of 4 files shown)


apps/frontend/src/components/tool-calls/display-chart.tsx
```diff
@@ -353,6 +368,59 @@ export const ChartDisplay = memo(function ChartDisplay({
...
+			}
+
+			const breakdown = row[s.breakdown_key] as string;
+			const value = row[s.data_key];
+
+			if (breakdown == null) {
+				continue;
+			}
+
+			target[breakdown] = value;
```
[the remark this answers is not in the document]
cainemerrick98: This is actually a good point - what if we breakdown two columns and they share a value

We will have to track the breakdown values from each column and see if we get dupes.

How likely is this though

apps/shared/src/tools/display-chart.ts
```diff
@@ -6,6 +6,12 @@ export const XAxisTypeEnum = z.enum(['date', 'number', 'category']);
 
 export const SeriesConfigSchema = z.object({
 	data_key: z.string().describe('Column name from SQL result to plot.'),
+	breakdown_key: z
```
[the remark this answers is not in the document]
cainemerrick98: The field is marked as optional so should be ok - perhaps we need to add validation against breakdown being included for pie charts ???

apps/backend/src/agents/tools/display-chart.ts
```diff
@@ -24,15 +24,25 @@ export default tool<displayChart.Input, displayChart.Output>({
...
 		}
 
-		// Stacked bar requires at least two series
-		if (chartType === 'stacked_bar' && series.length < 2) {
+		if (series.some((s) => s.breakdown_key) && series.some((s) => s.breakdown_key === undefined)) {
 			return {
 				_version: '1',
 				success: false,
-				error: 'Stacked bar chart requires at least two series. You may need to pivot the data to create a series for each stack.',
+				error: 'Cannot combine breakdown series and none breakdown series',
```
MatLBS: `Cannot combine non-breakdown series ...`

apps/frontend/src/components/tool-calls/display-chart.tsx
```diff
@@ -353,6 +368,59 @@ export const ChartDisplay = memo(function ChartDisplay({
...
+
+		const target = grouped.get(x)!;
+
+		for (const s of series) {
+			if (!s.breakdown_key) {
+				continue;
+			}
+
+			const breakdown = row[s.breakdown_key] as string;
+			const value = row[s.data_key];
```
MatLBS, line 401:
```
const breakdown = (row[s.breakdown_key] ?? 'Unknown') as string;
const value = row[s.data_key];

target[breakdown] = value;
```

We should group Null/undefined breakdown values into an "Unknown" series instead of being silently dropped.

apps/frontend/src/components/tool-calls/display-chart.tsx
```diff
@@ -353,6 +368,59 @@ export const ChartDisplay = memo(function ChartDisplay({
...
+					data_key: breakdown,
+					label: labelize(breakdown),
+				});
+			}
+		}
+	}
+
+	return {
+		pivotedData: Array.from(grouped.values()),
+		expandedSeries,
```
MatLBS:
```
return {
		pivotedData: Array.from(grouped.values()),
		expandedSeries: expandedSeries.sort((a, b) => a.data_key.localeCompare(b.data_key)),
	};
```

Would be nice to find a way to have consistant colors between the charts.
 
Try to return `expandedSeries` alphabetically so that colors stay consistent on different charts

Example : 

Colors are inconsistent on those charts

Image
Image

MatLBS (changes requested):
Hi @cainemerrick98 , thanks for the PR! 

There is small changes to do, also we could add something like this to the system-prompt so that the agent do not create PIVOT sql query anymore : 

```
For multi-series charts where data has a categorical column (e.g. payment_method, region),                                                                                                         
  use breakdown_key on the series instead of pivoting in SQL.
```

What do you think? 

Apply cubic recommendations:

- add validation against breakdown being included for pie charts 
...

PR #509, opened 2026-03-24
added reset password feature if SMTP enabled

MatLBS:
feature enabled only if SMTP enabled

Capture d’écran 2026-03-24 à 14 58 08

### DIFF
--- a/apps/backend/src/services/email.ts
+++ b/apps/backend/src/services/email.ts
@@ -43,7 +43,7 @@ class EmailService {
 	}
 
 	public async sendEmail(to: string, email: CreatedEmail): Promise<void> {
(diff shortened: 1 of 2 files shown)


apps/backend/src/services/email.ts
```diff
@@ -38,6 +38,10 @@ class EmailService {
 		}
 	}
 
+	public isEnabled(): boolean {
+		return this.enabled;
+	}
+
 	public async sendEmail(to: string, email: CreatedEmail): Promise<void> {
 		if (!this.enabled || !this.transporter) {
```
Bl3f: use the newly created isEnabled?
```suggestion
		if (!this.isEnabled() || !this.transporter) {
```

PR #523, opened 2026-03-26
Add pagination to SQL results table (#520)

### LINKED ISSUE #520: When execute SQL returns too many lines it makes the UI laggy when you open the results

MatLBS:
## Summary
- Adds client-side pagination to `TableDisplay` component (default 50 rows/page) to prevent rendering thousands of DOM nodes at once
- Refactors the SQL editor side panel to reuse `TableDisplay` instead of its own table rendering (with 100 rows/page)
- Pagination only appears when row count exceeds the page size, keeping the UI clean for small result sets

Closes #520

## Test plan
- [x] Execute a SQL query that returns 500+ rows — verify the table is paginated and the UI stays responsive
- [x] Verify pagination controls work (first/prev/next/last page, rows per page selector)
- [x] Verify small result sets (<50 rows) display without pagination controls
- [x] Expand a large result set to the side panel — verify pagination works there too
- [x] Verify the row count is still displayed accurately


MatLBS:
Better now, consistant with chats replay pagination

Capture d’écran 2026-03-31 à 10 09 00

PR #526, opened 2026-03-26
Add live stories

Bl3f:


apps/backend/src/services/live-story.ts
MatLBS, line 1: Shouldn't this file be a singleton to match other services ?

apps/backend/src/services/live-story.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export async function getStoryQueryData(
+	chatId: string,
+	slug: string,
+	code: string,
+	isLive: boolean,
+	cacheSchedule: string | null,
+): Promise<StoryQueryDataResult> {
+	if (!isLive) {
+		const { collectQueryData } = await import('../queries/shared-story.queries');
```
MatLBS: Move this import to the header of the file

apps/backend/src/services/live-story.ts
```diff
@@ -0,0 +1,292 @@
...
+		const { queryData } = await refreshStoryData(chatId, slug);
+		return {
+			queryData: Object.keys(queryData).length > 0 ? queryData : null,
+			cachedAt: new Date(),
+		};
+	} catch {
+		if (cache) {
+			return { queryData: cache.queryData, cachedAt: cache.cachedAt };
+		}
+		const { collectQueryData } = await import('../queries/shared-story.queries');
```
MatLBS: same

apps/frontend/src/components/side-panel/live-story-settings-dialog.tsx
```diff
@@ -0,0 +1,283 @@
...
+	isLiveTextDynamic,
+	cacheSchedule,
+	cacheScheduleDescription,
+	isUpdating,
+	onSaveSettings,
+}: LiveStorySettingsDialogProps) {
+	const [localIsLive, setLocalIsLive] = useState(isLive);
+	const [localPreset, setLocalPreset] = useState(() => resolvePresetValue(cacheSchedule));
+	const [localCustomCron, setLocalCustomCron] = useState(localPreset === 'custom' ? (cacheSchedule ?? '') : '');
+	const [localTextBlocksDynamic, setLocalTextBlocksDynamic] = useState(isLiveTextDynamic);
```
MatLBS, line 67: Do you think it's beneficial to keep the option to turn off 'Live text'?

Shouldn't it be activated in every case ? 

What could be the purpose for a user to refresh its graphs but do not update the corresponding text blocks ?

apps/backend/src/services/live-story.ts
```diff
@@ -0,0 +1,292 @@
...
+			.filter((value): value is number => value !== null);
+
+		if (!values.length) {
+			continue;
+		}
+
+		const sum = values.reduce((total, value) => total + value, 0);
+		summaries[column] = {
+			min: Math.min(...values),
+			max: Math.max(...values),
```
MatLBS:
```
min: values.reduce((a, b) => Math.min(a, b)),                             
max: values.reduce((a, b) => Math.max(a, b)),                                                                                                                                                      
```
reduce iterates without touching the call stack.

apps/backend/src/services/live-story.ts
```diff
@@ -0,0 +1,292 @@
...
+import { env } from '../env';
+import { renderToMarkdown } from '../lib/markdown';
+import * as chatQueries from '../queries/chat.queries';
+import * as projectQueries from '../queries/project.queries';
+import * as llmConfigQueries from '../queries/project-llm-config.queries';
+import * as storyQueries from '../queries/story.queries';
+import { getDefaultModelId, resolveProviderModel } from '../utils/llm';
+import { MAX_OUTPUT_TOKENS } from './agent';
+
+export const NO_CACHE_SCHEDULE = 'no-cache';
```
MatLBS: variable `NO_CACHE_SCHEDULE` is created 3 times in the codebase. 

Created it in `/shared` and reuse it

apps/frontend/src/components/story-embeds.tsx
```diff
@@ -0,0 +1,118 @@
...
+		<EmbedPlaceholder>
+			<span className='flex items-center justify-center'>
+				<Loader2 className='size-4 animate-spin mr-2' />
+				Loading live data...
+			</span>
+		</EmbedPlaceholder>
+	);
+}
+
+function useLiveQueryData(queryId: string, liveQuery?: LiveQueryConfig) {
```
MatLBS, line 36: Could be optimized like this : 

```
if (!liveQuery) return { data: undefined, isLoading: false };
return useQuery({ ...liveQuery.queryOptions(...), staleTime: 0 });
````

apps/backend/src/services/cron-nlp.ts
MatLBS, line 1: Shouldn't this file be a singleton to match other services ?

apps/backend/src/services/live-story.ts
Bl3f: Not sure it should be a singleton, I'm fine with a collection of function, but might be more a util in this case.

apps/backend/src/services/cron-nlp.ts
Bl3f: Not for this one, it's pretty straightforward with 2 functions and no state to manage.

apps/frontend/src/components/side-panel/live-story-settings-dialog.tsx
Bl3f: I'm up to keep the option, maybe activated by default tho, I think there are cases where I don't want to call an AI everytime it refreshes the charts.

apps/frontend/src/components/story-embeds.tsx
Bl3f: Nope, useQuery should be called in every branch

PR #545, opened 2026-04-01
Improved project settings

tarun-p10:
Screenshot 2026-04-01 235456
I have moved sidebar to the top so that it appears as a tab options


Bl3f:
I think we dont want to do it this way, will work on this later.

Issue #530, opened 2026-03-27
feature: export stories/report as HTML or png


dennisUhrskov:
As a user I want to be able to export a story/report as HTML or png, HTML would be better as to keep the interactive features. Marimo reports has this feature, it works well for storing reports locally and to share with the team.

PR #567, opened 2026-04-07
Feat/530 export story pdf or html

socallmebertille:
## Summary

- [x] The download button appears on both your own stories and shared stories

- You can export any version of a story, not just the latest one
- Access permissions are respected: shared story downloads follow the same visibility rules as viewing

image
image

image
image

- [x]  Implementation of PDF story export

- Users can download any story as a PDF document (A4 format, print-ready)
- Charts are rendered as vector graphics (SVG) for sharp output
- Tables are included with up to 10 rows and a row count summary

- [x]  Implementation of HTML story export

- Users can download a self-contained HTML file that works offline in any browser
- Charts include interactive tooltips on hover (just like in the app)
- Styled for readability with a clean layout, also optimised for printing

## Related issue

#530


apps/backend/src/utils/story-html.ts
```diff
@@ -0,0 +1,280 @@
...
+const MAX_TABLE_ROWS = 10;
+
+export function generateStoryHtml(story: StoryInput, queryData: QueryDataMap | null): string {
+	const segments = splitCodeIntoSegments(story.code);
+	const segmentHtml = segments.map((seg) => renderSegmentHtml(seg, queryData)).join('\n');
+	const timestamp = `<footer style="margin-top:48px;padding-top:16px;border-top:1px solid #e5e7eb;font-size:12px;color:#9ca3af">Generated on ${new Date().toLocaleDateString('en-US', { year: 'numeric', month: 'long', day: 'numeric' })}</footer>`;
+	return wrapHtmlDocument(story.title, `${segmentHtml}\n${timestamp}`);
+}
+
+function wrapHtmlDocument(title: string, body: string): string {
```
Bl3f: Can't this be a React component so it's easier to maintain?

apps/backend/src/trpc/shared-story.routes.ts
```diff
@@ -206,4 +208,48 @@ export const sharedStoryRoutes = {
 
 		await sharedStoryQueries.deleteSharedStory(input.id);
 	}),
+
+	download: protectedProcedure
```
Bl3f: should it be a projectProtectedProcedure? all the protections should be reviewed and a bit factorised because I feel most of this code is the same accross all routes.

Bl3f (approved):
LGTM

PR #577, opened 2026-04-08
fix: detect wrong-platform binary in nao chat and show actionable error (#287)

### LINKED ISSUE #287: nao chat fails on Windows — Linux binary shipped with no Windows equivalent
**Platform:** Windows 11
  **nao-core version:** 0.0.34
  **Python version:** 3.10

  ## Problem
  Running `nao chat` on Windows fails with:

      [WinError 193] %1 is not a valid Win32 application

  ## Root Cause
  The `nao-chat-server` binary shipped in `nao_core/bin/` is a Linux ELF binary.
  There is no `.exe` equivalent bundled for Windows. The same applies to the `rg` binary.

  ## Steps to Reproduce
  1. Install `nao-core` via pip on Windows
  2. Run `nao init` and `nao sync`
  3. Run `nao chat`
  4. Error occurs immediately after FastAPI starts

  ## Expected Behavior
  `nao chat` should work on Windows, or the docs should clearly state
  that Windows requires Docker/WSL.

  ## Workaround
  Using Docker or WSL to run the Linux binary works around the issue.

Abhijitam01:
## Summary

Fixes #287 — `nao chat` fails on Windows with `[WinError 193] %1 is not a valid Win32 application` when the wrong-platform binary is shipped in the wheel

The root cause (Linux binary shipped in the Windows wheel) was already fixed in CI since v0.1.4 — the per-platform build matrix makes cross-contamination impossible. This PR adds the guardrails so if it ever happens again, users get an actionable error and CI catches it before release.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -100,12 +100,12 @@ jobs:
               working-directory: cli
               run: |
                   if [ "${{ matrix.platform }}" = "win" ]; then
-                    ./nao_core/bin/nao-chat-server.exe --help
-                    STATUS=$?
+                    STATUS=0
+                    ./nao_core/bin/nao-chat-server.exe --help || STATUS=$?
                   else
                     chmod +x ./nao_core/bin/nao-chat-server
-                    ./nao_core/bin/nao-chat-server --help
-                    STATUS=$?
+                    STATUS=0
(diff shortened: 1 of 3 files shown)


Abhijitam01:
@Bl3f please review this

Bl3f:
Not sure it's useful.

PR #595, opened 2026-04-10
fix(cli): add Vertex AI handler for ai_summary template (#584)

### LINKED ISSUE #584: [bug] AI summary doesn't work with Vertex AI
## Problem

AI summary feature does not work when using Vertex AI as the model provider. The summary generation fails or produces no output when Vertex AI is configured.

## Steps to reproduce

1. Configure nao to use Vertex AI as the model provider
2. Run a conversation or data pipeline
3. Trigger AI summary generation
4. Observe that the summary fails or is empty

## Expected behavior

AI summary should work correctly regardless of the model provider (Vertex AI, OpenAI, etc.).

## Impact

Users relying on Vertex AI cannot use the AI summary feature, which is a core capability.

Rish-it:
The `prompt()` helper used by the `ai_summary` Jinja template dispatched to
provider-specific handlers but lacked a branch for `LLMProvider.VERTEX`. As a
result, all Vertex-configured projects fell through to:

`Unsupported LLM provider 'vertex' for ai_summary generation.`

Fixes getnao/nao#584

Screenshot 2026-04-10 at 3 35 44 PM

### DIFF
--- a/cli/nao_core/templates/engine.py
+++ b/cli/nao_core/templates/engine.py
@@ -123,6 +123,8 @@ class TemplateEngine:
                 return self._generate_ollama(model, prompt_text)
             if self.llm_config.provider == LLMProvider.BEDROCK:
                 return self._generate_bedrock(model, prompt_text)
+            if self.llm_config.provider == LLMProvider.VERTEX:
+                return self._generate_vertex(model, prompt_text)
         except ImportError as e:
             raise RuntimeError(
                 f"Provider '{self.llm_config.provider.value}' is not available in this environment: {e}"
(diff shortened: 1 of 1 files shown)


cli/nao_core/templates/engine.py
```diff
@@ -285,6 +287,120 @@ def _generate_bedrock(self, model: str, prompt_text: str) -> str:
...
+        credentials = self._build_vertex_credentials()
+
+        # Claude family on Vertex → Anthropic SDK.
+        if model.startswith("claude-"):
+            return self._generate_vertex_anthropic(model, prompt_text, project, location, credentials)
+
+        # Everything else (Gemini family) → google-genai.
+        return self._generate_vertex_gemini(model, prompt_text, project, location, credentials)
+
+    def _generate_vertex_gemini(
```
socallmebertille, line 329: I think we should refactor this method so that a utility method could be used by  _generate_vertex and _generate_gemini

cli/nao_core/templates/engine.py
```diff
@@ -285,6 +287,120 @@ def _generate_bedrock(self, model: str, prompt_text: str) -> str:
...
+            contents=prompt_text,
+            config=types.GenerateContentConfig(temperature=0),
+        )
+
+        content = getattr(response, "text", None)
+        if not content:
+            raise RuntimeError("Empty response from model.")
+        return str(content).strip()
+
+    def _generate_vertex_anthropic(
```
socallmebertille, line 345: Same for this one with _generate_vertex_anthropic and _generate_anthropic

socallmebertille (changes requested):

Rish-it:
@socallmebertille addressed the refactor in `9a88484`.

- Extracted `_run_genai_client` and `_run_anthropic_messages`
- Reused across direct and Vertex Gemini / Anthropic providers

Let me know if anything else needs changes.

socallmebertille (approved):
Thanks for your contribution and your changes ! LGTM :)

PR #598, opened 2026-04-11
Abhijitam01/notion easy add

### LINKED ISSUE #594: [feature] Easier way to add Notion pages without editing config for each one
## Problem

Currently, adding a new Notion page to the nao context requires manually appending it to the config file. For customers who want to regularly bring in new Notion pages as context, this is friction-heavy and scales poorly.

## Proposal

Provide a lighter-weight way to register Notion pages as context — options to consider:
- Support page patterns / glob-like selectors (e.g. all pages under a parent, all pages in a workspace section)
- A CLI command to add a page by URL without hand-editing the YAML
- Auto-discover pages from a root page

## Impact

Customers using nao's Notion integration (highlighted as one of the most-loved features) want to iterate faster on what's in their context without touching config.

## Reported by

Customer feedback (Manuel, open-source user).

Abhijitam01:
Added nao add notion command that lets users register Notion pages by URL or ID without hand-editing nao_config.yaml
Added nao remove notion command to remove pages the same way

Extracted shared extract_page_id() into nao_core/config/notion/ so both commands and the existing sync provider reuse the same parsing logic

Handles duplicates, invalid input, missing config, and preserves YAML formatting (comments, env var templates like ${{ env('NOTION_API_KEY') }})

fixes #594


cli/nao_core/commands/add/notion.py
```diff
@@ -0,0 +1,164 @@
...
+            page_id = extract_page_id(page)
+        except ValueError:
+            has_errors = True
+            UI.error(f"Invalid Notion page URL or ID: {page}")
+            continue
+
+        if page_id in existing_ids:
+            UI.warn(f"Page {page_id} already configured, skipping.")
+            continue
+
```
socallmebertille, line 111: It would be good to add a check to ensure that the page is accessible from the integration, so that its ID isn't added to the nao_config.yaml file in that case.

This check could call a method such as :

```
def _verify_page_accessible(page_id: str, api_key: str) -> bool:
    """Return True if the Notion page is reachable with the given API key."""
    try:
        from notion_client import Client

        client = Client(auth=api_key)
        client.pages.retrieve(page_id=page_id)
        return True
    except Exception:
        return False
```

And be added like :

```
if not _verify_page_accessible(page_id, resolved_key):
    has_errors = True
    UI.error(
        f"Page {page_id} is not accessible. Make sure the page is shared with your Notion integration."
    )
    continue
```

cli/nao_core/commands/remove/notion.py
```diff
@@ -0,0 +1,102 @@
...
+
+@track_command("remove_notion")
+def notion(
+    pages: Annotated[
+        list[str],
+        Parameter(
+            help="Notion page URLs or IDs to remove.",
+            show=True,
+        ),
+    ],
```
socallmebertille, line 28: Perhaps we should launch a sync that runs automatically when a notion page is removed, as it's already the case with the add command.

cli/nao_core/commands/add/notion.py
```diff
@@ -0,0 +1,164 @@
...
+            pass
+    return ids
+
+
+def _run_notion_sync() -> None:
+    """Run nao sync for notion provider only."""
+    from nao_core.commands.sync import sync
+
+    sync(provider=["notion"], render_templates=False)
+
```
socallmebertille, line 32: The shared helpers above the main add notion function currently live in `commands/add/notion.py` and get imported by `commands/remove/notion.py`, which creates an awkward coupling where `remove` depends on `add`. These are really config-level utilities — I'd move them into `config/notion/` alongside `extract_page_id` and `NotionConfig`, which is already the canonical place for Notion config logic (and already imported by the sync provider). This would also be the natural home for a shared `_resolve_page_ids()` and `_sync_if_possible()` to deduplicate the validation loop and post-sync extras check that are currently copy-pasted between add and remove.

cli/nao_core/config/notion/__init__.py
socallmebertille, line 1: The shared helpers between the `add` and `remove` commands should be there.

CLAUDE.md
```diff
@@ -44,3 +44,23 @@ These are useful scripts to run after changing the database schemas:
 ### Testing, Linting
 
 - Always end by verifying your work with `npm run lint` for the `apps/` or `make lint` for the `cli/`
+
```
socallmebertille, line 51: At the moment, this Skill routing section doesn’t seem to provide value to the repository. It references a set of external “skills” (e.g., investigate, ship, qa) that are not implemented or integrated into the project, so the configuration remains purely declarative and has no practical effect.

cli/tests/nao_core/commands/remove/test_notion.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def test_remove_multiple_pages(self, config_dir: Path):
+        with patch("nao_core.commands.remove.notion._find_config_path", return_value=config_dir / "nao_config.yaml"):
+            from nao_core.commands.remove.notion import notion
+
+            notion(pages=["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa", "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"])
+
+        content = _read_config(config_dir)
+        assert "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" not in content
+        assert "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" not in content
```
socallmebertille, line 138: To go along with the requested changes, add a `test_sync_runs_after_remove()` function.

cli/tests/nao_core/commands/add/test_notion.py
```diff
@@ -0,0 +1,194 @@
...
+                notion(
+                    pages=["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"],
+                    sync=False,
+                    api_key="test-key-456",
+                )
+
+        content = _read_config(config_notion_null)
+        assert "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" in content
+
+    def test_sync_skipped_when_extras_missing(self, config_dir: Path):
```
socallmebertille, line 168: We never test that a sync is successfully done.

cli/tests/nao_core/commands/add/test_notion.py
```diff
@@ -0,0 +1,194 @@
...
+
+import textwrap
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+
+@pytest.fixture()
+def config_dir(tmp_path: Path) -> Path:
```
socallmebertille, line 15: This function could be extracted in a `conftest.py` file so the add and the remove tests can use it. It is also the case for the `_read_config()` function below.

cli/tests/nao_core/commands/add/test_notion.py
```diff
@@ -0,0 +1,194 @@
...
+
+        content = _read_config(config_dir)
+        assert "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb" in content
+
+    def test_add_duplicate_skipped(self, config_dir: Path):
+        with patch("nao_core.commands.add.notion._find_config_path", return_value=config_dir / "nao_config.yaml"):
+            with patch("nao_core.commands.add.notion._run_notion_sync"):
+                from nao_core.commands.add.notion import notion
+
+                notion(pages=["aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"], sync=False)
```
socallmebertille, line 71: Please be consistent with the location of the import call.

cli/tests/nao_core/commands/add/test_notion.py
```diff
@@ -0,0 +1,194 @@
...
+def _read_config(config_dir: Path) -> str:
+    return (config_dir / "nao_config.yaml").read_text()
+
+
+class TestAddNotion:
+    def test_add_single_page(self, config_dir: Path):
+        from nao_core.commands.add.notion import notion
+
+        with patch("nao_core.commands.add.notion._find_config_path", return_value=config_dir / "nao_config.yaml"):
+            with patch("nao_core.commands.add.notion._run_notion_sync"):
```
socallmebertille: With `sync=False`, the second patch will never be call according to your implementation of the add command that contains a `if sync and added:` before launching the sync command. So it can me removed to avoid noise.

Same for other few tests.

socallmebertille (changes requested):
Thanks for your contribution, the PR is mostly great, if you could just make the requested  changes and it would be ok for me!

Abhijitam01:
@socallmebertille can you look into it , i have removed what is unecesary and fixed the bug

cli/nao_core/config/notion/__init__.py
```diff
@@ -1,7 +1,103 @@
...
+    from ruamel.yaml import YAML
+
+    yaml = YAML()
+    yaml.preserve_quotes = True
+    with config_path.open() as f:
+        data = yaml.load(f)
+    if data is None:
+        return {}
+    if not isinstance(data, dict):
+        raise ValueError("nao_config.yaml must contain a mapping at the top level.")
```
socallmebertille: The `raise` should be catch in a `try/catch` where it's been called or just return an empty object.

I don't know why you've changed this function but not the `save_yaml` that had the same structure, you should be consistent.

cli/tests/nao_core/commands/add/test_notion.py
socallmebertille: You didn't proceed to those changes.

cli/nao_core/config/notion/__init__.py
```diff
@@ -1,7 +1,103 @@
...
 from pydantic import BaseModel, Field
 
 from nao_core.ui import UI, ask_text
 
+# Notion page IDs are 32-character hex strings (UUID without dashes).
+# Anchored: matches raw IDs, IDs at the end of Notion URLs, or IDs followed by query params.
+NOTION_PAGE_ID_PATTERN = re.compile(r"(?:^|[-/])([a-f0-9]{32})(?:[?#]|$)")
+
+
+def extract_page_id(page_url: str) -> str:
```
socallmebertille, line 13: The file `cli/nao_core/templates/context.py` still import :
```
from nao_core.commands.sync.providers.notion.provider import (
    extract_page_id,
    get_page_title,
    strip_images,
)
```

Please turn it into :
```
from nao_core.config.notion import extract_page_id
from nao_core.commands.sync.providers.notion.provider import (
    get_page_title,
    strip_images,
)
```

socallmebertille (approved):
LGMT 👍

socallmebertille:
@Abhijitam01 To fix the import sort error from CI workflow, please run :
```
cd cli
source .venv/bin/activate
uv run ruff check --select I --fix .
```

PR #600, opened 2026-04-11
fix: show full error messages in debug table instead of truncating to 80 chars

### LINKED ISSUE #580: In debug if there is an error it should be visible in the table

Aaryan-549:
Closes #580

  Previously, error messages in `nao debug` were truncated to 80 characters, which often cut off the most useful part of the error. This
  change removes the truncation and highlights errors in red so they stand out in the table.

  Changes:
  - Removed 80-char truncation on error messages for both database and LLM tables
  - Styled error text in red for better visibility
  - Rich table handles wrapping naturally for long messages

### DIFF
--- a/cli/nao_core/commands/debug.py
+++ b/cli/nao_core/commands/debug.py
@@ -163,13 +163,11 @@ def debug():
                 )
             else:
                 console.print("[bold red]✗[/bold red]")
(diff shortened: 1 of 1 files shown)


Bl3f (approved):
LGTM

PR #604, opened 2026-04-11
fix: handle LLM quota errors gracefully instead of crashing the UI

### LINKED ISSUE #581: when there is a quota issue (openai) the UI is crashing

Aaryan-549:
Closes #581

  When an LLM provider returns a quota or rate limit error, the UI was crashing instead of showing a friendly error message. Two issues were
  causing this:

  1. **Backend**: If `handleAgentRoute()` threw before creating a stream (e.g. model resolution failure, quota error), Fastify returned a
  JSON error response instead of an SSE stream. The frontend's SSE parser couldn't handle this and threw an unhandled error.

  2. **Frontend**: No React Error Boundary existed anywhere in the app. Any unhandled rendering error during error state processing would
  crash the entire component tree.

  Changes:
  - Backend: Wrapped the agent route handler with try/catch that returns a proper SSE error stream on failure, so the frontend always
  receives a parseable response
  - Frontend: Added a ChatErrorBoundary around the chat messages component to catch rendering errors and show a fallback UI instead of
  crashing

### DIFF
--- a/apps/backend/src/routes/agent.ts
+++ b/apps/backend/src/routes/agent.ts
@@ -1,12 +1,18 @@
-import { createUIMessageStreamResponse } from 'ai';
+import { createUIMessageStream, createUIMessageStreamResponse } from 'ai';
 
 import type { App } from '../app';
 import { handleAgentRoute } from '../handlers/agent';
 import { authMiddleware } from '../middleware/auth';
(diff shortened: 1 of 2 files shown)


Bl3f:
CI fails, feel free to re-open when fixed.

PR #606, opened 2026-04-11
feat: auto-detect warehouse dialect and inject syntax-specific SQL rules into system prompt

### LINKED ISSUE #590: [feature] Auto-detect warehouse dialect and inject syntax-specific RULES
## Problem

When connected to a T-SQL warehouse (e.g. Microsoft Fabric), the LLM generates SQL syntax that is not valid for T-SQL:
- Uses `LIMIT` instead of `TOP`
- Uses `GROUP BY ALL` which is not valid T-SQL

The LLM eventually self-corrects after errors, but this wastes tokens and degrades UX.

## Steps to reproduce

1. Connect nao to a Microsoft Fabric warehouse
2. Ask a question that requires aggregation or row limits
3. LLM generates PostgreSQL-style SQL (`LIMIT`, `GROUP BY ALL`)
4. Query fails, LLM retries with corrected syntax

## Expected behavior

nao should detect the warehouse dialect from the connection type and automatically inject dialect-specific RULES so the LLM generates valid SQL on the first attempt.

Example RULES for Fabric/T-SQL:
- Use `TOP N` instead of `LIMIT N`
- Do not use `GROUP BY ALL`
- Use T-SQL date functions (`DATEADD`, `DATEDIFF`) instead of PostgreSQL-style intervals

## Impact

Affects any user on Microsoft Fabric / Synapse / SQL Server warehouses. Every query wastes an extra retry cycle. Reported by an integrator partner testing Fabric PoCs with clients.

## Notes

The reporter noted: "il faudrait peut-être lui expliciter dans le RULES la syntaxe compatible avec le Warehouse utilisé" — we should automate this based on connection type rather than requiring manual RULES configuration.

## Reported by

Customer (integrator partner)

Aaryan-549:
Closes #590

  When connected to T-SQL warehouses (Fabric/MSSQL), the LLM was generating PostgreSQL-style SQL — using LIMIT instead of TOP, GROUP BY ALL,
  and PostgreSQL date functions. This caused queries to fail and waste tokens on retries.

  The system prompt now auto-detects the warehouse dialect from the connection type and injects dialect-specific SQL rules:

  **T-SQL (Fabric/MSSQL):**
  - Use TOP N instead of LIMIT N
  - Do not use GROUP BY ALL — list columns explicitly
  - Use T-SQL date functions (DATEADD, DATEDIFF, CONVERT, FORMAT)
  - Use ISNULL() instead of COALESCE() for two arguments

  **BigQuery:**
  - Use backtick-quoted identifiers
  - Use SAFE_DIVIDE for division

  **MySQL:**
  - Use backtick-quoted identifiers
  - Use IFNULL() instead of COALESCE() for two arguments

  Rules are only injected when the relevant database type is detected in the project connections. PostgreSQL and other standard SQL dialects
  don't get extra rules since the LLM defaults to standard SQL.

### DIFF
--- a/apps/backend/src/components/ai/system-prompt.tsx
+++ b/apps/backend/src/components/ai/system-prompt.tsx
@@ -24,6 +24,9 @@ export const MEMORY_TOKEN_LIMIT = 1000;
 export function SystemPrompt({ memories = [], userRules, connections = [], skills = [], timezone }: SystemPromptProps) {
 	const visibleMemories = getMemoriesInTokenRange(memories, MEMORY_TOKEN_LIMIT);
 	const hasClickHouse = connections.some((connection) => connection.type.toLowerCase() === 'clickhouse');
+	const hasTSQL = connections.some((connection) => ['mssql', 'fabric'].includes(connection.type.toLowerCase()));
+	const hasBigQuery = connections.some((connection) => connection.type.toLowerCase() === 'bigquery');
+	const hasMySQL = connections.some((connection) => connection.type.toLowerCase() === 'mysql');
 
 	return (
 		<Block>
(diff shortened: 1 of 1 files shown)


Bl3f (approved):
LGTM

PR #611, opened 2026-04-12
feat: import Snowflake semantic views during sync

### LINKED ISSUE #582: import Snowflake semantic views objects

Aaryan-549:
Closes #582

  Snowflake semantic views are now imported during `nao sync` alongside regular tables. They provide a semantic layer (metrics, dimensions,
  relationships) that helps the LLM understand the correct way to query data.

  Changes:
  - Added `get_semantic_views()` method to `SnowflakeConfig` that queries `INFORMATION_SCHEMA.SEMANTIC_VIEWS`
  - Added base `get_semantic_views()` method to `DatabaseConfig` (returns empty for non-Snowflake databases)
  - After syncing tables for each schema, the sync provider now fetches and writes semantic views to `semantic_view=<name>/definition.md`
  - Gracefully handles Snowflake instances where semantic views are not available (silently skips)

  File structure:
  databases/type=snowflake/database=/schema=/
    table=/columns.md, preview.md, ...
    semantic_view=/definition.md     ← new

  ## Testing

  I don't have access to a Snowflake account, so I couldn't test this end-to-end against a live instance. Here's what I did verify:

  - **Mock-tested the `get_semantic_views()` method** — confirmed it correctly parses rows from `INFORMATION_SCHEMA.SEMANTIC_VIEWS` into `{name, comment, definition}` dicts
  - **Verified graceful fallback** — when the query fails (e.g. older Snowflake without semantic views), the method catches the exception and returns an empty list instead of crashing the sync
  - **Verified no impact on other databases** — the base `DatabaseConfig.get_semantic_views()` returns empty, so non-Snowflake syncs are unaffected
  - **All 479 existing CLI tests pass**, ruff lint and format clean

  Would appreciate if someone on the team with Snowflake access could validate the `INFORMATION_SCHEMA.SEMANTIC_VIEWS` query and the output file structure.

### DIFF
--- a/cli/nao_core/commands/sync/providers/databases/provider.py
+++ b/cli/nao_core/commands/sync/providers/databases/provider.py
@@ -177,9 +177,9 @@ def sync_database(
         for schema, tables in schema_tables.items():
             schema_path = db_path / f"schema={schema}"
             schema_path.mkdir(parents=True, exist_ok=True)
-            state.add_schema(schema)
 
             if tables:
+                state.add_schema(schema)
                 table_task = progress.add_task(
                     f"    [cyan]{schema}[/cyan]",
                     total=len(tables),
--- a/cli/nao_core/commands/sync/providers/databases/provider.py
+++ b/cli/nao_core/commands/sync/providers/databases/provider.py
@@ -159,15 +159,12 @@ def sync_database(
 
             tables = [t for t in all_tables if db_config.matches_pattern(schema, t)]
 
-            if not tables:
-                progress.update(schema_task, advance=1)
-                continue
-
-            list_dur = _fmt_duration(time.monotonic() - t_list)
-            console.print(
-                f"  [cyan]▸ {schema}[/cyan] [dim]— {len(tables)} tables "
-                f"(of {len(all_tables)} total, listed in {list_dur})[/dim]"
-            )
+            if tables:
+                list_dur = _fmt_duration(time.monotonic() - t_list)
+                console.print(
+                    f"  [cyan]▸ {schema}[/cyan] [dim]— {len(tables)} tables "
+                    f"(of {len(all_tables)} total, listed in {list_dur})[/dim]"
+                )
             schema_tables[schema] = tables
 
         selected_tables = [(schema, t) for schema, tables in schema_tables.items() for t in tables]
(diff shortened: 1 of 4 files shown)


cli/nao_core/config/databases/snowflake.py
```diff
@@ -257,6 +257,28 @@ def get_schemas(self, conn: BaseBackend) -> list[str]:
     def create_context(self, conn: BaseBackend, schema: str, table_name: str) -> SnowflakeDatabaseContext:
         return SnowflakeDatabaseContext(conn, schema, table_name)
 
+    def get_semantic_views(self, conn: BaseBackend, schema: str) -> list[dict[str, str]]:
+        """Fetch semantic views from a Snowflake schema."""
+        try:
+            query = f"""
+                SELECT SEMANTIC_VIEW_NAME, COMMENT, DEFINITION
```
socallmebertille: Semantic views expose their metadata with the property `NAME` (not `SEMANTIC_VIEW_NAME`) and `COMMENT`.

The full definition is not stored as a `DEFINITION column`, but can be retrieved using the `GET_DDL` function, which reconstructs the exact SQL definition used to create the object. `GET_DDL` is a Snowflake function that requires literal string arguments. So, it must be called once per view.

cli/nao_core/config/databases/snowflake.py
```diff
@@ -257,6 +257,28 @@ def get_schemas(self, conn: BaseBackend) -> list[str]:
...
+                FROM INFORMATION_SCHEMA.SEMANTIC_VIEWS
+                WHERE SEMANTIC_VIEW_SCHEMA = '{schema}'
+                ORDER BY SEMANTIC_VIEW_NAME
```
socallmebertille: The correct syntaxe is : 
```
WHERE "SCHEMA" = '{schema}'
ORDER BY NAME
```

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -253,6 +253,21 @@ def sync_database(
...
+            if semantic_views:
+                for sv in semantic_views:
+                    sv_path = schema_path / f"semantic_view={sv['name']}"
+                    sv_path.mkdir(parents=True, exist_ok=True)
+                    content_parts = [f"# {sv['name']}\n"]
+                    if sv.get("comment"):
+                        content_parts.append(f"{sv['comment']}\n")
+                    if sv.get("definition"):
+                        content_parts.append("## Definition\n")
+                        content_parts.append(f"```yaml\n{sv['definition']}\n```\n")
```
socallmebertille: In Snowflake, the definition of a semantic view is more a sql query than a yaml file, you should replace `yaml` by `sql`.

cli/nao_core/commands/sync/providers/databases/provider.py
socallmebertille: You can remove this, because it prevents the following code to be executed and with it the part with the semantic view implementation.

socallmebertille: To make sense with the previous comment, put this bloc into a `if tables:`.

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -253,6 +253,21 @@ def sync_database(
                 f"  [green]✓ {schema}[/green] [dim]— {len(tables)} tables synced in {schema_dur}{error_suffix}[/dim]"
             )
```
socallmebertille: Same for this block, put it into a `if tables:` condition.

cli/nao_core/config/databases/snowflake.py
```diff
@@ -257,6 +257,28 @@ def get_schemas(self, conn: BaseBackend) -> list[str]:
...
+            """
+            rows = conn.raw_sql(query).fetchall()  # type: ignore[union-attr]
+            return [
+                {
+                    "name": row[0],
+                    "comment": str(row[1]) if row[1] else "",
+                    "definition": str(row[2]) if row[2] else "",
+                }
+                for row in rows
+            ]
```
socallmebertille: To use the `GET_DDL` function I mentioned earlier, you can implement the extraction as follows :

```
results = []
for name, comment in rows:
    try:
        ddl_row = conn.raw_sql(f"SELECT GET_DDL('SEMANTIC VIEW', '{schema}.{name}')").fetchone()  # type: ignore[union-attr]
        definition = str(ddl_row[0]) if ddl_row and ddl_row[0] else ""
    except Exception:
        definition = ""
    results.append({"name": name, "comment": str(comment) if comment else "", "definition": definition})
return results
```

socallmebertille (changes requested):
Hi, thanks for your contribution, please find the requested changes below. 👍

Aaryan-549:
@socallmebertille Thanks for reviewing. Made all the necessary changes.

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -178,80 +175,98 @@ def sync_database(
...
-            schema_start = time.monotonic()
-
-            for table in tables:
-                table_path = schema_path / f"table={table}"
-                table_path.mkdir(parents=True, exist_ok=True)
-
-                progress.update(
-                    table_task,
-                    description=f"    [cyan]{schema}[/cyan] [dim]→ {table}[/dim]",
+            if tables:
```
socallmebertille: This if shouldn't be there, please remove it along with the indentation.

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -178,80 +175,98 @@ def sync_database(
...
+                    table_task,
+                    description=f"    [cyan]{schema}[/cyan]",
+                )
+                schema_dur = _fmt_duration(time.monotonic() - schema_start)
+                error_suffix = f" [red]({schema_errors} errors)[/red]" if schema_errors else ""
+                console.print(
+                    f"  [green]✓ {schema}[/green] [dim]— {len(tables)} tables synced in {schema_dur}{error_suffix}[/dim]"
+                )
+
+            semantic_views = db_config.get_semantic_views(conn, schema)
```
socallmebertille: Also add this to take in consideration the `include` and `exclude` filters from `nao_config.yaml` :
```
semantic_views = [sv for sv in semantic_views if db_config.matches_pattern(schema, sv["name"])]
```

cli/nao_core/commands/sync/providers/databases/provider.py
```diff
@@ -178,80 +175,98 @@ def sync_database(
...
+            if semantic_views:
+                for sv in semantic_views:
+                    sv_path = schema_path / f"semantic_view={sv['name']}"
+                    sv_path.mkdir(parents=True, exist_ok=True)
+                    content_parts = [f"# {sv['name']}\n"]
```
socallmebertille, line 274: Please add this description : 
```
content_parts.append(
    "This is a Snowflake **semantic view** — use this to understand the intended way to query and aggregate data.\n"
)
```

socallmebertille (dismissed):
I didn't think about that before but it would be great if you could implement a tracking of the semantic views in `cli/nao_core/commands/sync/cleanup.py` like : 
```
class DatabaseSyncState:
    def add_semantic_view(self, schema: str, name: str) -> None:
```
And remove semantic views that weren't synced in the same file in `def cleanup_stale_paths(state: DatabaseSyncState, verbose: bool = False) -> int:
`

Issue #560, opened 2026-04-04
[Bedrock] Support inference profile ARNs + memory extraction ignores custom model ARN


LaurentBorel:
Here's below Claude's sum-up of the workaround he's come up with to make bedrock work with ARN's profile
HTH

Laurent

 ---                                                                                                                                    
  Environment                                                                                                                            
  - nao-core version: 0.0.37                                                                                                             
  - LLM provider: AWS Bedrock                                                                                                            
  - Auth method: AWS SSO (aws sso login) with temporary credentials                                                                      
  - OS: macOS                                                                                                                            
                                                                                                                                         
  ---                                                                                                                                    
  Issue 1 — SSO credentials not propagated to subprocesses
                                                          
  Problem
                                                                                                                                         
  When authenticating with AWS SSO (aws sso login --profile <profile>), nao chat launches two subprocesses (FastAPI + Bun chat server).  
  These subprocesses fail to refresh SSO credentials independently because chat.py only injects AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY 
  and AWS_REGION — but not AWS_SESSION_TOKEN, which is required for temporary SSO credentials.                                           
                  
  Additionally, if the AWS config uses the legacy SSO profile format (no sso-session block), boto3's ambient credential chain            
  (AWS_PROFILE) fails with:
  Credentials were refreshed, but the refreshed credentials are still expired.                                                           
                                                                                                                                         
  Workaround                                                                                                                             
                                                                                                                                         
  Users must manually export credentials before running nao chat:                                                                        
  aws sso login --profile <profile>                                                                                                      
  eval "$(aws configure export-credentials --profile <profile> --format env)"                                                            
  nao chat                                                                                                                               
                                                                                                                                         
  Fix needed in commands/chat.py                                                                                                         
                                                                                                                                         
  Add AWS_SESSION_TOKEN propagation in the Bedrock env injection block:                                                                  
  if config.llm.provider == LLMProvider.BEDROCK:
      ...                                                                                                                                
      session_token = os.environ.get("AWS_SESSION_TOKEN")                                                                                
      if session_token:                                                                                                                  
          env["AWS_SESSION_TOKEN"] = session_token                                                                                       
                                                                                                                                         
  ---                                                                                                                                    
  Issue 2 — Inference profile ARNs work for chat but not for memory extraction
                                                                                                                                         
  Problem         
                                                                                                                                         
  When using an Application Inference Profile ARN as the model (e.g.                                                                     
[...]

PR #615, opened 2026-04-14
all issues due to config bedrock resolved

socallmebertille:
## Summary

- [x]  **Fix SSO authentication for Bedrock** — AWS SSO credentials (profile, session token) were not propagated to subprocesses, causing auth failures. Added fromNodeProviderChain as a fallback so SSO, IAM roles, and named profiles work out of the box.

- [x] **Fix background tasks (memory, compaction, title) failing with 403** — These tasks used a hardcoded model ID that restrictive IAM policies don't allow. They now resolve the model from NAO_ANNOTATION_MODEL env var or the project's enabled models before falling back to the provider default.

- [x] **Fix wrong regional prefix for cross-region inference** — resolveBedrockModelId now replaces an incorrect geographic prefix (e.g. us. → eu.) instead of keeping it as-is.

## Related issue

#560

### DIFF
--- a/apps/backend/package.json
+++ b/apps/backend/package.json
@@ -43,6 +43,7 @@
 		"@ai-sdk/google-vertex": "^4.0.80",
 		"@ai-sdk/mistral": "^3.0.13",
 		"@ai-sdk/openai": "^3.0.1",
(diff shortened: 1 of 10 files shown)


Bl3f (approved):
LGTM

PR #625, opened 2026-04-16
Feat/ask selection from own thread

socallmebertille:
## Summary

- Add a selection option from chats and stories for the user to ask the agent from a particular range of text

1.1. From a story in chat
image

1.2. From a chat
image

2. Pending citation
image

3. Go to original selection
image

4. Also works in chat replay
image


Bl3f (approved):
LGTM

Issue #564, opened 2026-04-07
Left panel - order chats by last message date instead of creation date


ClaireGz:
On nao UI, in the left panel, convesations are sorted from most recent created to least recent.
If I write again in a previous conversation, I want it to appear on top - ie that my conversations are sorted by date of last message of conversation

Prasad8830:
Hi @ClaireGz, can I take this issue? I am new to this repository, so I would like to start with an easy issue.

PR #630, opened 2026-04-17
Feat/filter chat list item

socallmebertille:
## Summary

- [x] Title re-generated with context when chat is from a selection / ask content
- [x] Chat list ordonated by updatedAt date
- [x] `Show more` button can be clicked to see more than 10 chats and `Show less` to revert that
- [x] Chat list cann be grouped by stars, date, projects, ownership or none and filtered
image
image
image

## Related issue

#564

### DIFF
--- a/apps/backend/src/handlers/agent.ts
+++ b/apps/backend/src/handlers/agent.ts
@@ -61,9 +61,13 @@ export const handleAgentRoute = async (opts: HandleAgentMessageInput): Promise<H
 
 	const agent = await agentService.create({ ...chat, userId, projectId }, model);
 
+	const isFirstForkMessage =
+		!isNewChat && !!chat.forkMetadata && chat.messages.filter((m) => m.role === 'user' && !m.isForked).length === 1;
+
 	const stream = agent.stream(chat.messages, {
 		mentions,
 		timezone: opts.timezone,
+		generateTitle: isFirstForkMessage,
 		events: {
 			newChat: isNewChat
 				? {
--- a/apps/backend/src/handlers/agent.ts
+++ b/apps/backend/src/handlers/agent.ts
@@ -51,7 +51,7 @@ export const handleAgentRoute = async (opts: HandleAgentMessageInput): Promise<H
 		newMessageId = messageId;
 	}
 
-	const [chat] = await chatQueries.loadChat(chatId);
+	const [chat] = await chatQueries.getChat(chatId);
 	if (!chat) {
 		throw new HandlerError('NOT_FOUND', `Chat with id ${chatId} not found.`);
 	}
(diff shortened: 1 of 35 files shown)


apps/backend/src/handlers/agent.ts
```diff
@@ -61,9 +61,13 @@ export const handleAgentRoute = async (opts: HandleAgentMessageInput): Promise<H
 
 	const agent = await agentService.create({ ...chat, userId, projectId }, model);
 
+	const isFirstForkMessage =
```
Bl3f: This is badly named in English i'd say. Would be better to name it isFirstMessageForked or isForkedFirstMessage.

Tho still I think it should be eventually done differently.

apps/backend/src/handlers/agent.ts
```diff
@@ -61,9 +61,13 @@ export const handleAgentRoute = async (opts: HandleAgentMessageInput): Promise<H
...
+		!isNewChat && !!chat.forkMetadata && chat.messages.filter((m) => m.role === 'user' && !m.isForked).length === 1;
+
 	const stream = agent.stream(chat.messages, {
 		mentions,
 		timezone: opts.timezone,
+		generateTitle: isFirstForkMessage,
```
Bl3f: Same here, I'd name this param `shouldGenerateTitle` as it's a boolean it's better to prefix it with is or should so we understand better when we read the variable.

Tho I think the pattern is a bit ugly and there is just below the events.newChat param, I think if possible it would be better to consider this case as a new chat also so we don't change the stream method?

apps/backend/src/queries/chat.queries.ts
```diff
@@ -28,24 +21,124 @@ export const checkChatExists = async (chatId: string): Promise<boolean> => {
...
+		fetchOwnChats(userId),
+		needsShared ? fetchSharedWithMeChats(userId, projectId) : Promise.resolve([]),
+	]);
+
+	const allItems = [...ownItems, ...sharedItems];
+	const filtered = applyChatFilters(allItems, effective);
+	return { groups: buildChatGroups(filtered, groupBy) };
+};
+
+async function fetchOwnChats(userId: string): Promise<EnrichedChat[]> {
```
Bl3f: We need some convention between using list, fetch, load, get word when we just retrieve items from the db and unify the namings.

apps/backend/src/queries/chat.queries.ts
```diff
@@ -28,24 +21,124 @@ export const checkChatExists = async (chatId: string): Promise<boolean> => {
...
+		.leftJoin(
+			s.sharedChatAccess,
+			and(eq(s.sharedChatAccess.sharedChatId, s.sharedChat.id), eq(s.sharedChatAccess.userId, userId)),
+		)
+		.where(and(eq(s.chat.projectId, projectId), isNull(s.chat.deletedAt)))
+		.orderBy(desc(s.chat.updatedAt))
+		.execute();
+
+	return rows
+		.filter((r) => {
```
Bl3f: Why this filter is not a where?

apps/backend/src/queries/chat.queries.ts
```diff
@@ -28,24 +21,124 @@ export const checkChatExists = async (chatId: string): Promise<boolean> => {
...
+			isStarred: false,
+			createdAt: r.chatCreatedAt.getTime(),
+			updatedAt: r.chatUpdatedAt.getTime(),
+			kind: 'shared' as const,
+			shareId: r.shareId,
+			ownerName: r.ownerName,
+			projectId: r.projectId,
+			projectName: r.projectName,
+			isSharedByMe: false,
+			sourcePlatform: deriveSourcePlatform(r),
```
Bl3f: This might be feasible to be done in the database as well and mainly this map is great but requires to do compute on JS side when it might be pushed backed to the db.

apps/backend/src/services/agent.ts
```diff
@@ -249,6 +249,7 @@ class AgentManager {
 			provider?: Provider;
 			timezone?: string;
 			chatUrl?: string;
+			generateTitle?: boolean;
```
Bl3f: See comment up there, I think we might be able to remove it.

Bl3f (commented):
Clean, a few naming convention remarks and some small changes.

PR #633, opened 2026-04-18
fix(cli): include psycopg binary in postgres extra so driver works without system libpq

### LINKED ISSUE #589: [bug] PostgreSQL regression in v0.1.0 — works on v0.0.56, fails on v0.1.0
## Problem

PostgreSQL connections that work correctly on v0.0.56 fail on v0.1.0. Queries executed via the chat return errors on the newer version.

## Steps to reproduce

1. Connect nao to a PostgreSQL database (on-prem)
2. Run any query via the chat
3. On v0.0.56 (WSL): works fine, returns results
4. On v0.1.0 (WSL or Windows): fails with errors

Screenshots were provided showing working v0.0.56 output vs broken v0.1.0 output.

## Expected behavior

Queries should execute and return results on v0.1.0 the same way they do on v0.0.56.

## Impact

Blocking for an on-prem PoC with a public sector client using PostgreSQL. The user is staying on v0.0.56 as a workaround.

## Notes

- The Fabric SQL execution bug that was also present in v0.1.0 has since been fixed in a later release (confirmed 2026-04-09)
- PostgreSQL still broken as of latest version
- Client runs LLM on OVH or Scaleway (on-prem constraint)

## Reported by

Customer (integrator partner)

Aaryan-549:
Closes #589

  PostgreSQL connections regressed in v0.1.0 because extras split in PR #477 shipped `ibis-framework[postgres]` alone. Latest ibis pulls
  plain `psycopg` (psycopg3), which is pure-Python and requires system `libpq` at runtime — fine on dev machines with `libpq-dev`, broken on
  Windows, WSL, and most on-prem Linux installs. Import fails with:

  > ImportError: no pq wrapper available. libpq library not found

  which surfaces in nao as `MissingDependencyError` from `require_database_backend("postgres")`. Adding `psycopg[binary]` to the `postgres`
  (and `redshift`, which reuses the backend) extra bundles a prebuilt libpq wheel so `pip install 'nao-core[postgres]'` works out of the box.

  Reproduced both before and after: pre-fix `PostgresConfig.check_connection` returns False and `execute_sql` raises; post-fix both succeed
  and return correct rows for SELECT / aggregation / GROUP BY against a Postgres 16 container.

  Changes:
  - `cli/pyproject.toml`: add `psycopg[binary]>=3.2.0` to `postgres` and `redshift` extras
  - `cli/uv.lock`: regenerated

### DIFF
--- a/cli/pyproject.toml
+++ b/cli/pyproject.toml
@@ -47,7 +47,7 @@ dependencies = [
 
 [project.optional-dependencies]
 # Database backends
-postgres = ["ibis-framework[postgres]>=9.0.0"]
+postgres = ["ibis-framework[postgres]>=9.0.0", "psycopg[binary]>=3.2.0"]
 bigquery = ["ibis-framework[bigquery]>=9.0.0", "google-cloud>=0.34.0"]
 snowflake = ["ibis-framework[snowflake]>=9.0.0", "snowflake-connector-python[secure-local-storage]>=4.2.0"]
 duckdb = ["ibis-framework[duckdb]>=9.0.0"]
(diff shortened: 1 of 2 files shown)


Bl3f:
I don't think this is needed.

PR #675, opened 2026-04-26
feat: add assertion-based validation for nao test

pranavworks100:
## What

Adds support for assertion-based validation in `nao test`.

## Features

* Supports assertion types:

  * `contains_metric`
  * `reference_table`
  * `no_hallucination`
  * `match_regex`
* Validates AI response text in addition to existing SQL/data checks
* Maintains backward compatibility for tests without assertions

## Example

```yaml
assertions:
  - type: contains_metric
    value: revenue
  - type: reference_table
    value: fact_sales
  - type: no_hallucination
    value: user_password_hash
```

## Testing

* Verified assertion logic with multiple pass/fail scenarios
* Ensured no regression in existing functionality

### DIFF
--- a/cli/nao_core/commands/test/case.py
+++ b/cli/nao_core/commands/test/case.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from pathlib import Path
+from dataclasses import field
 
 import yaml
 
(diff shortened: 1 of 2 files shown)


pranavworks100:
Thanks for the review — added validation for assertion type/value and handled invalid regex patterns to prevent crashes.

Bl3f:
Most of the checks are 'in' checks but written differently, so it's weird to have different kind of assertions. Not sure it's the best way, we have to think more about it. Closing for now.

PR #677, opened 2026-04-27
re-naming of several function to match a unique standard

socallmebertille:
## Summary

Renames the functions to match a generic route type (get, list, create, update and delete)


Bl3f (approved):
Awesome, LGTM

Issue #697, opened 2026-04-30
[bug] Connecting to Trino fails when a password is configured


dugarsumit:
### Problem
The error is from the Trino connector code. Ibis Trino 12 accepts auth=..., but the repo was passing password=..., which gets forwarded to trino.dbapi.Connection and raises:

```
Connection.__init__() got an unexpected keyword argument 'password'
```

### Expected behavior
A Trino connection with user and password should authenticate successfully. Existing password: config in nao_config.yaml should continue to work without users needing to know the internal Ibis/Trino auth API.

### Steps to reproduce
1. Configure a Trino database connection with a password:

  ```
    databases:
      - type: trino
        name: trino_int
        host: <trino-host>
        port: 8080
        catalog: <catalog>
        user: <username>
        password: <password>
  ```
  2. Run a command that checks or uses the connection, for example:
 
  ```
    nao debug
  ```

### Impact
Users cannot connect to password-protected Trino clusters. This blocks context sync, SQL execution, and any analytics-agent workflows relying on Trino as a data source.

dugarsumit:
Hi @ClaireGz can you please provide some feedback on this issue and the corresponding PR? :)

dugarsumit:
@ClaireGz  can we please open this as I have already also created a PR that fixes the bug? I am also open to any feedback incase the issue/PR is not as per the standards :)

PR #698, opened 2026-04-30
fix: resolve trino connector failure when password is configured

dugarsumit:
### Summary

- Ibis / Trino expects authenticated sessions via the auth argument (e.g. trino.auth.BasicAuthentication), not a legacy-style password kwarg passed through to ibis.trino.connect.
- Passing only password could fail or behave incorrectly depending on Ibis and trino client versions, breaking Trino connections that require HTTP basic auth (sync, profiling, nao debug, etc.).
- Unit tests did not previously assert how TrinoConfig.connect() maps config into ibis.trino.connect, so regressions were easy to miss.

### What changed
`cli/nao_core/config/databases/trino.py`

- Added _build_basic_auth(user, password) (lazy import of BasicAuthentication from trino.auth).
- When password is set, connect() now passes auth=BasicAuthentication(user, password) into ibis.trino.connect instead of password=....
- user (and the rest of the existing kwargs) stay unchanged so passwordless and authenticated flows both keep the correct Trino principal.

`cli/tests/nao_core/config/test_trino.py (new)`

- Mocks ibis.trino.connect and checks:
- No password: user / database / host / port; no auth kwarg.
- With password: user present, auth is a BasicAuthentication, no password key.
- Optional schema_name: schema is forwarded.

### Test plan

- cd cli && .venv/bin/pytest tests/nao_core/config/test_trino.py -v — 3 tests pass (or equivalent with your venv).
- cd cli && make lint — ty + ruff clean (or rely on CI if local dev extras block mysqlclient).
- Manual (optional): nao debug / nao sync against Trino with username + password and confirm connectivity.

This PR was written using Cursor.

### DIFF
--- a/cli/nao_core/config/databases/trino.py
+++ b/cli/nao_core/config/databases/trino.py
@@ -28,6 +28,12 @@ def _is_excluded_schema(value: object) -> bool:
     return not schema or schema in {"none", "null"} or schema in EXCLUDED_SCHEMAS or schema.startswith("pg_")
 
 
+def _build_basic_auth(user: str, password: str):
+    from trino.auth import BasicAuthentication
+
+    return BasicAuthentication(user, password)
+
+
 class TrinoDatabaseContext(DatabaseContext):
     def _array_unnest_join(self, table_sql: str, col_sql: str, alias: str) -> str:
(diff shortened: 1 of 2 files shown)


dugarsumit:
This PR should resolve #697

Issue #648, opened 2026-04-22
[feature] nao MCP


ClaireGz:
## Problem

Customers with existing agents (Claude, Codex) want to plug nao's context and SQL generation capabilities into those pipelines via MCP.

## Expected behavior

nao exposes an MCP server that makes available:
- Context (schemas, metrics, rules, docs) as MCP resources
- tools to search context and execute SQL
- tool to directly send a message to a nao chat (same as SlackBot or TeamsBot or WhatsappBot)

Aaryan-549:
Just had some questions from the team before I start implementing this:
  1. Should the MCP server be its own thing or live in the main backend?
  2. How do I handle project context when external clients hit the MCP? Is it per-request (project_id in each call) or something setup once?
  3. What's the auth here? API keys, tokens, or something else?

PR #708, opened 2026-05-03
first nao MCP version

socallmebertille:
## Summary

- [x] Adds an HTTP MCP server (Streamable + SSE) on the backend to expose nao to external AI clients (Claude, Cursor, Codex…), with session management, TTL, and Better-Auth (Bearer + mcp plugin).

- [x] Exposes 4 families of MCP tools: `ask_nao` (runs the agent and streams progress), `execute_sql` + `build_chart` (data warehouse), `grep` + `ls` (project files), and full story CRUD (list, get, create, update, archive, delete).

- [x] Introduces standalone stories (stories without a parent chat): new userId column on story, new queries, dedicated frontend route, and integration into the stories list.

- [x] Adds an MCP settings page : toggles for enabled / agent mode / tools mode / objects mode, Bearer token display, and a recent call logs table.

### Consent new page to give nao account access to MCP client

image

## Related issue

#648


apps/backend/src/agents/tools/index.ts
```diff
@@ -27,16 +27,24 @@ export const tools = {
...
 	const { execute_python, execute_sandboxed_code, ...baseTools } = tools;
+	const include = (key: string) => !excludeTools?.includes(key);
 
 	return {
-		...baseTools,
-		...mcpTools,
-		...(agentSettings?.experimental?.pythonSandboxing && execute_python && { execute_python }),
-		...(agentSettings?.experimental?.sandboxes && execute_sandboxed_code && { execute_sandboxed_code }),
-		...extraTools,
+		...Object.fromEntries(Object.entries(baseTools).filter(([k]) => include(k))),
```
Bl3f: This pattern of check should be factorised (as it's used 3 times just here)

apps/backend/src/agents/tools/index.ts
```diff
@@ -29,16 +29,24 @@ export const tools = {
...
-export const getTools = (agentSettings: AgentSettings | null, extraTools?: Record<string, unknown>) => {
+export const getTools = (
+	agentSettings: AgentSettings | null,
+	extraTools?: Record<string, unknown>,
+	excludeTools?: readonly string[],
+) => {
 	const mcpTools = mcpService.getMcpTools();
-
 	const { execute_python, execute_sandboxed_code, ...baseTools } = tools;
+	const include = (key: string) => !excludeTools?.includes(key);
```
Bl3f: include is badly named, `isToolIncluded` is potentially better (even if i don't like the `lI` which makes it hard to read)

```suggestion
	const isToolIncluded = (key: string) => !excludeTools?.includes(key);
```

apps/backend/src/db/pg-schema.ts
```diff
@@ -662,3 +677,104 @@ export const log = pgTable(
...
+		calledAt: timestamp('called_at').defaultNow().notNull(),
+	},
+	(t) => [
+		index('mcp_call_log_projectId_idx').on(t.projectId),
+		index('mcp_call_log_userId_idx').on(t.userId),
+		index('mcp_call_log_calledAt_idx').on(t.calledAt),
+	],
+);
+
+export const oauthApplication = pgTable(
```
Bl3f: should these tables be prefixed with mcp in the name? or we dont have control over it as it comes from better auth?

apps/backend/src/auth.ts
```diff
@@ -1,5 +1,7 @@
 import { APIError, betterAuth } from 'better-auth';
 import { drizzleAdapter } from 'better-auth/adapters/drizzle';
+import { mcp as mcpPlugin } from 'better-auth/plugins';
```
Bl3f: its written this plugin will be deprecated here https://better-auth.com/docs/plugins/mcp

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+						chatId: '',
+						agentSettings: null,
+						envVars,
+						azureAccessToken,
+						queryResults: new Map(),
+					},
+				);
+
+				const rows = result.data.slice(0, limit);
+				const queryId = `query_${crypto.randomUUID().slice(0, 8)}`;
```
Bl3f: queryId is already returned by executeQuery

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+export function registerDataTools(server: McpServer, ctx: McpContext): void {
+	server.registerTool(
+		'execute_sql',
+		{
+			title: 'Execute SQL',
+			description:
+				'Run a SQL query against the connected data warehouse. Returns rows as JSON. The response includes a `query_id` — pass it to `build_chart` or reference it in story `<table query_id="...">` blocks. Use ask_nao instead if you want Nao to write the SQL for you.',
+			inputSchema: {
+				sql: z.string().describe('The SQL query to execute'),
+				limit: z
```
Bl3f: as executeQuery does not have this param i'd remove it and eventually says the LLM to add a LIMIT in SQL if it wants to limit, because here it's just a cosmetic limit as the backend received the whole data as the .slice is done afterwards

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+						azureAccessToken,
+						queryResults: new Map(),
+					},
+				);
+
+				const rows = result.data.slice(0, limit);
+				const queryId = `query_${crypto.randomUUID().slice(0, 8)}`;
+				const output = { query_id: queryId, columns: result.columns, row_count: rows.length, data: rows };
+				return {
+					content: [{ type: 'text' as const, text: JSON.stringify(output) }],
```
Bl3f: why do we duplicate the answer? output is present twice, so it means more tokens in the LLM answer? (looks like it's the same for every tool)

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+import { getAzureAccessTokenForUser } from '../../services/microsoft-auth.service';
+import { logger } from '../../utils/logger';
+import type { McpContext } from '../logging';
+import { withLogging } from '../logging';
+
+export function registerDataTools(server: McpServer, ctx: McpContext): void {
+	server.registerTool(
+		'execute_sql',
+		{
+			title: 'Execute SQL',
```
Bl3f: i dont like the fact that we describe twice the signature of the execute_sql, display_chart, ls, grep, it means we need to maintain 2 tools def for these, the best would be to wrap the already existing tools definition for the MCP

apps/backend/src/mcp/tools/agent.ts
```diff
@@ -0,0 +1,137 @@
...
+import { chatUrl } from '../urls';
+
+export function registerAgentTools(server: McpServer, ctx: McpContext): void {
+	server.registerTool(
+		'ask_nao',
+		{
+			description:
+				'Ask nao an analytics question in natural language. Creates a chat that is visible in the UI.\n\n' +
+				'Use this for ad-hoc data exploration and Q&A. ' +
+				'To create a persistent Nao Story (markdown dashboard with embedded charts/tables), do NOT ask nao in natural language — use `execute_sql` → `build_chart` → `create_story` instead. ' +
```
Bl3f: If the story mode or the tool mode is not activated the llm will run in full circle, that's a bit why i don't like the tools exclusion and im more in favour to let the asknao working as the agent would work eventually, but if the orchestrator what to do a story directly it can

apps/backend/src/mcp/tools/agent.ts
```diff
@@ -0,0 +1,137 @@
...
+				'To browse or update existing stories, use `list_stories` / `get_story` / `update_story`.\n\n' +
+				'Returns `chatId` and a `url` that opens the chat in the Nao UI — surface the URL to the user as a clickable link so they can jump to the conversation (and any rendered charts/tables). ' +
```
Bl3f: instead can we type the outputSchema to explain what will be returned?

apps/backend/src/mcp/tools/agent.ts
```diff
@@ -0,0 +1,137 @@
...
+				'The returned `chatId` can also be passed to `create_story` as `chat_id` to attach a follow-up story to this conversation.',
+			inputSchema: {
+				question: z.string().describe('The analytics question'),
+				conversation_id: z
+					.string()
+					.optional()
+					.describe(
+						'Existing chat ID to continue a conversation. Omit to start a new chat. ' +
```
Bl3f: as it's a uuid should we say it?

apps/backend/src/mcp/tools/agent.ts
```diff
@@ -0,0 +1,137 @@
...
+					ctx.projectId,
+					ctx.userId,
+					question,
+					conversation_id,
+				);
+
+				const agent = await agentService.create(chat, undefined, ['story', 'suggest_follow_ups']);
+
+				const progressToken = extra._meta?.progressToken;
+				const result = await agent.streamWithProgress(uiMessages, makeProgressHandler(extra, progressToken));
```
Bl3f: why do we need a streamWithProgress method for this?

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+		}),
+	);
+
+	server.registerTool(
+		'build_chart',
+		{
+			title: 'Build Chart',
+			description:
+				'Generate a Nao-compatible `<chart>` block to embed in story content. Always use this tool instead of writing `<chart>` blocks manually — it ensures the correct syntax for the Nao UI renderer. Workflow: execute_sql → build_chart → create_story/update_story (pass the returned block in `content` and the SQL rows in `query_data`).',
+			inputSchema: {
```
Bl3f: Really annoying that this is duplicated, as it's not a straightforward schema

apps/backend/src/mcp/tools/data.ts
```diff
@@ -0,0 +1,122 @@
...
+					source: 'tool',
+					context: { sql, userId: ctx.userId },
+				});
+				return { content: [{ type: 'text' as const, text: `SQL execution error: ${message}` }], isError: true };
+			}
+		}),
+	);
+
+	server.registerTool(
+		'build_chart',
```
Bl3f: Whats the goal of this tool actually? In this case this tool is just a tool to convert a tool input to a <chart> tag, maybe it's useless and we should "just" teach the llm to write the tag directly?

apps/backend/src/mcp/tools/files.ts
```diff
@@ -0,0 +1,87 @@
...
+import { logger } from '../../utils/logger';
+import type { McpContext } from '../logging';
+import { withLogging } from '../logging';
+
+function makeExecutionOptions(projectFolder: string) {
+	const context = { projectFolder, chatId: '', agentSettings: null, envVars: {}, queryResults: new Map() };
+	return { toolCallId: '', messages: [] as [], experimental_context: context };
+}
+
+export function registerFileTools(server: McpServer, ctx: McpContext): void {
```
Bl3f, line 8: should wrap better the already existing tools I think

apps/backend/src/queries/story.queries.ts
```diff
@@ -14,153 +31,111 @@ export async function getStoryByChatAndSlug(chatId: string, slug: string): Promi
...
+		.from(s.story)
+		.leftJoin(s.chat, eq(s.story.chatId, s.chat.id))
+		.innerJoin(latestVersions, eq(s.story.id, latestVersions.storyId))
+		.innerJoin(
+			s.storyVersion,
+			and(eq(s.storyVersion.storyId, s.story.id), eq(s.storyVersion.version, latestVersions.maxVersion)),
+		)
+		.where(
+			and(
+				eq(s.story.id, storyId),
```
Bl3f, line 78: it means a user with a storyId can read a story he does not have access to I think

apps/backend/src/mcp/tools/stories.ts
```diff
@@ -0,0 +1,418 @@
...
+			title: 'Get Story',
+			description:
+				'Retrieve a full story including its latest content/code. Returns a `url` that opens the rendered story in the Nao UI, and a `chatUrl` that opens the underlying chat conversation (null for standalone stories).',
+			inputSchema: {
+				story_id: z.string().describe('The story ID to retrieve'),
+			},
+		},
+		withLogging('get_story', ctx, async ({ story_id }) => {
+			try {
+				const story = await resolveStory(story_id, ctx);
```
Bl3f, line 61: care i think a user with a story id can read another user story

apps/backend/src/mcp/auth.ts
```diff
@@ -0,0 +1,24 @@
...
+}): Promise<string | null> {
+	const auth = await getAuth();
+	const headers = convertHeaders(fastifyRequest.headers);
+
+	const session = await auth.api.getSession({ headers });
+	if (session?.user) {
+		return session.user.id;
+	}
+
+	const host = fastifyRequest.headers['host'] ?? 'localhost';
```
Bl3f: Not sure about what this does (and the next 5 lines)

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+				server.close().catch(() => {});
+			},
+		});
+
+		await server.connect(transport);
+		await transport.handleRequest(request.raw, reply.raw, request.body as Record<string, unknown>);
+		reply.hijack();
+	});
+
+	app.get('/', async (request, reply) => {
```
Bl3f: put the get as first endpoint

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+		const userId = await resolveUserId(request);
+		if (!userId) {
+			return replyUnauthorized(reply);
+		}
```
Bl3f: This pattern is written 4 times, can it be wrapped in a fastify middleware?

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+
+		const sessionId = request.headers['mcp-session-id'] as string | undefined;
```
Bl3f: is this added by better auth?

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+	});
+
+	app.delete('/', async (request, reply) => {
+		const userId = await resolveUserId(request);
+		if (!userId) {
+			return replyUnauthorized(reply);
+		}
+
+		const sessionId = request.headers['mcp-session-id'] as string | undefined;
+		const session = sessionId ? sessions.get(sessionId) : undefined;
```
Bl3f: should be factorised

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+export const mcpServerRoutes = async (app: App) => {
+	app.post('/', async (request, reply) => {
+		const userId = await resolveUserId(request);
+		if (!userId) {
+			return replyUnauthorized(reply);
+		}
+
+		const existingSessionId = request.headers['mcp-session-id'] as string | undefined;
+		if (existingSessionId) {
+			const session = sessions.get(existingSessionId);
```
Bl3f: why here we check different things for the mcp session?

apps/backend/src/mcp/routes.ts
```diff
@@ -0,0 +1,99 @@
...
+				await session.transport.handleRequest(request.raw, reply.raw, request.body as Record<string, unknown>);
+				reply.hijack();
+				return;
+			}
+			return reply.status(404).send({ error: 'Session not found or expired. Please reinitialize.' });
+		}
+
+		const projectId = await resolveProjectId(userId);
+		const settings = await getMcpEndpointSettings(projectId);
+		if (!settings.enabled) {
```
Bl3f: i think for the get or delete we should also check if enabled (and btw what the get and delete endpoint are for?)

Bl3f (changes requested):
First review of the backend, I think there are a few things to fix still (a lot of duplication)) and I'm afraid also having a lot of execution code that is similar but not factorised to execute things.

Bl3f:
@socallmebertille you have a migration conflict :'(

Bl3f (approved):
Awesme lgtm

PR #709, opened 2026-05-04
fix(databricks): sanitize username in memtable volume name

jeannicolasfine:
## Problem

Users with a dot in their OS username (e.g. `john.doe`) get a `UC_VOLUME_NOT_FOUND` error when connecting to Databricks:

```
[UC_VOLUME_NOT_FOUND] Volume `catalog`.`schema`.`john`.`doe-py=3124-pid=98765` does not exist.
```

Ibis builds the `memtable_volume` name as `{getpass.getuser()}-py={version}-pid={pid}`. When the username contains a dot, Databricks Unity Catalog interprets it as a path separator, producing an invalid 4-part identifier instead of a valid 3-part one.

## Fix

Sanitize the username before constructing the volume name by replacing any character outside `[a-zA-Z0-9_-]` with an underscore, then pass the result explicitly as `memtable_volume` to `ibis.databricks.connect()`.

## Test

Run `nao debug` with a Databricks connection configured — previously failed with `UC_VOLUME_NOT_FOUND`, now connects successfully.


cli/nao_core/config/databases/databricks.py
```diff
@@ -138,6 +138,10 @@ def promptConfig(cls) -> "DatabricksConfig":
 
     def connect(self) -> BaseBackend:
         """Create an Ibis Databricks connection."""
+        import getpass
```
Bl3f: These 3 imports should be imported at the top level of the file.

cli/nao_core/config/databases/databricks.py
```diff
@@ -156,6 +160,13 @@ def connect(self) -> BaseBackend:
         if self.schema_name:
             kwargs["schema"] = self.schema_name
 
+        # Usernames with dots (e.g. "jean-nicolas.fine") break UC volume name parsing:
+        # Databricks treats the dot as a catalog/schema separator, creating a 4-part
+        # path instead of a valid 3-part one. Sanitize before ibis constructs the name.
+        safe_user = re.sub(r"[^a-zA-Z0-9_-]", "_", getpass.getuser())
```
Bl3f: Genuine question why do you need to use getpass.getuser() which gets the name of the machine the server is running on. Which might be weird as we have already a user/password given in the nao_config.yaml. 

I'm not sure this is a good idea to get on-the-fly a username like this?

Bl3f (commented):
Small question

cli/nao_core/config/databases/databricks.py
Bl3f: Ok I just had a look at https://github.com/ibis-project/ibis/blob/e0402155f017fb94bf1129e26773d9ffff2136f5/ibis/backends/databricks/__init__.py#L642 I dont fully understand why Databricks/Ibis needs this but 🤷

jeannicolasfine: Thanks for your feedback @Bl3f 
You're right, there is a better and shorter way to handle it. 
Instead of using the getpass.getuser(), we can use the connection name set in nao_config.yaml. This will avoid relying on the username.

cli/nao_core/config/databases/databricks.py
jeannicolasfine: moved and cleaned to limit the strict necessary ones.

PR #720, opened 2026-05-04
fix(sync): strip .git/ from cloned repos to fix gitlink bug

### LINKED ISSUE #716: [CRITICAL bug] nao sync clones git repos but agent cannot access their content
## Problem

When a git repo is added as a context source in `nao_config.yaml` and `nao sync` runs, the repo is cloned into `repos/<name>/` on disk locally - but anyone else (the agent, a teammate cloning the context repo, GitHub itself) sees an empty folder. Example: https://github.com/getnao/chat-analytics/tree/main/repos shows nothing under `dbt/`, despite `nao sync` having pulled real files locally.

**Same problem when `nao sync` runs in CI.** The CI job clones the parent context repo, runs `nao sync`, and tries to commit/push - the synced `repos/<name>/` content is reduced to a gitlink at push time, so the deployed agent sees an empty folder. CI is the most common path to deployment, so this affects every team using the standard sync-on-push workflow.

## Root cause

`nao sync` runs a plain `git clone` inside `repos/<name>/` (`cli/nao_core/commands/sync/providers/repositories/provider.py:20-73`, lines 28-34 for pull, 52-56 for clone). This leaves a `.git/` directory inside `repos/<name>/`.

From the parent context repo's perspective, that nested `.git/` turns `repos/<name>/` into a **gitlink** (mode `160000`) instead of regular tracked files. Verified on chat-analytics:

```
$ git ls-tree -r HEAD repos/
160000 commit bf48ed0a7b71ab771b44fd4e6faed08b4b9ad3d2    repos/dbt
```

And there is **no `.gitmodules` file** declaring it as a proper submodule.

So when the parent context repo is committed and pushed (locally or from CI):
- Only the commit SHA pointer is stored in git, not the actual file content
- `git submodule update --init` cannot recover the files either - no `.gitmodules` URL is declared
- Anyone cloning the context repo (including the deployed agent) sees `repos/<name>/` as an empty directory

The files only exist on the machine that ran `nao sync`. Everywhere else: empty.

## Expected behavior

After `nao sync`, the files in `repos/<name>/` should be portable: committable to the context repo, pushable to GitHub, and readable by the agent on deploy. Possible fixes:
- Strip `.git/` from the cloned repo after sync (keep files only, treat as a snapshot)
- OR properly register it as a git submodule (write a `.gitmodules` entry with the source URL) so `git submodule update --init` works downstream
- OR clone to a different location and copy file content into a tracked path

## Impact

Critical - the git repo context source feature is silently broken for anyone other than the user who ran `nao sync` locally, including (and especially) when sync runs in CI. Any team or agent relying on the synced repo content gets nothing.

shashank-poola:
Closes #716

## Problem

When `nao sync` clones a git repo into `repos/<name>/`, it leaves a `.git/` folder inside. Git sees this nested `.git/` and treats the whole directory as a **gitlink** (`mode 160000`) instead of regular tracked files.

So when you commit and push the context repo, Git only stores a SHA pointer, not the actual file content. Anyone who clones that repo after you, a teammate, CI pipeline, or the deployed agent, gets an **empty folder, silently, with no errors.**

The files only ever exist on the machine that ran `nao sync`.

```bash
# Before this fix, git was storing a pointer, not files
$ git ls-tree -r HEAD repos/
160000 commit bf48ed0a...    repos/dbt
```

---

### `cli/nao_core/commands/sync/providers/repositories/provider.py`

- After a successful clone, we now call `shutil.rmtree(repo_path / ".git")` to strip the nested Git repo
- `repos/<name>/` becomes a plain folder, `git add` now commits actual file content, not a SHA pointer
- Since `.git/` no longer exists after first sync, `git pull` on re-sync is not possible, changed to always **re-clone fresh** (delete → clone → strip)

### `cli/nao_core/commands/sync/cleanup.py` *(bonus fix)*

- `cleanup_stale_repos` called `base_path.iterdir()` before `repos/` existed on first sync
- This crashed every first-time sync with `[Errno 2] No such file or directory: 'repos'`
- Fixed with a simple early return: `if not base_path.exists(): return`

---

## Proof

```bash
$ uv run nao sync
  Cloning dbt
  ✓ dbt

# .git/ is gone — no more gitlink
$ ls repos/dbt/.git
No such file or directory

# actual files are there and committable
$ ls repos/dbt/
Dockerfile  README.md  dbt_project.yml  models  seeds  profiles.yml ...
```

> Screenshot attached below, shows sync success, `.git/` stripped, and real files present in one run.
Screenshot 2026-05-05 010608

---

## Impact

- Fixes silent data loss for all teams using git repo context sources
- Fixes first-time sync crash when `repos/` doesn't exist yet
- Files in `repos/<name>/` are now portable: committable, pushable, and readable by the deployed agent

### DIFF
--- a/cli/nao_core/commands/sync/cleanup.py
+++ b/cli/nao_core/commands/sync/cleanup.py
@@ -172,6 +172,9 @@ def cleanup_stale_databases(active_databases: List, base_path: Path, verbose: bo
 def cleanup_stale_repos(config_repos: list, base_path: Path, verbose: bool = False) -> None:
     """Remove repositories that are not present in the config file."""
 
+    if not base_path.exists():
+        return
+
     repo_names = {repo.name for repo in config_repos}
     for repo_dir in base_path.iterdir():
         if repo_dir.is_dir() and repo_dir.name not in repo_names:
--- a/cli/nao_core/commands/sync/providers/repositories/provider.py
+++ b/cli/nao_core/commands/sync/providers/repositories/provider.py
@@ -21,18 +21,25 @@ def clone_or_pull_repo(repo: RepoConfig, base_path: Path) -> bool:
     """Clone a repository and strip .git/ so files are tracked as regular content."""
     repo_path = base_path / repo.name
 
+    # Guard against path traversal via malicious repo.name (e.g. "../other")
+    resolved_repo_path = repo_path.resolve()
+    if not resolved_repo_path.is_relative_to(base_path.resolve()):
+        console.print(f"  [yellow]⚠[/yellow] Invalid repo path: {repo.name}")
+        return False
+
+    tmp_path = base_path / f"{repo.name}.tmp"
+
     try:
-        if repo_path.exists():
-            console.print(f"  [dim]Re-cloning[/dim] {repo.name}")
-            shutil.rmtree(repo_path)
-        else:
-            console.print(f"  [dim]Cloning[/dim] {repo.name}")
+        console.print(f"  [dim]{'Re-cloning' if repo_path.exists() else 'Cloning'}[/dim] {repo.name}")
+
+        if tmp_path.exists():
+            shutil.rmtree(tmp_path)
 
         cmd = ["git", "clone"]
         if repo.branch:
             cmd.extend(["-b", repo.branch])
         if repo.url:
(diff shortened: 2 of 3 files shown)


Bl3f (approved):
LGTM

PR #745, opened 2026-05-07
fix: support all LLM providers in NLP cron and live story refresh

### LINKED ISSUE #643: getProjectModelProvider only defaults on Anthropic or OpenAI for nlp and story live refresh
in `apps/backend/src/queries/project-llm-config.queries.ts` getProjectModelProvider default only on OpenAI or Anthropic, but it's used only for cron fallback in live Story and live Story refresh.

This method should probably removed in favour of other method that are returning the correct env setup or data from the db.

```
export const getProjectModelProvider = async (projectId: string): Promise<LlmProvider | undefined> => {
	const configs = await getProjectLlmConfigs(projectId);

	// Return first configured provider, preferring anthropic
	const anthropicConfig = configs.find((c) => c.provider === 'anthropic');
	if (anthropicConfig) {
		return 'anthropic';
	}

	const openaiConfig = configs.find((c) => c.provider === 'openai');
	if (openaiConfig) {
		return 'openai';
	}

	// Fall back to env providers
	return getDefaultEnvProvider();
};
```

shashank-poola:
## Problem

`getDefaultEnvProvider` hardcoded only `ANTHROPIC_API_KEY` → `OPENAI_API_KEY`, silently returning `null` for 7 of 9 providers (google, mistral, openrouter, ollama, bedrock, vertex, azure).

This broke two features for non-Anthropic/OpenAI users:
- NLP cron parsing (`story.parseCronFromText`)
- Live story dynamic text refresh

No error shown, features just silently did nothing.

## Fix

Deleted `getProjectModelProvider` and replaced with `resolveFirstProjectModel`:

```ts
// Before: hardcoded
if (process.env.ANTHROPIC_API_KEY) return 'anthropic';
if (process.env.OPENAI_API_KEY) return 'openai';
return undefined; // all other providers silently fail

// After: works for all 9 providers
export async function resolveFirstProjectModel(
  projectId: string,
  selectModelId: (provider: LlmProvider) => string,
): Promise<ProviderModelResult | null>
Tries DB-configured providers first, falls back to env, skips providers with no model ID (azure), validates credentials before returning.
```

Evidence
Tested with only GEMINI_API_KEY set (no Anthropic/OpenAI key):
image

Backend log confirming Gemini API was called (responseTime: 2719ms proves real API call, not null return):
image

Tests
8 unit tests added - apps/backend/tests/resolve-first-project-model.test.ts

Screenshot 2026-05-08 013651
Tests  8 passed (8)

Closes #643

### DIFF
--- a/apps/backend/src/queries/project-llm-config.queries.ts
+++ b/apps/backend/src/queries/project-llm-config.queries.ts
@@ -3,7 +3,7 @@ import { and, eq } from 'drizzle-orm';
 
 import s, { DBProjectLlmConfig, NewProjectLlmConfig } from '../db/abstractSchema';
 import { db } from '../db/db';
-import { getDefaultEnvProvider, getDefaultModelId } from '../utils/llm';
+import { getDefaultModelId } from '../utils/llm';
 
 export const getProjectLlmConfigs = async (projectId: string): Promise<DBProjectLlmConfig[]> => {
 	return db.select().from(s.projectLlmConfig).where(eq(s.projectLlmConfig.projectId, projectId)).execute();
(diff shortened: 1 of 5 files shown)


shashank-poola:
@Bl3f - this PR got auto-closed by the contributor gate but the fix is ready. Handles all 9 LLM providers, 8 unit tests passing, real Gemini API call as evidence. Would love a review when you have a moment.

PR #776, opened 2026-05-11
added rerun feature for execute_sql tool call

### LINKED ISSUE #764: [Feature] Add rerunnable tool calls for repeatable analyses
Allow users to re-run successful tool calls directly from the chat UI, starting with `execute_sql`.

For example, after nao runs a SQL query, the SQL tool card could show actions like:

- Re-run
- Copy SQL
- Save as analysis

Clicking **Re-run** would execute the original tool input again, without calling the LLM. The result would be shown as a fresh tool result in the chat.

## Why This Is Useful

- Makes correct analyses repeatable.
- Lets users refresh results against newer data.
- Reduces unnecessary LLM calls and latency.
- Improves trust because the exact same SQL/input is reused.
- Turns good chat outputs into reusable analytics workflows over time.

devxspectre:
fixes #764 

## Summary

Adds support for rerunning successful `execute_sql` tool calls from the chat UI.

- Adds a rerun action to SQL tool call cards, with pending/disabled state handling.
- Adds a backend `chat.rerunExecuteSqlToolCall` mutation that validates ownership, role, tool type, and successful tool state before re-executing the stored SQL input.
- Appends the fresh SQL result as a new assistant tool message and refreshes the chat/list queries in the frontend.
- Adds backend tests for successful reruns and viewer permission rejection.

## Testing

- Added `apps/backend/tests/chat-rerun-execute-sql.test.ts`

### DIFF
--- a/apps/backend/src/queries/chat.queries.ts
+++ b/apps/backend/src/queries/chat.queries.ts
@@ -17,7 +17,7 @@ import s, {
 } from '../db/abstractSchema';
 import { db } from '../db/db';
 import dbConfig, { Dialect } from '../db/dbConfig';
-import { ForkMetadata, StopReason, TokenUsage, UIChat, UIMessage, UIMessagePart } from '../types/chat';
+import { ForkMetadata, StopReason, TokenUsage, ToolState, UIChat, UIMessage, UIMessagePart } from '../types/chat';
 import { applyChatFilters, buildChatGroups, type EnrichedChat, type SourcePlatform } from '../utils/chat-list';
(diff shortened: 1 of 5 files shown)


Bl3f:
I dont think we will merge this PR this way as the re-run re-add a new entry in the chat which is not really friendly. We will pbly do this internally :)

devxspectre:
Hey @Bl3f , do you guys have some internal slack or something. I saw this issue https://github.com/getnao/nao/issues/779 and would love to work with you guys on it.

PR #788, opened 2026-05-14
feat(config): resolve AWS and Kubernetes secrets in nao_config.yaml

afranzi:
## Summary

- Extends the existing `{{ env('VAR') }}` template substitution in `nao_config.yaml` with two new resolvers: `{{ aws('secret_name/field') }}` and `{{ k8s('namespace/secret/field') }}`, so deployments can keep credentials in AWS Secrets Manager or Kubernetes Secrets instead of plain env vars.
- Lazy imports via `require_dependency`; new optional extras `nao-core[aws-secrets]` and `nao-core[k8s-secrets]` keep the install surface lean.
- Auto-generated configuration reference updated to document the new syntax.

## Syntax

```yaml
databases:
  - type: postgres
    user:     ${{ env('DB_USER') }}                    # unchanged
    password: ${{ aws('prod/db/credentials/password') }}
    api_key:  '{{ aws("arn:aws:secretsmanager:eu-west-1:123:secret:foo-AbCdEf/key") }}'

slack:
  bot_token: '{{ k8s("default/slack-secrets/bot_token") }}'   # namespace explicit
  signing_secret: '{{ k8s("slack-secrets/signing_secret") }}'  # falls back to pod namespace
```

## Behaviour

| Resolver | On failure |
|---|---|
| `env`   | empty string + warning at validation failure (unchanged) |
| `aws`   | raises `ValueError`, surfaced with a clear message |
| `k8s`   | raises `ValueError`, surfaced with a clear message |

AWS payloads are parsed as JSON and the trailing `/field` selects the key; region is auto-detected from the ARN. Kubernetes namespace defaults to the pod namespace when omitted.

## Implementation notes

- `_process_env_vars` delegates to the new `process_secrets` helper. All 20 existing env-resolution tests pass untouched, so this is a strict superset of the previous behaviour.
- AWS/K8s SDKs are imported inside the resolver functions, guarded by `require_dependency` — users who don't use those backends never need to install boto3 or kubernetes.

## Test plan

- [x] 29 new unit tests in `cli/tests/nao_core/config/test_secrets.py` covering env/aws/k8s resolution, identifier parsing, ARN region detection, pod-namespace fallback, and dispatch via `process_secrets`
- [x] All 20 existing `test_base.py` env-resolution tests still pass
- [x] `ruff check`, `ruff format --check`, `ruff check --select I` clean on all changed files
- [ ] Smoke test against a real AWS Secrets Manager secret
- [ ] Smoke test against a real Kubernetes Secret in-cluster


scripts/generate-config-docs.py
```diff
@@ -328,8 +328,20 @@ def generate_markdown() -> str:
         "[`cli/nao_core/config/`](https://github.com/getnao/nao/tree/main/cli/nao_core/config).\n"
     )
     parts.append(
```
Bl3f, line 330: You don't need to add this, we manage the docs evolution a bit differently for the moment (it's in another repo not public).

Bl3f (commented):
I think this is good. In addition to this we will work internally to also support these kind of secrets on the backend side (so in Typescript, but it requires more infra stuff to do and I have it ready on another branch so don't do it).

I will just ask you to resolve the conflict with the uv.lock and then it will be good.

afranzi:
@Bl3f comments addressed :) Thank you!

Bl3f (approved):
Awesome ty!

Bl3f:
Looks like pytest are failing because they require glom to be installed, can you change the cli-lint so it install the extra you added for the test to run

Bl3f:
kubernetes extra should be added as well

afranzi:
It should be ready :) thank you @Bl3f !

Bl3f:
Thank you @afranzi! And welcome for this first contribution

PR #797, opened 2026-05-18
feat: make optional 'Max output tokens per model'

ido177:
Added configurable max_output_tokens per provider/model in Project Settings, while keeping 16,000 as the default fallback.

This change fixes provider-side failures where a global hardcoded value exceeded model limits (for example, models capped at 4096 total tokens), causing requests to be rejected before generation. Making the limit model-specific keeps current behavior by default, but allows safe overrides for constrained models/endpoints and prevents runtime token-limit errors.

```
[ERROR] [agent] Agent stream error: 
Error: max_output_tokens=16000 cannot be greater than 
max_model_len=max_total_tokens=4096. Please request fewer output tokens. (parameter=max_output_tokens, 
value=16000)
```

Screenshot 2026-05-18 at 2 13 10 PM

### DIFF
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -195,6 +195,14 @@ export const updateAgentSettings = async (projectId: string, settings: AgentSett
 	const next: AgentSettings = {
 		...current,
 		...settings,
+		llm: {
+			...current.llm,
+			...settings.llm,
+			maxOutputTokensByProviderModel: {
+				...current.llm?.maxOutputTokensByProviderModel,
+				...settings.llm?.maxOutputTokensByProviderModel,
+			},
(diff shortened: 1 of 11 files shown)


Bl3f:
For which model this error was encountered? Because I'm not fond of this because it clutters the UI and I'd keep for the moment the max value in provider-meta.ts

ido177:
I encountered this issue while integrating our custom model. After making the limit configurable, the model started working correctly

The main issue is that for some models, the token limit cannot be adjusted dynamically on the provider side without affecting other services that rely on the same configuration. In such cases, keeping a hardcoded global value causes requests to fail before generation starts

From what I can see, this is currently the only practical way to support models with lower token limits without introducing runtime token limit errors

PR #800, opened 2026-05-19
Fix timestamp type for period calculation of providers budget  in Postgres

socallmebertille:
## Summary

- Issue : Postgres returned an error when calculating the budget overrun for budgets allocated to model providers because it did not receive the value with the correct data type
- Solution : Make the CASE expression return a real timestamp on Postgres by emitting the cast inline


Bl3f:
Automation triage check: reviewed for merge readiness on behalf of christophe@getnao.io.

PR #802, opened 2026-05-19
fix: disable clarification tool during test runs

### LINKED ISSUE #792: Agent should not ask for clarification during tests
When using the `nao test` framework I'm getting some failed tests because the agent asks for clarification.

The agent should know that in test runs the user cannot give input so it should either make assumptions and give its best attempt or it should reply that it cannot answer.

ClaireGz:
## Summary
- Adds a `testMode` flag to `AgentChat`, set automatically in `TestAgentService.runTest()`
- Removes the `clarification` tool from the agent's toolset during test runs so the agent cannot call it
- Removes `clarification` from the `stopWhen` conditions in test mode
- Adds a "Test Mode" section to the system prompt instructing the agent to make assumptions or state it cannot answer instead of asking for clarification

## Test plan
- [ ] Run `nao test` with prompts that previously triggered clarification — agent should now attempt an answer
- [ ] Run `nao test` with unambiguous prompts — no behavior change expected
- [ ] Verify interactive chat still has the clarification tool available and working

Closes #792

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/apps/backend/src/agents/tools/index.ts
+++ b/apps/backend/src/agents/tools/index.ts
@@ -31,13 +31,18 @@ export const tools = {
 	suggest_follow_ups: suggestFollowUps,
 };
 
-export const getTools = (agentSettings: AgentSettings | null, extraTools?: Record<string, unknown>) => {
+export const getTools = (
+	agentSettings: AgentSettings | null,
+	extraTools?: Record<string, unknown>,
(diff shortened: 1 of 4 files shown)


Bl3f (commented):
LGTM

PR #824, opened 2026-05-21
fix(skills): use nao init --no-tty in setup-context skill

ClaireGz:
## Summary
- Update the `setup-context` skill to use `nao init --no-tty` instead of running `nao init` interactively
- Matches the non-interactive mode added in #809 — the flag skips all prompts, scaffolds folders, and auto-installs deps when a pre-written `nao_config.yaml` exists
- Strengthens the guardrail to explicitly forbid running `nao init` without the flag

## Test plan
- [ ] Verify the `setup-context` skill runs `nao init --no-tty` without hanging on prompts
- [ ] Confirm folder scaffolding and dependency install still work end-to-end

### DIFF
--- a/skills/setup-context/SKILL.md
+++ b/skills/setup-context/SKILL.md
@@ -28,12 +28,12 @@ Send a single message asking for:
 
 2. **Write `nao_config.yaml`** from the answers (skeleton in appendix below).
 
-3. **Run `nao init`** — it detects the existing yaml and offers to update; confirm. Folder scaffold gets created. Say "no" to optional providers (skills / MCPs / Notion / Slack); edit the yaml directly afterwards if needed.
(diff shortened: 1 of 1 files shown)


skills/setup-context/SKILL.md
```diff
@@ -28,12 +28,12 @@ Send a single message asking for:
 
 2. **Write `nao_config.yaml`** from the answers (skeleton in appendix below).
 
-3. **Run `nao init`** — it detects the existing yaml and offers to update; confirm. Folder scaffold gets created. Say "no" to optional providers (skills / MCPs / Notion / Slack); edit the yaml directly afterwards if needed.
+3. **Run `nao init --no-tty`** — it detects the pre-written yaml, skips all prompts, scaffolds the folder structure, and auto-installs missing dependencies. No interactive confirmation needed.
```
Bl3f, line 31: should add the --name param as well (that has been just added)

Bl3f:
up

Bl3f (approved):
LGTM

PR #855, opened 2026-05-27
fix(cli): forward base_url to Anthropic client in template engine

d-axel-b:
The Anthropic provider in _generate_anthropic() was ignoring the base_url from nao_config.yaml, preventing users behind custom proxies from using the Anthropic provider in the CLI.

Necessary to run nao sync with ai-summary


Bl3f:
Ty @d-axel-b

PR #902, opened 2026-06-10
Add tool call density (compact/detailed) + user pref table

Bl3f:
Screenshot 2026-06-10 at 02 04 52

Added a tool call density setting to compact or not SQL / sandbox tools.

Screenshot 2026-06-10 at 02 05 26

Screenshot 2026-06-10 at 02 05 39

---

> [!NOTE]
> **Low Risk**
> Additive schema and user-scoped preference APIs; no changes to auth or billing paths beyond normal protected user mutations.
> 
> **Overview**
> Introduces **per-user preference storage** so settings can follow the account instead of living only in the client. Migration **0048** adds a `user_preference` row per user (JSON `preferences`, timestamps, cascade delete on `user`) for both Postgres and SQLite, with Drizzle schema and journal updates.
> 
> That store backs a new **tool call density** preference (**compact** vs **detailed**). Users can change it in settings; the backend exposes read/update via authenticated user APIs, and chat rendering uses the value when grouping tool steps (e.g. SQL / sandbox calls) so compact mode shows less UI noise.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 00ec1de8318dcb66d40511b5c1f45e2a669e1b91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


apps/backend/src/trpc/user.routes.ts
```diff
@@ -76,6 +78,20 @@ export const userRoutes = {
 			});
 		}),
 
+	getPreferences: protectedProcedure.query(async ({ ctx }) => {
```
socallmebertille, line 82: I think that it should at least be `projectProtectedProcedure` with a `user.id` check or create a procedure that does both so it can be re-used even on the route below

socallmebertille (approved):
LGTM 👍

apps/backend/src/trpc/user.routes.ts
Bl3f: I think it can stay protectedProcedure, because the userPref is not related to project i'd say

PR #915, opened 2026-06-15
fix(backend): serialize list_stories timestamps to ISO strings

### LINKED ISSUE #914: list_stories MCP tool returns Date objects and fails output validation
Calling the `list_stories` MCP tool errors out instead of returning stories. In `apps/backend/src/mcp/tools/asset-tools.ts` the handler builds `structuredContent` with `createdAt`/`updatedAt` taken straight from the Drizzle row — those are `Date` objects — but the tool's `outputSchema` declares both as `z.string()`. The SDK validates `structuredContent` against the schema, so the `Date` fails `z.string()` and the whole call fails.

```
MCP error -32602: Output validation error: Invalid structured content for tool list_stories: [
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "stories",
      0,
      "createdAt"
    ],
    "message": "Invalid input: expected string, received Date"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "stories",
      0,
      "updatedAt"
    ],
    "message": "Invalid input: expected string, received Date"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "stories",
      1,
      "createdAt"
    ],
    "message": "Invalid input: expected string, received Date"
  },
  {
    "expected": "string",
    "code": "invalid_type",
    "path": [
      "stories",
      1,
      "updatedAt"
    ],
    "message": "Invalid input: expected string, received Date"
  },

  ...

]
```

The `text` content is fine because `JSON.stringify` turns the Dates into ISO strings — only the structured path breaks. It hits both SQLite and Postgres, since both map these columns to `Date`. I ran into it with a handful of stories in a project, and it makes the tool unusable for reading story history.

It's isolated to `list_stories`. `get_story`/`create_story`/`update_story` build their `structuredContent` from a slim, string-only payload, so their date fields never reach validation.

Fix I have in mind: serialize the two timestamps with `.toISOString()` in the handler (both columns are `NOT NULL`), via a small typed mapping so the Date/string mismatch becomes a compile error, plus a unit test that validates the mapped row against the output schema. I'd like to implement this myself.

manypozo:
## What this does

`list_stories` was returning Drizzle `Date` objects for `createdAt`/`updatedAt` in its MCP `structuredContent`, while the tool's output schema declares them as `z.string()`. The SDK validates structured output against that schema, so the call failed validation and the tool was unusable for listing stories.

This serializes both timestamps to ISO strings. I pulled the row->item mapping out of the handler into a small typed `toStoryListItem()` (in `story-list-item.ts`) that returns the schema's inferred type, so a `Date`/`string` mismatch is now a compile error. The mapper takes the two URLs as arguments so it stays free of the `bun:sqlite` DB chain and is unit-testable under vitest.

The `text` content is unchanged — `JSON.stringify` already rendered the Dates as ISO strings; only the `structuredContent` path was broken. It's isolated to `list_stories`; the other story tools build structured content from a slim, string-only payload. Both timestamp columns are `NOT NULL` in the SQLite and Postgres schemas, so `.toISOString()` is always safe.

## How I tested

- Added a vitest unit test that maps a row with `Date` timestamps and asserts the result parses against the tool's actual output schema (the same Zod object the SDK validates). Watched it fail on the `Date` first, then pass after the fix.
- `tsc --noEmit` + eslint clean; the full pre-commit hook (lint/format/migrations/cli) is green.
- I didn't run a live MCP round-trip — there's no MCP integration harness in the repo and no warehouse/LLM-backed instance here — but the unit test exercises the exact schema validation that was failing.

Fixes #914

---

> [!NOTE]
> **Low Risk**
> Scoped to `list_stories` response shaping and shared list-item types; no auth, persistence, or other story tool behavior changes.
> 
> **Overview**
> Fixes **`list_stories`** MCP failures where SDK validation rejected `structuredContent` because `createdAt`/`updatedAt` were Drizzle **`Date`** values while the output schema expects **`z.string()`** ISO timestamps. The text JSON path was already fine via `JSON.stringify`; only structured output was broken.
> 
> Introduces **`story-list-item.ts`** with shared **`STORY_LIST_ITEM_SCHEMA`**, **`toStoryListItem()`** (maps rows and calls **`.toISOString()`** on both timestamps), and a typed return so date/string mismatches fail at compile time. **`list_stories`** in `asset-tools.ts` now uses that helper instead of inline mapping.
> 
> Adds a vitest test that asserts ISO serialization and that the mapped item parses against the real output schema.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 34721ccd0f4527e62f62a2d940cf70c0bbced80e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

<a href="https://cubic.dev/pr/getnao/nao/pull/915?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/mcp/tools/story-list-item.ts
+++ b/apps/backend/src/mcp/tools/story-list-item.ts
@@ -1,0 +1,27 @@
+import { z } from 'zod';
+
+import type { UserStoryRow } from '../../queries/story.queries';
+
+export const STORY_LIST_ITEM_SCHEMA = z.object({
+	id: z.string().describe('Story UUID.'),
+	title: z.string().describe('Story title.'),
+	url: z.url().describe('URL to open the story in the nao UI.'),
+	chatUrl: z.url().nullable().describe('Source chat URL, or null for standalone stories.'),
+	archived: z.boolean().describe('True if soft-deleted via `archive_story` (still recoverable).'),
+	createdAt: z.string().describe('ISO timestamp of creation.'),
+	updatedAt: z.string().describe('ISO timestamp of last edit.'),
+});
+
+export type StoryListItem = z.infer<typeof STORY_LIST_ITEM_SCHEMA>;
+
+export function toStoryListItem(story: UserStoryRow, urls: { url: string; chatUrl: string | null }): StoryListItem {
+	return {
+		id: story.id,
+		title: story.title,
+		url: urls.url,
+		chatUrl: urls.chatUrl,
+		archived: story.archivedAt !== null,
+		createdAt: story.createdAt.toISOString(),
+		updatedAt: story.updatedAt.toISOString(),
+	};
+}
--- a/apps/backend/tests/story-list-item.test.ts
+++ b/apps/backend/tests/story-list-item.test.ts
@@ -34,4 +34,15 @@ describe('list_stories output mapping', () => {
 		expect(item.updatedAt).toBe('2024-01-02T03:04:05.000Z');
 		expect(() => STORY_LIST_ITEM_SCHEMA.parse(item)).not.toThrow();
 	});
+
+	it('marks archived stories and preserves chatUrl', () => {
+		const item = toStoryListItem(buildStoryRow({ archivedAt: new Date('2024-02-01T00:00:00.000Z') }), {
+			url: 'http://localhost:5005/stories/standalone/story-1',
+			chatUrl: 'http://localhost:5005/chats/chat-1',
+		});
+
+		expect(item.archived).toBe(true);
+		expect(item.chatUrl).toBe('http://localhost:5005/chats/chat-1');
+		expect(() => STORY_LIST_ITEM_SCHEMA.parse(item)).not.toThrow();
+	});
 });
--- a/apps/backend/tests/story-list-item.test.ts
+++ b/apps/backend/tests/story-list-item.test.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from 'vitest';
+
+import { STORY_LIST_ITEM_SCHEMA, toStoryListItem } from '../src/mcp/tools/story-list-item';
+import type { UserStoryRow } from '../src/queries/story.queries';
+
+function buildStoryRow(overrides: Partial<UserStoryRow> = {}): UserStoryRow {
+	return {
+		id: 'story-1',
+		chatId: null,
+		projectId: 'project-1',
+		userId: 'user-1',
+		slug: 'revenue-dashboard',
+		title: 'Revenue Dashboard',
+		isLive: false,
+		isLiveTextDynamic: true,
+		cacheSchedule: null,
+		cacheScheduleDescription: null,
+		archivedAt: null,
+		createdAt: new Date('2024-01-01T00:00:00.000Z'),
+		updatedAt: new Date('2024-01-02T03:04:05.000Z'),
+		code: '# Revenue\n',
+		...overrides,
+	};
+}
+
+describe('list_stories output mapping', () => {
+	it('serializes createdAt/updatedAt as ISO strings that satisfy the output schema', () => {
+		const item = toStoryListItem(buildStoryRow(), {
+			url: 'http://localhost:5005/stories/standalone/story-1',
+			chatUrl: null,
+		});
+
+		expect(item.createdAt).toBe('2024-01-01T00:00:00.000Z');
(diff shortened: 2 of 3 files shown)


apps/backend/src/mcp/tools/story-list-item.ts
```diff
@@ -0,0 +1,27 @@
...
+	return {
+		id: story.id,
+		title: story.title,
+		url: urls.url,
+		chatUrl: urls.chatUrl,
+		archived: story.archivedAt !== null,
+		createdAt: story.createdAt.toISOString(),
+		updatedAt: story.updatedAt.toISOString(),
+	};
+}
```
socallmebertille, line 27: I think you could have left them in `asset-tools.ts` instead of creating a new file

apps/backend/tests/story-list-item.test.ts
```diff
@@ -0,0 +1,37 @@
...
+	it('serializes createdAt/updatedAt as ISO strings that satisfy the output schema', () => {
+		const item = toStoryListItem(buildStoryRow(), {
+			url: 'http://localhost:5005/stories/standalone/story-1',
+			chatUrl: null,
+		});
+
+		expect(item.createdAt).toBe('2024-01-01T00:00:00.000Z');
+		expect(item.updatedAt).toBe('2024-01-02T03:04:05.000Z');
+		expect(() => STORY_LIST_ITEM_SCHEMA.parse(item)).not.toThrow();
+	});
```
socallmebertille, line 36: It's a bit thin though — only the happy path is covered. Could you add a case for `archived: true` (the `archivedAt !== null` branch is never exercised) and a non-null `chatUrl` ?

```suggestion
	});
	it('marks archived stories and preserves chatUrl', () => {
	  const item = toStoryListItem(
	    buildStoryRow({ archivedAt: new Date('2024-02-01T00:00:00.000Z') }),
	    { url: 'http://localhost:5005/stories/standalone/story-1', chatUrl: 'http://localhost:5005/chats/chat-1' },
	  );
	  expect(item.archived).toBe(true);
	  expect(item.chatUrl).toBe('http://localhost:5005/chats/chat-1');
	  expect(() => STORY_LIST_ITEM_SCHEMA.parse(item)).not.toThrow();
	});
```

socallmebertille (changes requested):
Thanks a lot for your contribution, it's a great PR ! As soon as you'v done the changes, I think we'll be able to merge it ! 👍

apps/backend/src/mcp/tools/story-list-item.ts
manypozo: I actually tried. There is an internal dependency with vitest that did not make it simple to keep it there.

manypozo: The split is deliberate: the backend vitest suite runs under Node, but `asset-tools.ts` transitively imports `src/db/db.ts`, which does `import { Database } from 'bun:sqlite'` (via `story.queries` and `urls`). Importing `asset-tools.ts` from a test fails with `Cannot find package 'bun:sqlite'`.

Keeping `toStoryListItem` / `STORY_LIST_ITEM_SCHEMA` in their own db-free module is what makes the mapper unit-testable under vitest — including the archived/`chatUrl` cases you asked for below. Happy to relocate it if you'd prefer a different home, but it needs to stay clear of the db import chain. Let me know what you'd like.

apps/backend/tests/story-list-item.test.ts
manypozo: Added in 65048bb — a second case maps a row with `archivedAt` set and a non-null `chatUrl`, asserting `archived === true`, the `chatUrl` passthrough, and that the item parses against `STORY_LIST_ITEM_SCHEMA`.

Bl3f (commented):
Awesome, looks like we can merge?

manypozo:
ok for me, waiting for @socallmebertille

socallmebertille (approved):
Thanks for your answer, the PR is really great !

PR #928, opened 2026-06-18
feat(mcp): per-user OAuth for remote MCP servers

dennisUhrskov:
Issue: #929

Adds per-user OAuth so a deployed, multi-user nao can connect to OAuth-only remote MCP servers — concretely Mixpanel's hosted MCP (`https://mcp-eu.mixpanel.com/mcp`), but the design is generic. Today nao only supports static `headers` / stdio `env` for remote servers; for OAuth-only servers `McpService` hands off to mcporter's built-in flow, which opens the system browser and caches a single token under `~/.mcporter/`. That works for local `nao chat` but breaks in a deployed pod (no browser, and one shared vault means no per-user identity — yet these servers scope permissions to the logged-in user).

For servers that opt in via a new `oauth` config block, nao bypasses mcporter and drives the `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport` with a nao-owned, DB-backed `OAuthClientProvider` and a web callback. stdio and static-header HTTP servers are untouched.

## What's included (commit by commit)

1. **Config schema** — optional `oauth` block (`clientId?`, `clientSecretEnv?`, `scopes?`, `dynamicRegistration?`) in `apps/shared/src/mcp.ts`, mirrored in the CLI (`cli/nao_core/config/mcp/`) with a Mixpanel template. Fully backward-compatible.
2. **Token store (dual-DB)** — `mcp_oauth_token` table keyed by `(userId, projectId, serverName)` in both the Postgres and SQLite schemas, with `0049` migrations in both dirs, plus `queries/mcp-oauth.queries.ts`.
3. **`OAuthClientProvider`** — `services/mcp-oauth-provider.ts` implementing the SDK interface against the token store, scoped to one `(userId, projectId, server)` session. Durable tokens + dynamically-registered client info in the DB; the transient PKCE verifier and CSRF state in memory. Supports dynamic client registration (RFC 7591) with a fallback to a configured `clientId`.
4. **Web connect + callback routes** — `routes/mcp-oauth.ts` (`/api/mcp-oauth/connect/:projectId/:serverName` and `/api/mcp-oauth/callback`) + `services/mcp-oauth.service.ts`, backed by a durable `mcp_oauth_flow` table (`0050` migrations) holding the PKCE verifier/state across the redirect boundary. Membership-gated; open-redirect guard on `returnTo`.
5. **Per-user execution** — `McpService` diverts `oauth` servers off mcporter and opens a cached per-`(user, server)` SDK connection authenticated with the user's own tokens (`services/mcp-user-connection.ts`). Tool schemas fold into the shared per-project state (enable/disable keeps working); execution runs under the calling user's token. `userId` is threaded `ToolContext → getTools → getMcpTools`.
6. **Settings UI + docs** — per-server Connect/Disconnect + connection status in Settings → MCP Servers (any member; tool enable/disable stays admin-only), `oauthStatus`/`disconnectOAuth` tRPC, search-index entry, and `apps/backend/docs/mcp-oauth.md`.

## Design decisions for review

- **OSS, not EE-gated.** This is an outbound connector (nao → an external tool's MCP), like `routes/slack.ts` / `teams.ts` / `telegram.ts`, not a nao-login identity provider. Kept modular so it can be wrapped in `hasFeature(...)` later if you disagree (a one-line guard).
- **Bypass mcporter** rather than upstreaming an injected-provider feature request — mcporter 0.7.3 exposes no hook to inject a custom `OAuthClientProvider`. Open to upstreaming instead if you'd prefer.
- **Durable pending-flow table** (`mcp_oauth_flow`) for the PKCE handshake rather than an in-memory map, so it survives restarts / multiple pods.
- **Per-user schema discovery.** A server's tools appear after that user connects, rather than being pre-discovered from a shared service token.
- **Deferred:** the in-conversation Connect card (surfacing `UnauthorizedError` inline in chat). The settings flow is the entry point; the card is a follow-up.

## Testing

- Unit tests: provider, connect/callback service, per-user connection manager, and `McpService` routing (proving stdio/static-header servers are unaffected) — 47 tests.
- **Live, against the real Mixpanel hosted MCP** (behind a public HTTPS tunnel):
  - Dynamic client registration works with no configured `clientId`.
  - Mixpanel accepts the web callback URL; the SDK exchanged the code and persisted tokens.
  - With empty `scopes`, the full tool list is returned and tools appear in Settings → MCP Servers immediately after connecting.
  - Connect → Connected status → tools listed → Disconnect all work from the settings UI.

Most of the diff is generated Drizzle migration snapshots (`0049`/`0050`).

### DIFF
--- a/apps/backend/docs/mcp-oauth.md
+++ b/apps/backend/docs/mcp-oauth.md
@@ -1,0 +1,62 @@
+# OAuth for remote MCP servers
+
+nao can connect to OAuth-protected remote MCP servers (for example Mixpanel's hosted MCP at `https://mcp-eu.mixpanel.com/mcp`) on a **per-user** basis. Each user authorizes with their own credentials, so the remote server's existing permissions and roles apply to that user's tool calls.
+
+This is distinct from the static `headers` / stdio `env` mechanism, which shares a single credential across everyone. It is also distinct from OIDC/SSO, which governs how users _log in to nao_ — this feature is an **outbound connector** (nao → an external tool's MCP) and ships in the open-source edition (no license required).
+
+## How it works
+
+A remote HTTP server opts into the per-user OAuth flow by adding an `oauth` block to its entry in `agent/mcps/mcp.json`. Servers without an `oauth` block are unchanged: stdio and static-header HTTP servers keep using the bundled mcporter runtime.
+
+For an OAuth server, nao bypasses mcporter and drives the `@modelcontextprotocol/sdk` `StreamableHTTPClientTransport` directly with a nao-owned, database-backed `OAuthClientProvider`:
+
+1. A user starts the flow from **Settings → MCP Servers** (the "Connect" button next to the server).
+2. nao runs OAuth 2.1 discovery and, if the server supports it, **dynamic client registration** (RFC 7591). Otherwise it uses a pre-registered `clientId`.
+3. The browser is redirected to the server's authorization page. The PKCE `code_verifier` and a CSRF `state` are persisted server-side for the duration of the flow.
+4. The server redirects back to nao's callback, which exchanges the authorization code for tokens and stores them for that `(user, project, server)`.
+5. The user's tools for that server now appear in chat and execute under their own token. The SDK refreshes the access token automatically using the stored refresh token.
+
+Tokens are stored in the `mcp_oauth_token` table; the short-lived authorization handshake is stored in `mcp_oauth_flow`. Both are keyed per user and cascade-deleted with the user or project.
+
+## Configuration
+
+Add an `oauth` block to the server entry in `agent/mcps/mcp.json`:
+
+```json
+{
+	"mcpServers": {
+		"mixpanel": {
+			"transport": "streamable-http",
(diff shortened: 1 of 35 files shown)


Bl3f (commented):
The current annoying part with the MCP is that it requires to go in the Settings in the MCP page to connect for the LLM to be aware of the MCP server. Which means a user who want to use Mixpanel will have first to go in the settings to auth in order for the MCP to be discoverable

I think it should be done a bit differently, if as an admin you have registered the MCP the LLM should be aware it exists but that i requires OAuth connection, when a user asks something related the auth should be inline in the conversation so it’s convenient for the user.

For this reason I think we cant merge as is. We will work on it asap to implement what I just said AND significantly improve how MCP behaves in the app.

Bl3f:
@dennisUhrskov we did not forget you! I've opened a new PR that implements Oauth + improve how the MCP client works (cf. #1049). I will close this PR once the other has been merged!

Bl3f:
The other PR has been merged, it will land in 0.2.6 asap.

PR #985, opened 2026-06-25
fix(compaction): ignore blank summaries during rehydration

### LINKED ISSUE #950: [bug] Compaction produces empty text blocks - breaks conversation
## Problem

Bug of non compaction during a long conversation with multiple stories:
The chat shows "No response" and retrying produces the API error:
> messages: text content blocks must be non-empty

The conversation cannot continue at this point with Claude, can continue with GPT though.
But the compaction sign stills stays full

Image

sahniaditya007:
## Summary

Fixes #950.

When conversation compaction fails, an empty summary (`summary: ""`) is persisted. During conversation reconstruction, this empty summary is treated as a valid compaction and reconstructed into an assistant message containing an empty text block.

Anthropic rejects requests containing empty text content blocks, causing the conversation to fail.

## Changes

* Ignore blank compaction summaries in `findLastCompactionPart()`.
* Add a regression test covering failed compactions with empty summaries.

## Result

Failed compactions are treated as absent during rehydration, preventing invalid empty text blocks from being constructed while preserving the existing compaction metadata.

## Testing

* Added a regression test covering failed compactions with blank summaries.
* Verified the new regression test passes.
* Confirmed no unrelated files are modified by this PR.

<a href="https://cubic.dev/pr/getnao/nao/pull/985?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/tests/compaction.test.ts
+++ b/apps/backend/tests/compaction.test.ts
@@ -169,4 +169,23 @@ describe('compactionService.useLastCompaction', () => {
 		expect(result[3]).toEqual(messages[4]);
 		expect(result).toHaveLength(4);
 	});
+
+	it('returns messages unchanged when the only compaction entry has a blank summary (failed compaction)', () => {
+		// Regression test for issue #950:
+		// When compaction fails, summary: '' is persisted to the DB.
+		// Re-hydrating that blank summary creates { type: 'text', text: '' } in the assistant message,
+		// which Anthropic rejects with "text content blocks must be non-empty".
+		const messages: UIMessage[] = [
+			{ id: '1', role: 'user', parts: [{ type: 'text', text: 'Hello' }] },
+			{
+				id: '2',
+				role: 'assistant',
+				parts: [{ type: 'data-compaction', data: { summary: '', error: 'LLM error' } }],
+			},
+			{ id: '3', role: 'user', parts: [{ type: 'text', text: 'Follow-up' }] },
+		];
+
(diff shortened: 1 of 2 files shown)


apps/backend/tests/compaction.test.ts
```diff
@@ -169,4 +169,23 @@ describe('compactionService.useLastCompaction', () => {
 		expect(result[3]).toEqual(messages[4]);
 		expect(result).toHaveLength(4);
 	});
+
+	it('returns messages unchanged when the only compaction entry has a blank summary (failed compaction)', () => {
+		// Regression test for issue #950:
+		// When compaction fails, summary: '' is persisted to the DB.
+		// Re-hydrating that blank summary creates { type: 'text', text: '' } in the assistant message,
+		// which Anthropic rejects with "text content blocks must be non-empty".
```
socallmebertille, line 177: I think these comments are just a duplicate of the previous description so maybe you can remove them

socallmebertille (commented):
Thanks a lot for your contribution, I would appreciate it if you can take my next comment in consideration, run `npx run prettier .` to pass the check and then we'll be able to merge it ! 👍

PR #992, opened 2026-06-26
[Fix] resolve bug from clustered columns from several databases

### LINKED ISSUE #932: [bug] Profiling fails on clustered tables on BigQuery - creates empty profiling.md
## Problem

When running `nao sync` on BigQuery tables that use clustering, the profiling step fails silently and produces an empty `profiling.md` file. Non-clustered tables in the same dataset profile correctly.

This is a similar pattern to previously fixed issues (#542, #351) but with a different root cause specific to clustered tables.

## Expected behavior

Profiling should complete successfully on clustered BigQuery tables and generate a populated `profiling.md` with column-level statistics (mean/max for numbers, top values for text).

socallmebertille:
## Summary

- fixed a bug where profiling failed on BigQuery tables that use clustering, which left an empty `profiling.md` file
- clustering and partitioning are now handled separately, so the profiling query runs correctly and the file gets filled with column statistics
- added the same clustering support for Databricks, Redshift and Snowflake, so it works consistently across databases
- the generated `profiling.md` now also shows which columns a table is clustered on

## Related issue

#932 

<a href="https://cubic.dev/pr/getnao/nao/pull/992?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/nao_core/config/databases/bigquery.py
+++ b/cli/nao_core/config/databases/bigquery.py
@@ -68,11 +68,13 @@ class BigQueryDatabaseContext(DatabaseContext):
         self._custom_partition_filter = custom_partition_filter
 
     def partition_columns(self) -> list[str]:
-        if self._partition_metadata is not None and self._partition_metadata.partition_column is not None:
+        if self._partition_metadata is not None:
+            if self._partition_metadata.partition_column is None:
+                return []
             return self._filter_excluded_names([self._partition_metadata.partition_column])
         try:
             return self._filter_excluded_names(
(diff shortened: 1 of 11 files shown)


Bl3f (approved):
LGTM

PR #1083, opened 2026-07-06
Feat/add data labels to chart 396

ad4mou:
Added a toggle to chart settings which adds value labels to charts

CleanShot 2026-07-06 at 11 15 36@2x
CleanShot 2026-07-06 at 11 15 46@2x

**One thing to note:**

To avoid it being too cluttered, when there are many points in the chart ( > 12), we keep only local and global maximums. See here for example (if you think the chart looks weird, it was generated that way regardless of this feature)

image

<a href="https://cubic.dev/pr/getnao/nao/pull/1083?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ad4mou:
Just noticed I commited changes from my previous PR without realizing, sorry about that

.cursor/rules/manager-delegates-to-subagents.mdc
```diff
@@ -0,0 +1,34 @@
+---
```
Bl3f: I'd not add this in the repo, can you keep it locally on your computer? Pbly you can add a .gitignore rule to ignore the `rules/` folder

apps/frontend/src/components/tool-calls/display-chart-edit-dialog.tsx
```diff
@@ -183,6 +184,17 @@ export function ChartConfigEditDialog({
 						</div>
 					</div>
 
+					<div className='flex items-center justify-between rounded-lg border border-border/60 bg-panel px-3 py-2'>
```
Bl3f: The style of this component is not really integrated in the dialog:
* i'd like the height of the component to be the same as other inputs in the dialog to be uniform
* i think the text can be normal (and not bold)
* i'd move this to a "Options" part below the series (potentially in this case remove the border and the background of the component, but this one up to you)

apps/frontend/src/components/command-menu.tsx
```diff
@@ -53,6 +54,7 @@ export function CommandMenu() {
 			{
```
Bl3f, line 54: As these changes comes from another PR it has to be removed when you create this PR.

The best way is when you create this PR you have to start again from main instead of starting from your previous PR. 

Pls remove it (or rebase from main as it will disapear as we merge it into main)

apps/shared/src/chart-builder.tsx
```diff
@@ -25,6 +26,14 @@ import * as displayChart from './tools/display-chart';
 export const DEFAULT_COLORS = ['#104e64', '#f54900', '#009689', '#ffb900', '#fe9a00'];
 
 const AXIS_TICK = { fontSize: 12 };
+const DATA_LABEL_PROPS = {
```
Bl3f, line 29: I'm a bit afraid of all the changes added to this file, I feel it's a bit too much code, are we sure we need all the things that have been added here?

apps/shared/src/chart-builder.tsx
```diff
@@ -257,7 +317,14 @@ function buildAreaChart(props: ResolvedProps) {
 				})}
 			</defs>
 			{showGrid && <CartesianGrid horizontal vertical={false} strokeDasharray='3 3' />}
-			<YAxis tick={AXIS_TICK} tickLine={false} axisLine={false} minTickGap={12} tickFormatter={formatYAxisTick} />
+			<YAxis
```
Bl3f: Yaxis is moving when you add a label (cf. below) with show data labels the max Y is 683,4 which is a bit weird, i think the max Y should not move or either the number should be at least round like 700 or 680, but even 680 is weird, staying on the previous value without the option activated might be the best. 

I think it's linked to the fact here there are too many changes to display labels, I'd try to write it in a more compact and readable way.

Image

Image

Bl3f (changes requested):
Hello, the feature works great! But a few changes to be done to the code and a simplification if possible.

.cursor/rules/manager-delegates-to-subagents.mdc
ad4mou: Sure, it's done

apps/frontend/src/components/tool-calls/display-chart-edit-dialog.tsx
ad4mou: You're right it was sticking out.
I think this should be good now
CleanShot 2026-07-10 at 10 31 15@2x

apps/frontend/src/components/command-menu.tsx
ad4mou: Yeah I'll be more careful to switch from main not from my previous PR.
I rebased from main so my previous PR's changes don't appear here anymore

apps/shared/src/chart-builder.tsx
ad4mou: You're right, this was done to ensure that the data label would fit in the graph even if the bar hit the max value on the axis.
I ajusted that to the following:
If the bar is close to being the max value on the y axis, meaning its label likely won't be visible or will have the graph's title touching it -> Then we compress the whole graph a little bit so that it stays properly scaled but we get a little padding at the top where we can put that label.

apps/shared/src/chart-builder.tsx
ad4mou: Well I tried cleaning it up but I actually ended up nearly making it bigger because there were a few issues in my code.

In the end, this file now contains:

-  There's the label element code
- Code to handle stacked charts so we show only one label at the top of the bar of stacked values
- Code to make sure labels don't clutter the graph by only showing them for local/global maxima when there are many points
- Code that compresses the graph to add padding at the top if and only if one bar is close to the maximum value on the y axis.


I couldn't remove any of those without making the feature bad in one way. I suggest making a few auxiliary files like `chart-labels.tsx` and `chart-utils.tsx` to declutter this one. Should I do that?

apps/shared/src/chart-builder.tsx
```diff
@@ -141,18 +138,85 @@ function buildResolved(props: BuildChartProps) {
...
+	for (const row of data) {
+		const total = sumStackValue(row, series);
+		if (total != null && (max == null || total > max)) {
+			max = total;
+		}
+	}
+	return max;
+}
+
+export function niceAxisMax(dataMax: number, tickCount = 5): number {
```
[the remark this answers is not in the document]
ad4mou: fair enough

Bl3f (approved):
LGTM now, ty

PR #1106, opened 2026-07-08
fix: stop reporting 64-bit integer columns as int32 in profiling

### LINKED ISSUE #1104: profiling.md reports 64-bit integer columns as int32
`cli/nao_core/config/databases/context.py` normalizes the ibis type `int64` to `int32` in `_normalize_type`.

As a result, generated `profiling.md` files document 64-bit integer columns as `int32`, including:

- BigQuery `INT64`
- Snowflake `INTEGER` / `NUMBER(38,0)`, which ibis reads back as `int64`
- `BIGINT` columns on other backends

This appears to only affect generated documentation: I could not find code that depends on this normalized label directly. 
However, the docs are read by the agent, so the schema description is wrong and can influence generated SQL.

The mapping was introduced in #399, but I could not find a recorded rationale. 
@socallmebertille, do you remember whether this was intentional, for example to harmonize tests across warehouses?

If there is no compatibility reason to keep it, I’d like to submit a PR removing the `int64 -> int32` mapping.

Related context: I found this while preparing a separate fix for BigQuery `DATETIME` / `TIMESTAMP` being documented incorrectly in `columns.md`. I’ll open that as a separate issue.

tckaros:
Closes #1104

## Problem

`_normalize_type` in `cli/nao_core/config/databases/context.py` mapped the ibis type `int64` to `int32`, so every 64-bit integer column was documented as `int32` in generated `profiling.md` files: BigQuery `INT64`, Snowflake `INTEGER` (`NUMBER(38,0)`, which ibis renders as `int64`), and any `BIGINT` column on other backends. The schema section of the same doc reports the raw type, so the two sections contradicted each other — and since the doc is read by the agent, the wrong label can influence generated SQL.

The mapping was introduced in #399 without a recorded rationale, and nothing in the codebase depends on the `int32` label (downstream checks only do substring matching on `"int"`).

## Changes

- Remove the `int64` → `int32` mapping from `_normalize_type`
- Update BigQuery and Snowflake integration expectations to `int64` (the only suites seeding 64-bit columns — all other backends seed 32-bit `INTEGER` and are unaffected)
- Add unit tests locking `_normalize_type` behavior

<a href="https://cubic.dev/pr/getnao/nao/pull/1106?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/nao_core/config/databases/context.py
+++ b/cli/nao_core/config/databases/context.py
@@ -459,8 +459,6 @@ class DatabaseContext:
         lowered = normalized.lower()
         if lowered.startswith("string(") and normalized.endswith(")"):
             return "string"
-        if normalized == "int64":
-            return "int32"
         return normalized
 
     @staticmethod
--- a/cli/tests/nao_core/commands/sync/integration/test_bigquery.py
+++ b/cli/tests/nao_core/commands/sync/integration/test_bigquery.py
@@ -147,7 +147,7 @@ def spec(db_config, temp_datasets):
         users_profiling_rows=[
             {
                 "column": "id",
-                "type": "int32",
+                "type": "int64",
                 "total_count": 3,
                 "null_count": 0,
                 "null_percentage": 0.0,
@@ -195,7 +195,7 @@ def spec(db_config, temp_datasets):
         orders_profiling_rows=[
             {
                 "column": "id",
-                "type": "int32",
+                "type": "int64",
                 "total_count": 2,
                 "null_count": 0,
                 "null_percentage": 0.0,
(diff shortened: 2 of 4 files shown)


Bl3f:
LGTM, ty @tckaros

PR #1141, opened 2026-07-10
fix(slack): render result tables as monospace instead of Block Kit table blocks

btalx:
## Problem

When an answer contains a markdown table, the Slack integration renders it as a Block Kit `table` block. Real query output breaks that block's schema, so `@slack/web-api` rejects the whole message and the user gets no answer:

- empty cells omit a required cell value
- long cell text exceeds the allowed length
- some values map to an invalid element type

`ProjectSlackBot.postMessage` attaches the blocks with no fallback, so a rejected payload drops the entire answer. Observed as repeated `WebClient` validation errors (`missing required field ... /blocks/N/rows/N`, `must be more than N characters`, `failed to match any allowed schemas`).

## Fix

- Render markdown tables as fixed-width monospace text inside a code block (a `section`/mrkdwn block), which is not subject to the `table` block schema. Tables are chunked to stay under the 3000-char section limit, cells are clamped, and very long tables are capped with an overflow hint.
- Apply the same monospace rendering to the plain-text `text` field so the notification fallback is readable.
- As defense in depth, retry `postMessage` without blocks if a block payload is ever rejected, so an answer is never silently dropped.

Both the direct post path and the streaming edit path go through `createTextBlocks`, so both are covered.

## Testing

Updated `slack-message-blocks.test.ts` to the new contract and added cases for the exact triggers: empty cells, missing/NULL cells, an oversized cell, a wide table that chunks, and the row cap.

<a href="https://cubic.dev/pr/getnao/nao/pull/1141?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/services/slack.ts
+++ b/apps/backend/src/services/slack.ts
@@ -4,7 +4,7 @@ import { createSlackAdapter } from '@chat-adapter/slack';
 import { createMemoryState } from '@chat-adapter/state-memory';
 import { CITATION_TAG_REGEX } from '@nao/shared';
 import type { LlmSelectedModel } from '@nao/shared/types';
-import { type ChatPostMessageArguments, WebClient } from '@slack/web-api';
+import { type ChatPostMessageArguments, type ChatPostMessageResponse, WebClient } from '@slack/web-api';
 import { InferUIMessageChunk, readUIMessageStream } from 'ai';
 import { Card, Chat, deriveChannelId, Message, SentMessage, Thread, ThreadImpl } from 'chat';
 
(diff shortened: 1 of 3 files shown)


Bl3f:
Hey @btalx not sure I understand why we should do this as Table blocks works well on you side. 

Do you have a precise example of this not working?

PR #1149, opened 2026-07-11
Honor repo include/exclude globs for URL repos

### LINKED ISSUE #1140: nao sync: repo include/exclude ignored for url repositories (only applied to local_path)
## Summary
In `nao_config.yaml`, a `repos:` entry with a `url:` accepts `include`/`exclude` globs (schema-valid) but they have **no effect** — the full repo is always synced. The filters only work for `local_path:` repos.

## Repro
```yaml
repos:
- name: dbt
  url: https://github.com/getnao/nao-dbt.git
  include:
  - models/silver/**
```
Run `nao sync -p repos`.
- **Expected:** only `models/silver/**` synced into `repos/dbt/`.
- **Actual:** the entire repo is cloned into `repos/dbt/` (include ignored).

## Root cause
`commands/sync/providers/repositories/provider.py`: `sync_repo()` dispatches `url` repos to `clone_or_pull_repo()`, which `git clone`s the whole repo, strips `.git/`, and atomically swaps it in — it never calls `_matches_patterns()`. Only `sync_local_repo()` (the `local_path` path) applies include/exclude.

Confirmed in **0.2.1** and latest **0.2.10**.

## Impact
- Can't scope a large shared dbt/BI repo down to just the relevant models via config.
- Forces manual pruning after each sync, or a post-sync cleanup step in CI.
- `include`/`exclude` silently do nothing for `url` repos, which is surprising given the schema accepts them.

## Suggested fix
Apply `_matches_patterns()` in the `url` path too (filter cloned files before the atomic swap), so `include`/`exclude` behave the same for `url` and `local_path` repos. Alternatively, document that filtering is `local_path`-only and/or warn when `include`/`exclude` is set on a `url` repo.

ClaireGz:
# Issue
Closes #1140

# Implemented

* Apply `include`/`exclude` glob filtering to `url` repositories before the atomic swap (previously only `local_path` repos were filtered)
* Remove excluded and dangling symlinks safely; skip directory symlinks when pruning empty directories

# Tests done

* Unit tests for include/exclude on `url` repos and symlink edge cases
* `make lint` clean; CI green

### DIFF
--- a/cli/nao_core/commands/sync/providers/repositories/provider.py
+++ b/cli/nao_core/commands/sync/providers/repositories/provider.py
@@ -138,13 +138,14 @@ def _filter_repo_files(repo_dir: Path, include: list[str], exclude: list[str]) -
     if not include and not exclude:
         return
 
-    for file_path in repo_dir.rglob("*"):
-        if not file_path.is_file():
+    for entry in repo_dir.rglob("*"):
+        # Match symlinks against the link itself; never follow them.
+        if not entry.is_symlink() and not entry.is_file():
             continue
 
-        relative = PurePosixPath(file_path.relative_to(repo_dir)).as_posix()
+        relative = PurePosixPath(entry.relative_to(repo_dir)).as_posix()
         if not _matches_patterns(relative, include, exclude):
-            file_path.unlink()
+            entry.unlink()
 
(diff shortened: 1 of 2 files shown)


Bl3f (approved):
LGTM

PR #1150, opened 2026-07-11
User friendly automation scheduling without cron

### LINKED ISSUE #1130: [feature] Friendly automation scheduling (daily at HH:MM) instead of raw cron
## Problem
Setting up a scheduled automation at "Daily" defaults to using the hour of automation setup. Setting up a specific hour requires writing a cron expression, which is unfriendly for non-technical users.

## Expected behavior
Provide a simple scheduling UI (e.g. "daily at 09:00", "every Monday at 08:00") that generates the cron under the hood, so users don't have to write cron syntax.

## Customer ID

ClaireGz:
# Issue
Closes #1130

# Implemented

* Replace the raw-cron trigger with friendly pickers: hourly (free minute), daily, weekdays, weekly (multiple days), monthly (days 1–31), each with an HH:MM time
* Keep an Advanced (raw cron) option showing a live human-readable description
* Times run in the server timezone (shown in the UI)

# Tests done

* Unit tests for cron build/parse round-trips across all frequencies
* ESLint/Prettier CI green
* Checked UI - screenshot below

image

### DIFF
--- a/apps/frontend/src/components/automations-form.tsx
+++ b/apps/frontend/src/components/automations-form.tsx
@@ -167,7 +167,7 @@ const scheduleFrequencyOptions: Array<{ value: ScheduleFrequency; label: string 
 ];
 
 const hourlyMinuteOptions = Array.from({ length: 12 }, (_, index) => index * 5);
-const dayOfMonthOptions = Array.from({ length: 28 }, (_, index) => index + 1);
+const dayOfMonthOptions = Array.from({ length: 31 }, (_, index) => index + 1);
 
 export function AutomationForm({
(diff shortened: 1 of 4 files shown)


Bl3f (changes requested):
Advanced mode visual is truncated. I also think we can remove the "Monthly on the 1st at 08:00" because when complex cron it writes "Custom schedule" which is useless and takes a row visually for nothing.

I'd also be in favour of putting space between the "On schedule" and the different inputs so it's aligned right like it was before.

Screenshot 2026-07-15 at 17 24 51

PR #1151, opened 2026-07-11
Write context rules: add personalised design-system section

ClaireGz:
## What
Adds a **Chart & Visualization Guidelines** section to the `write-context-rules` skill so generated `RULES.md` files tell the agent how charts should look:

- **Brand color palette** for chart series (single-series → primary; multi-series → spread across the palette, then tints/shades beyond ~6).
- **Semantic colors** (good vs. bad) used only when a chart encodes positive/negative.


Bl3f (approved):
LGTM

Bl3f:
I accept this but imho it should be done deterministically rather than with guidelines that consuming tokens for nothing.

PR #1154, opened 2026-07-11
Add BigQuery dialect rules to system prompt

### LINKED ISSUE #1099: [feature] Add BigQuery dialect rules to the agent system prompt
## Problem
These are BigQuery mistakes agent has made on internal nao analytics agent.
We should add these to the BigQuery specific prompt

### Cases observed

**1. Window function inside a `GROUP BY` query** (chat `eebbf778`)

```sql
SUM(SUM(deal_arr_usd)) OVER (ORDER BY DATE_TRUNC(DATE(created_at), MONTH))
```

BigQuery error: `Window ORDER BY expression references column created_at which is neither grouped nor aggregated`. Every column referenced in a window `ORDER BY` / `PARTITION BY` inside a `GROUP BY` query must appear in the `GROUP BY` or be an aggregate.

**2. Inline `VALUES` table constructor** (chat `ba179377`)

`FROM (VALUES (...)) AS t(...)` is standard SQL but not supported in BigQuery, producing `400 Syntax error: Expected keyword JOIN but got ","`. BigQuery needs `SELECT ... UNION ALL SELECT ...` or `UNNEST([STRUCT(...), STRUCT(...)])` for inline literal rows.

**3. `STRING_AGG(DISTINCT ... IGNORE NULLS)`** (chat `7e4cf747`)

`STRING_AGG(DISTINCT CASE WHEN ... THEN project_domain END IGNORE NULLS)` produces `400 IGNORE NULLS and RESPECT NULLS are not allowed for aggregate function string_agg`. BigQuery does not support `IGNORE NULLS` / `RESPECT NULLS` on `STRING_AGG`.

**4. `BOOL_OR` vs `LOGICAL_OR`, and `ORDER BY` on unaggregated columns** (chats `1323a875`, `8404b36d`, `c9cca11f`, `9e8d6692` - recurring in the OSS new-projects daily automation)

`BOOL_OR()` (PostgreSQL/DuckDB) is not supported in BigQuery; the equivalent is `LOGICAL_OR()`. Companion failures: an `ORDER BY` referencing a column that is not in `GROUP BY` and not aggregated (e.g. `e.is_local`) raises a 400.

ClaireGz:
# Issue
Closes #1099

# Implemented

* Add BigQuery-specific rules to the agent system prompt:
  * window `ORDER BY` / `PARTITION BY` columns must be grouped or aggregated
  * no inline `VALUES` table constructor
  * no `IGNORE NULLS` / `RESPECT NULLS` on `STRING_AGG`
  * use `LOGICAL_OR` / `LOGICAL_AND` instead of `BOOL_OR` / `BOOL_AND`
* Rules inject only when a BigQuery connection is present

# Tests done

* ESLint/Prettier CI green (prompt-only change)

### DIFF
--- a/apps/backend/src/components/ai/dialect-rules.tsx
+++ b/apps/backend/src/components/ai/dialect-rules.tsx
@@ -68,6 +68,26 @@ export const DIALECT_GUIDANCE: DialectGuidance[] = [
 			<>
 				Use <Code>SAFE_DIVIDE</Code> for division to avoid division-by-zero errors.
 			</>,
+			<>
+				In a <Code>GROUP BY</Code> query, every column referenced inside a window <Code>ORDER BY</Code>/
+				<Code>PARTITION BY</Code> must be in the <Code>GROUP BY</Code> or wrapped in an aggregate. Wrong:{' '}
+				<Code>SUM(SUM(x)) OVER (ORDER BY DATE_TRUNC(DATE(created_at), MONTH))</Code>. Right: group by the month
+				first, e.g. <Code>SUM(SUM(x)) OVER (ORDER BY month)</Code>.
+			</>,
+			<>
+				Do not use the inline <Code>VALUES</Code> table constructor (<Code>FROM (VALUES (...)) AS t(...)</Code>)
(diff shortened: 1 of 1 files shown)


Bl3f (approved):
LGTM

PR #1164, opened 2026-07-12
Bug fix - / makes charts invisible in stories

### LINKED ISSUE #1084: [bug] \ in chart series definition makes the chart not show
If in series, label as a "\", the json cannot be parsed and the story shows "series was not selected" instead of the cart

## Customer ID

ClaireGz:
# Issue
Closes #1084

# Implemented

* Repair series JSON when a label contains a literal backslash (e.g. `Disc\Rebate`) so the chart still renders
* Require `\u` escapes to be followed by 4 hex digits
* Quote-aware `<chart>` / `<table>` tag parsing so `/` or `>` inside a title or label (e.g. `title="13/07 update"`) no longer truncates the tag; centralized into shared helpers

# Tests done

* Regression tests for backslash labels, `\u` escapes, and slashes in titles/labels
* ESLint/Prettier CI green
* Tested in UI:

image


apps/shared/src/story-segments.ts
```diff
@@ -20,6 +20,37 @@ export type Segment =
 	| { type: 'table'; table: ParsedTableBlock }
 	| { type: 'grid'; cols: number; children: Segment[] };
 
+/**
```
Bl3f: Remove all these comments pls, it's useless the LLM adds it but we don't want them in the code.

apps/shared/src/story-segments.ts
```diff
@@ -87,15 +118,33 @@ export function getGridClass(cols: number): string {
 	return GRID_CLASSES[Math.min(cols, 4)] ?? GRID_CLASSES[2];
 }
 
-function tryParseSeriesJson(value: string): ParsedChartBlock['series'] | null {
+/**
+ * Parses a chart `series` JSON array. When the first parse fails, it retries after
```
Bl3f: Same here, remove the comments.

Bl3f (approved):
Can you just remove the comments pls.

apps/shared/src/story-segments.ts
ClaireGz: Done — removed these comments (and the other added JSDoc/regex comments in this file). Thanks!

apps/shared/src/story-segments.ts
ClaireGz: Done — removed these comments (and the other added JSDoc/regex comments in this file). Thanks!

PR #1178, opened 2026-07-13
fix(slack): keep result tables within Slack table block limits

### LINKED ISSUE #1143: Slack: answers containing a table are dropped (Block Kit table block rejected)
We self-host nao and Slack is the main way our team asks questions. Whenever an
answer contains a table, the message never arrives. The user just gets an error
or nothing back.

The Slack client renders tables as a Block Kit `table` block, and real query
output fails Slack's schema, so @slack/web-api rejects the whole message. From
our logs:

    WebClient missing required field: element / elements / value [/blocks/N/rows/N/N]
    WebClient must be more than N characters [.../text]
    WebClient failed to match any allowed schemas

It's triggered by ordinary data: empty cells, very long cell values, and so on.
And postMessage sends the blocks with no fallback, so a rejected payload drops
the whole answer instead of degrading to text.

Why it matters: for a Slack-first deployment this silently loses answers for
real users any time the result has a table.

I've already fixed this on our fork (render tables as monospace text instead of
the table block, plus a text-only fallback if a block is ever rejected) and
tested it against our live Slack workspace. I'd like to open a PR.

btalx:
Follow-up to #1141. That PR dropped the native table block for monospace; @Bl3f asked to keep the table block, so this version keeps it and just makes it fit Slack's limits.

## The precise example you asked for

Slack's `table` block has hard limits: 100 rows, 20 columns, and **10,000 characters summed across every cell** (https://docs.slack.dev/reference/block-kit/blocks/table-block). `@chat-adapter/slack` already caps rows/columns and pads empty cells, but nothing enforces the total-character budget or bounds a single oversized cell.

So an ordinary result tips it over: ~80 rows with a longish text/URL column, or one cell holding a big value, sums past 10,000 chars. `@slack/web-api` then rejects the whole `table` block, and because `postMessage` attaches blocks with no fallback, the entire answer is dropped. That is what we see in our Slack-first self-hosted deployment — the table renders fine until a result is large enough, then the message never arrives.

## Fix

Keep the native table block, but trim it to fit before building it:

- clamp each cell to a max length,
- keep rows within the 100-row and 10,000-char budget,
- append a small `…N more rows, open in nao` note when rows are dropped.

No monospace fallback — the rich table stays. As defense in depth, `postMessage` retries text-only if Slack ever rejects a block payload, so an answer is never silently lost. Both the direct post and the streaming edit path build tables through `createTextBlocks`, so both are covered.

## Testing

- `npm run -w @nao/backend test` — extended `slack-message-blocks.test.ts` with cell clamping, the 100-row cap, the 10,000-char drop + note, and a payload-level check that the rendered `table` block stays within 10,000 chars using the real `@chat-adapter/slack`.
- `npm run lint` and `prettier --check` pass.
- The drop was observed against our live Slack workspace (self-hosted nao); this change keeps the block within Slack's documented limits.

Fixes #1143.

<a href="https://cubic.dev/pr/getnao/nao/pull/1178?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f (approved):
LGTM, ty so much!

ad4mou:
This PR was great work. I just couldn't tell if it was fully functionnal because of other bugs so I added other fixes to be sure Slack now works perfectly.

## Summary

Fixes Slack answers being dropped or cut short, and cleans up the buttons around them.

- **Long answers.** Replies over ~4,000 characters used to fail silently, freezing mid-sentence with no explanation. Text is now split into pieces Slack accepts, so an answer of any realistic length arrives in one message. If someone truly asks for a very very long answer (more than 145k chars) , they get a short "too long to show fully in Slack, open in nao" line rather than silence.
- **Tables.** Slack only allows one table per message. Extra tables now leave an indented **[ Table 2 ]** marker where they belong, and the button at the bottom of the answer becomes "Open the other 2 tables in nao" instead of adding more buttons. Tables are trimmed to what Slack accepts before sending, so a large result no longer kills the whole message. Text written before a table is no longer dropped.

Closes #1323
Closes #1143

ad4mou (approved):

Bl3f (approved):
LGTM

Bl3f:
@ad4mou there is a conflict with main!

ad4mou:
Ok I'll take care of it

PR #1181, opened 2026-07-14
Number table data rows starting at one

### LINKED ISSUE #1179: Table component row numbering should start at 0
# Problem

Today, it's confusing for the user because the title line counts as 1
So the bottom number doesn't actually give number of rows in the table

Image

# Issue
We should either start counting at 0, or put no number on the header lines

ClaireGz:
# Issue
Closes #1179

# Implemented

* Header row's index cell is now blank (no number) instead of showing "1".
* Data rows are numbered starting at 1 (first data row = 1, then 2, 3, …), so the index column counts only data rows.
* The bottom row count already reflects the number of data rows now that the header is not counted.

# Preview
image


Bl3f (approved):
LGTM

PR #1189, opened 2026-07-15
feat: Add an annotations file for each table, and a header whether files are managed by the agent or by the user

### LINKED ISSUE #1042: [feature] Add annotations.md in each table folder for manual annotations + headers to distinguish auto / manual files
## Problem

- Auto-generated files are indistinguishable from manual ones (which can be erased)
- There is no clear way to write manual annotations on one specific table

## Expected behavior

- [ ] **Header on generated files** - add a header on every auto-generated file marking it as machine-generated.
- [ ] **Persistent `annotations.md`** - always create an empty `annotations.md` file that is never overwritten on future syncs/regenerations.

## Related issues
#1011 -> make sure the templates selected in nao init matches the new naming

ad4mou:
Small change to the cli
`nao sync` now generates an `Annotations.md` file for each table where the user can add some rules for the agent.

Additionally, all .md files of tables now have a header which specifies if they're managed by the agent, hence if they can't be modified. Or if they're managed by the user, which means the user can freely modify it.

Closes #1042

<a href="https://cubic.dev/pr/getnao/nao/pull/1189?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cli/nao_core/commands/sync/markers.py
```diff
@@ -0,0 +1,48 @@
...
+    ContextFileKind.GENERATED: "This file is managed and generated by the agent, do not modify it!",
+    ContextFileKind.MANUAL: "These are manual notes you want the agent to keep in mind, safe to edit.",
+}
+
+_MARKER_PATTERN = re.compile(r"<!--\s*nao:(generated|manual)\b")
+
+
+def marker_for(kind: ContextFileKind) -> str:
+    """Build the single-line HTML marker carrying the stable token and human message."""
+    return f"<!-- nao:{kind.value} | {_MESSAGES[kind]} -->"
```
Bl3f: I don't like the header to be this way. I'd go for a header like the skills cf below:

```
---
type: manual/generated
comment: These are manual notes you want the agent to keep in mind, safe to edit.
---
```

and eventually later we would add groups or something similar

Bl3f (changes requested):
Small comment, i'd change the format of the header

cli/nao_core/commands/sync/markers.py
ad4mou: ok sure, that's done now

Bl3f (approved):
LGTM

PR #1191, opened 2026-07-15
feat(cli): show reference SQL and readable agent SQL in test server

### LINKED ISSUE #1190: Test server: show reference SQL and render executed SQL readably
When a test fails, the test server UI doesn’t make it easy to compare what the agent ran with the reference query from the test YAML:
- The YAML sql: value isn’t saved in tests/outputs/results_*.json. It’s sent to the backend to produce expectedData, then discarded.
- SQL executed by the agent is difficult to read. Tool-call arguments are rendered as stringified JSON, so queries appear on one line with escaped \n characters.

Proposal (CLI-only; no backend changes):
1. Save the YAML query as reference_sql in the results JSON, including for errored runs where this context is especially useful.
2. Display it in a small “Reference SQL” section in the result modal.
3. Render queries from execute_sql tool calls as multi-line SQL instead of escaped JSON.

Existing result files without reference_sql would continue to render normally, without the new section.
I’m deliberately calling it “reference” rather than “expected”: the agent’s SQL isn’t expected to match the query, only the resulting data.

I’ve attached screenshots of the result.
I’ve opened a PR with a working implementation to make the proposal concrete. I’m happy to rework it based on feedback, and no worries if this isn’t a direction you’d like to pursue.

Image
Image

tckaros:
Closes #1190

- Save the test YAML query as `reference_sql` in `tests/outputs/results_*.json` (also on errored runs, where it's the only debugging context available).
- Show it in a "Reference SQL" section in the result modal.
- Render `execute_sql` tool call queries as multi-line SQL instead of escaped JSON; other tools keep the JSON rendering.

Old result files (without `reference_sql`) still render fine — the new section is conditional, same pattern as the other modal sections.

Note: `TestCase.sql` is annotated `str` but `from_yaml` can assign `None` (pre-existing); `reference_sql` is typed `str | None` to handle it. Happy to fix the annotation here or in a follow-up if you prefer.

Tested with `make lint` and pytest (3 new tests covering the field across all `run_test` branches), plus the UI checked against result files covering pass/fail/errored/legacy cases. Before/after screenshots are on the issue.

This PR was written using Claude Fable 5 (claude-fable-5).

<a href="https://cubic.dev/pr/getnao/nao/pull/1191?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/nao_core/commands/test/runner.py
+++ b/cli/nao_core/commands/test/runner.py
@@ -51,6 +51,7 @@ class TestRunDetails:
     expected_data: list[dict] | None = None
     comparison: str | None = None
     tool_calls: list[dict] | None = None
+    reference_sql: str | None = None
 
 
 @dataclass
@@ -225,6 +226,7 @@ def run_test(
                     expected_data=result.verification.expectedData,
                     comparison=comparison,
                     tool_calls=result.tool_calls,
+                    reference_sql=test_case.sql,
                 ),
             )
 
@@ -241,6 +243,7 @@ def run_test(
             details=TestRunDetails(
                 response_text=result.text,
                 tool_calls=result.tool_calls,
+                reference_sql=test_case.sql,
             ),
         )
 
@@ -252,7 +255,7 @@ def run_test(
             passed=False,
             message="error",
             error=str(e),
-            details=None,
+            details=TestRunDetails(reference_sql=test_case.sql),
         )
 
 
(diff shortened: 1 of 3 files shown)


Bl3f (approved):
LGTM

Issue #1010, opened 2026-07-01
Improve AI summary template


ClaireGz:
## Problem

- AI summary logic produce low-quality output (e.g. conclusions on a 10-line preview).
- It also doesn't use the descriptions on table that could be in the dbt repo

## Expected behavior

- [x] **Improve AI summary** - stop forcing conclusions into a 10-line preview; If available, feed summary with profiling too. Precise in the system prompt that the previes is only 10 lines and not representative of ANY quality issues (but profiling is)
- [ ] See if we can fetch docs on this table in other parts of the repo (eg dbt repo)

Bl3f:
done

PR #1195, opened 2026-07-15
feat: Improve AI summary template by giving it the database profiling

ad4mou:
## Summary
- The AI table summary now bases its data-quality and distribution comments on real column statistics (empty-value counts, unique-value counts, min/max, most common values) computed over the whole table instead of guessing from only the preview. The preview is now clearly treated as a non-representative sample, and the summary stays quiet about quality when no statistics are available.
- Added a refresh setting for the AI summary so you can control how often it regenerates (`always`, `once`, or every N days), just like profiling already has. Existing setups keep regenerating every sync by default.
- Profiling is now computed only once per table and shared with the summary, so nothing is calculated twice or thrown away. Each file still refreshes on its own schedule.

## Out of scope (follow-up)
- Pulling table/column docs from dbt repos

Closes only one half of  #1010

<a href="https://cubic.dev/pr/getnao/nao/pull/1195?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -172,8 +172,23 @@ After syncing, any Jinja templates (`*.j2` files) in the project directory are r
 Optional `ai_summary` generation:
 
 - Add `ai_summary` to a database connection `templates` list to render `ai_summary.md`.
+- AI summaries use profiling statistics for data-quality and distribution observations. The row preview is a tiny, non-representative shape sample.
 - Use `prompt("...")` inside Jinja templates to generate `ai_summary` content.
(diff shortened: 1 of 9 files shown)


Bl3f (approved):
LGTM

PR #1209, opened 2026-07-17
Fix/nao admin bug 1145

### LINKED ISSUE #1145: [bug] nao admin mode switch back to non-admin mode in same conversation
# Problem

I created a conversation in admin mode
At second question, it answered with the real data, not nao data

Image

See conversation in moon:
https://moon.getnao.io/c6e99dec-0f73-4f89-9810-b76086f6684e

ad4mou:
## Summary
In admin mode, follow-up questions used to quietly switch back to the normal warehouse data after the first message. Now admin mode stays on for the whole chat (#1145).

- Ask several questions in a row in admin mode ->  they all stay in admin mode.
- Turn admin mode off mid-chat and the next question uses normal data, as expected.
- Reopening a chat picks up where you left off, in whichever mode the last question used.

Closes #1145

<a href="https://cubic.dev/pr/getnao/nao/pull/1209?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/src/handlers/agent.ts
```diff
@@ -25,14 +26,19 @@ interface HandleAgentMessageResult {
...
-	const { userId, message, messageToEditId, model, mentions, projectId, adminMode } = opts;
+	const { userId, message, messageToEditId, model, mentions, projectId, isProjectAdmin } = opts;
 
 	if (!projectId) {
 		throw new HandlerError('BAD_REQUEST', noProjectMessage());
 	}
 
 	await agentService.assertBudget(projectId, model);
 
+	let adminMode = opts.adminMode ?? false;
```
Bl3f, line 37: unpack the adminMode if used twice

apps/backend/src/handlers/agent.ts
```diff
@@ -25,14 +26,19 @@ interface HandleAgentMessageResult {
...
+	if (opts.adminMode === undefined && isProjectAdmin && opts.chatId) {
+		adminMode = await chatQueries.wasLastUserMessageAdmin(opts.chatId);
```
Bl3f, line 39: not sure if last user message or first user message should be considered, to me if the first message is source=admin then all the messages in the convo should be but the issue with this is that it forces it under the hood even if the user deactivate the admin mode in the frontend

Bl3f (commented):
Not sure i'd do it this way because here if you deactivate the mode on the frontend if the last message is admin then it will stay in admin, i'd stick to admin mode if first message source is admin, wdyt

ad4mou:
Really? I think I made it so that if you deactivate the mode on the frontend it takes priority over the last message mode and hence deactivates the mode. I actually tested it


https://github.com/user-attachments/assets/6095fef2-339d-46aa-a621-1b617ef70b14


This is the behavior I wanted. Do you prefer something else?

PS: sorry for the music, I forgot to mute

apps/backend/src/handlers/agent.ts
ad4mou: Yes actually that's what caused the issue.
I think it makes more sense to just follow the most recent message but if the user changes manually the mode, then that takes priority and hence we change the mode.

Whether that should be the first message or the last, I think most recent makes more sense
But it's true in a practical scenario the user should stick to one mode in one chat. So perhaps we could lock the mode after the first prompt in a chat

PR #1210, opened 2026-07-17
fix: Share dialog now lists everybody with access: org members and project members

### LINKED ISSUE #1088: [bug] Story/chat share dialog only lists org members, not project members
## Problem
When sharing a story or a chat, the share dialog only lists users who are explicitly members of the **project**. Users who belong to the **organization** (and therefore can already access the project through org-level inheritance) do not appear in the share lis

## Expected behavior
The share dialog should respect org to project inheritance: any user who can access the project (whether added directly to the project or via org membership) should be selectable when sharing a story or chat.

ad4mou:
## Summary
Sharing a chat or story now lets you pick anyone who can access the project, including people who have access through the organization, not just those added directly to the project (#1088).

- The share list now includes org members, so you can share a specific chat/story with them.
- Editing who a chat is shared with no longer silently drops org members.
- Team settings is unchanged -> it still shows only direct project members.
- Added a backend test covering direct, org-only, both (project role wins), project-only `context_admin`, and self-hosted (no org) cases.

Closes #1088

<a href="https://cubic.dev/pr/getnao/nao/pull/1210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -126,7 +126,7 @@ export const getUserRoleInProject = async (projectId: string, userId: string): P
 	return orgMember ? orgMember.role : null;
 };
 
-export const listAllUsersWithRoles = async (projectId: string): Promise<UserWithRole[]> => {
+export const listProjectMembersWithRoles = async (projectId: string): Promise<UserWithRole[]> => {
 	const results = await db
 		.select({
 			id: s.user.id,
@@ -143,7 +143,7 @@ export const listAllUsersWithRoles = async (projectId: string): Promise<UserWith
 	return results;
 };
 
-export const listProjectAccessibleUsersWithRoles = async (projectId: string): Promise<UserWithRole[]> => {
+export const listUsersWithProjectAccess = async (projectId: string): Promise<UserWithRole[]> => {
 	const project = await getProjectById(projectId);
 	const results = await db
 		.select({
@@ -197,7 +197,7 @@ export const getProjectByUserId = async (
 };
 
 export const checkProjectHasMoreThanOneAdmin = async (projectId: string): Promise<boolean> => {
-	const userWithRoles = await listAllUsersWithRoles(projectId);
+	const userWithRoles = await listProjectMembersWithRoles(projectId);
 	const nbAdmin = userWithRoles.filter((u) => u.role === 'admin').length;
 	return nbAdmin > 1;
 };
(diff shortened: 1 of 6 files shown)


apps/backend/src/trpc/shared-chat.routes.ts
```diff
@@ -134,7 +134,7 @@ export const sharedChatRoutes = {
 				throw new TRPCError({ code: 'FORBIDDEN', message: 'Only the creator or an admin can update this.' });
 			}
 
-			const projectMembers = await projectQueries.listAllUsersWithRoles(ctx.resource.projectId);
+			const projectMembers = await projectQueries.listProjectAccessibleUsersWithRoles(ctx.resource.projectId);
```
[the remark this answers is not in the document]
ad4mou: fair enough, done

apps/backend/src/queries/project.queries.ts
```diff
@@ -143,6 +143,25 @@ export const listAllUsersWithRoles = async (projectId: string): Promise<UserWith
 	return results;
 };
 
+export const listProjectAccessibleUsersWithRoles = async (projectId: string): Promise<UserWithRole[]> => {
```
Bl3f: Q: why do we create a new function instead of editing `listAllUsersWithRoles`?

Bl3f (commented):
Small question first.

apps/backend/src/queries/project.queries.ts
ad4mou: Because `listAllUsersWithRoles` is necessary for team settings (which only shows project members) and a few other settings which need to know only project members.
Whereas for the share dialog we need everybody (including org members) who has access to this project so we have a separate function for that.
But I realize now that the function naming is not great. I think this would be much clearer:
CleanShot 2026-07-20 at 11 45 53@2x

Bl3f: I like this function naming and perfect thanks for the explanation of the split.

ad4mou: ok sure sounds good

PR #1211, opened 2026-07-19
feat: Create tabs in stories

### LINKED ISSUE #1126: [feature] Create tabs in stories
## Problem
As a user, I want to create a story with several tabs to avoid having very very long stories

## Expected behavior
- Being able to create tabs in a story with the agent + manually in the UI
- Being able to sort / reorder tabs with the agent + manually
- Being able to rename tabs with the agent + manually
- Being able to delete tabs with the agent + manually

## Customer ID

ad4mou:
## Summary

Adds tabs to stories via `<tabs><tab title="…">…</tab></tabs>` markup.

- Renders tabs in the side panel, expanded, standalone, and shared views.
- Tabs can be added, renamed, reordered, and deleted in the edit view (pencil button).
- Edit is tab-aware: edits one tab at a time, saves all tabs as one version. Non-tabbed stories unchanged.
- Validates tab markup.
- Agent can produce `<tabs>` when the prompt calls for it (schema + prompt examples).
- Exports (HTML & PDF) flatten tabs into sequential titled sections, **Interactive tabs in the HTML export** could be a follow-up.

Closes #1126

### Demo

https://github.com/user-attachments/assets/619a1b2e-25bb-4619-8867-5b5b47626129

<a href="https://cubic.dev/pr/getnao/nao/pull/1211?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/src/agents/tools/story.ts
```diff
@@ -17,6 +17,8 @@ export default createTool<story.Input, story.Output>({
 		'Charts are embedded via <chart query_id="..." chart_type="..." x_axis_key="..." series=\'[...]\' title="..." />.',
 		'SQL result tables are embedded via <table query_id="..." title="..." />.',
 		'Use <grid cols="2">...</grid> to display charts side by side in a responsive grid.',
+		'Use <tabs> with <tab title="...">...</tab> blocks to organize a story into top-level tabs.',
+		'Default to a single flowing story. Use <tabs> only when the user asks for tabs, or when the content splits into clearly distinct sections that are better separated than stacked (e.g. overview vs. detail, one topic/department/metric per tab). Avoid tabs for a short or single-topic story. Always follow the user\'s explicit request (e.g. "a tab per chart" means one chart per tab). When using tabs, the story must start with <tabs> and contain only <tab title="...">...</tab> blocks — no text outside a tab.',
```
Bl3f: Is it the best thing to force the story to start with <tabs> when we have tabs? cant we just insert a <tab></tab> or even just insert a <tab> and consider what after in the tab?

Or maybe there is a token optimisation to do, because the thing is when you move a story from non-tab to tab it requires to rewrite the full story then in the current situation of the story tool (mode=update), a single search and replace...

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
```diff
@@ -0,0 +1,218 @@
...
+						{dropSlot === index && (
+							<div className='pointer-events-none h-9 w-[3px] shrink-0 self-center rounded-full bg-primary shadow-[0_0_6px] shadow-primary/40' />
+						)}
+						<div
+							draggable={Boolean(editable) && editingIndex !== index}
+							onDragStart={(event) => handleDragStart(event, index)}
+							onDragEnd={clearDragState}
+							onDragOver={(event) => handleDragOver(event, index)}
+							className={cn(
+								'group/tab flex h-9 shrink-0 items-center rounded-t-md border border-transparent transition-[color,background-color,transform,opacity]',
```
Bl3f: The style here could be slightly improved:
- Remove the border bottom of the active tab so it looks better like a "current tab"
- add cursor:pointer sur l'onglet pour savoir qu'on peut cliquer
- in edit more the "+" button is not vertically centered
- the gap between the tab title and the "x" to close the tab is to large when compared to the right padding of the tab

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
```diff
@@ -0,0 +1,218 @@
...
+									onDoubleClick={() => editable && startRenaming(index)}
+									className='h-full max-w-48 truncate px-3 text-sm font-medium outline-none'
+								>
+									{tab.title || 'Untitled'}
+								</button>
+							)}
+
+							{editable && editingIndex !== index && (
+								<div className='mr-1 flex items-center opacity-0 transition-opacity group-hover/tab:opacity-100 group-focus-within/tab:opacity-100'>
+									<AlertDialog>
```
Bl3f: I feel the dialog style is not the same as elsewhere in the app, the button are not correctly rounded, the color of the button is not the right one, the button should be aligned right, the text should not be align center but left (get inspiration on the story live settings dialog)

Image

apps/shared/src/tools/story.ts
```diff
@@ -19,7 +19,17 @@ export const InputSchema = z.object({
 		.string()
 		.optional()
 		.describe(
-			'The markdown content. Required for "create" (initial content) and "replace" (new content). Can include charts via <chart query_id="..." /> blocks and SQL tables via <table query_id="..." /> blocks. Use <grid cols="2">...</grid> to lay out charts side by side in a responsive grid.',
+			`The markdown content. Required for "create" (initial content) and "replace" (new content). Can include charts via <chart query_id="..." /> blocks and SQL tables via <table query_id="..." /> blocks. Use <grid cols="2">...</grid> to lay out charts side by side in a responsive grid. Default to a single flowing story. Use <tabs> only when the user asks for tabs, or when the content splits into clearly distinct sections that are better separated than stacked (e.g. overview vs. detail, one topic/department/metric per tab). Avoid tabs for a short or single-topic story. Always follow the user's explicit request (e.g. "a tab per chart" → one chart per tab). When using tabs, the story must start with <tabs> and contain only <tab title="…">…</tab> blocks — no text outside a tab.
```
Bl3f: Not sure we need to details or copy the same text as the description of the tools, we can mention the tabs feature but it does not require to repeat the same desc as we define in apps/backend/src/agents/tools/story.ts

apps/shared/src/story-segments.ts
```diff
@@ -198,7 +199,7 @@ function extractSeriesFromRawAttrs(attrString: string): ParsedChartBlock['series
 export function splitCodeIntoSegments(code: string): Segment[] {
 	const segments: Segment[] = [];
 	const blockRegex = new RegExp(
-		`<grid\\s+([^>]*)>([\\s\\S]*?)<\\/grid>|<chart\\s+(${TAG_ATTRS})\\/?>|<table\\s+(${TAG_ATTRS})\\/?>`,
+		`<grid\\s+([^>]*)>([\\s\\S]*?)<\\/grid>|<chart\\s+(${TAG_ATTRS})\\/?>|<table\\s+(${TAG_ATTRS})\\/?>|<tabs\\b[^>]*>([\\s\\S]*?)(?:<\\/tabs>|$)`,
```
Bl3f: why is the closing </tabs> in a capturing group?

Bl3f (changes requested):
I think there are 2 bugs that should be fixed in the workflow (and rather than fixed it's more than your PR added this kind of wrong behaviour):
- when you edit a story (there is already a story and you asked for a change) to add a tab it empties the story without any warning which is a bit weird
- add the end of the edition (when the streaming restart on the chat) the story falls back to the previous one and when the stream ends on the chat the story finally ends on the edited story

ad4mou:
You're right there were two weird bugs in the preview when adding tabs. One where preview went blank and another where we would come back to old version for a moment before showing the final version with tabs added.
Both are fixed now

apps/backend/src/agents/tools/story.ts
ad4mou: You're right
At first I wanted to do it this way because I wanted no content outside tabs since that would mean lower level tabs and we didn't want that. I also wanted it to be efficient for the UI to know if we're using tabs or not.

But actually even so we don't need this <tabs> at the beginning, we can just grep "<tab" to know if tabs are enabled and declare a compilation error if there is content outside tabs when they're enabled.
So I'll remove the `<tabs>` at the beginning

Still for token cost it's still as costly and that's because need to ensure there is no content outside tabs blocks when tabs are enabled. So we need to rewrite the full story it seems.

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
ad4mou: ok that's done

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
ad4mou: CleanShot 2026-07-20 at 14 59 48@2x

So I tried to make a dialog similar to live story's but got stuck because there was built-in dialog code for deletion buttons which was overriding my round button. It seems deletion buttons should be rectangles.
So is this ok for the style or I need round buttons?

apps/shared/src/tools/story.ts
ad4mou: you're right, I changed that

apps/shared/src/story-segments.ts
ad4mou: Since I removed the whole <tabs> thing, this shouldn't be a problem now

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
Bl3f: I prefer we stick to the round button and keep the same style. It's weird, you should be able to do whatever you want.

ad4mou: okay okay I'll do that

apps/frontend/src/components/side-panel/hooks/use-story-viewer-content.ts
```diff
@@ -21,9 +22,32 @@ export const useStoryViewerContent = ({
...
+			return;
+		}
+		const committedCaughtUp = Boolean(currentVersion && draftStory && currentVersion.code === draftStory.code);
+		if (!draftStory || committedCaughtUp) {
+			setIsBridgingDraft(false);
+		}
+	}, [isBridgingDraft, draftStory, currentVersion]);
+
+	const shouldUseDraftStory = Boolean(
+		draftStory && (draftStory.isStreaming || isAgentRunning || isBridgingDraft || !currentVersion),
```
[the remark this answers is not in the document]
ad4mou: already fixed

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
```diff
@@ -0,0 +1,218 @@
...
+										editable ? 'pr-1' : 'pr-3',
+									)}
+								>
+									{tab.title || 'Untitled'}
+								</button>
+							)}
+
+							{editable && editingIndex !== index && (
+								<div className='mr-1 flex items-center opacity-0 transition-opacity group-hover/tab:opacity-100 group-focus-within/tab:opacity-100'>
+									<Dialog>
```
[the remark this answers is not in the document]
ad4mou: Well I based myself on the delete dialog when removing a project member and I observe the exact same behavior of: "escape, outside-click, and the built-in close button can silently dismiss the prompt"
so I'll assume we're good

apps/frontend/src/components/side-panel/story-tabs-bar.tsx
```diff
@@ -0,0 +1,218 @@
...
+												aria-label={`Delete ${tab.title || 'tab'}`}
+											>
+												<X />
+											</Button>
+										</DialogTrigger>
+										<DialogContent>
+											<DialogHeader>
+												<DialogTitle>Delete this tab?</DialogTitle>
+											</DialogHeader>
+											<p className='text-sm text-muted-foreground'>
```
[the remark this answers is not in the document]
ad4mou: just not true, a description does show

apps/frontend/src/components/side-panel/hooks/use-story-viewer-content.ts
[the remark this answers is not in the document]
ad4mou: fair enough, fixed now

Bl3f (approved):
LGTM

Issue #869, opened 2026-05-29
[bug] Charts don't render in story view mode until user enters edit mode and saves


ClaireGz:
## Problem

Charts and tables in stories sometimes fail to render in view mode, showing "Chart data unavailable" / "Table data unavailable" placeholders instead of the actual visualizations.

**Steps to reproduce:**
1. Open a story that contains charts/tables
2. Ask and edit to a table/chart in the story (ex: change the X axis granularity, refresh data)
3. this will show chart data unavailable

Bl3f:
Looks like it's still here

PR #1217, opened 2026-07-20
[fix: chart/table data] resolve data unavailable when chart/table changes on live stories

socallmebertille:
## Summary

Live stories showed "chart/table data unavailable" after editing a chart, because the story kept serving its old cached data which no longer matched the edited chart.

=> This fix fills in the missing data from the chat's existing query results, so edited charts and tables render right away in view mode without needing to refresh or re-save.

## Related issue

#869 

<a href="https://cubic.dev/pr/getnao/nao/pull/1217?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/services/live-story.ts
+++ b/apps/backend/src/services/live-story.ts
@@ -5,6 +5,7 @@ import { z } from 'zod';
 
 import { llmTelemetry } from '../agents/telemetry';
 import { LiveStoryRefreshPrompt } from '../components/ai/live-story-refresh-prompt';
+import type { DBStoryDataCache } from '../db/abstractSchema';
 import { env } from '../env';
 import { renderToMarkdown } from '../lib/markdown';
 import * as chatQueries from '../queries/chat.queries';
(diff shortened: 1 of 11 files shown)


apps/backend/src/services/live-story.ts
```diff
@@ -116,12 +117,54 @@ export async function getStoryQueryData(
 		};
 	} catch {
 		if (cache) {
-			return { queryData: cache.queryData, cachedAt: cache.cachedAt };
+			const queryData = await backfillQueryDataFromChat(chatId, code, cache.queryData);
```
Bl3f: I feel the behaviour is a bit weird algorithmically: if cache then you backfill, but in the backfill method depending on some ifs you return the cache which means it does not backfill entirely

apps/backend/src/services/live-story.ts
```diff
@@ -116,12 +117,54 @@ export async function getStoryQueryData(
...
+	for (const id of missingIds) {
+		if (fromChat[id]) {
+			merged[id] = fromChat[id];
+		}
+	}
+
+	return Object.keys(merged).length > 0 ? merged : cached;
+}
+
+function extractStoryQueryIds(code: string): Set<string> {
```
Bl3f: Don't we already have some methods to extract ids?

Bl3f (commented):
Some comments, I'm not sure about the naming and the backfill function

Bl3f:
LGTM

PR #1237, opened 2026-07-22
fix: The add to story button now support tabs

ad4mou:
## Summary
- Fix the "Add to story" button on in-chat charts and tables for tabbed stories. It now adds the chart to the end of the tab you're currently viewing, instead of dropping it after the last tab which wouldn't compile.
- Stories without tabs are unchanged.
- If the story isn't open on screen, the chart goes to the last tab and the story opens on that tab so you see it right away.
- Moved the insertion logic into one shared helper (`appendBlockToStoryCode`) used by both charts and tables, with unit tests for every case.

<a href="https://cubic.dev/pr/getnao/nao/pull/1237?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f (approved):
LGTM

PR #1240, opened 2026-07-22
fix: total no longer shows up when series are unrelated, or can't be summed for any reason

### LINKED ISSUE #1116: [bug] Line charts shows grand total on all series
## Problem
On several series chart, the agent still computes and shows **total** while those series are maybe not the same units.

Ex:

Image

## Expected behavior
On line charts with several series, do not show a grand total
Only show a grand total if it's one serie with several dimensions

## Customer ID

ad4mou:
## Summary
- Add a chart-wide `hide_total` option so the assistant can turn off the tooltip's "Total" row when a chart's series can't be meaningfully added up (unrelated metrics, different units or currencies). Additive charts still show the total; percentage charts still show 100%.
- The flag is part of the schema the model actually sees, so the agent can set it itself
- The per-series `is_total` behavior is unchanged

Closes #1116

<a href="https://cubic.dev/pr/getnao/nao/pull/1240?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/utils/story-html.tsx
+++ b/apps/backend/src/utils/story-html.tsx
@@ -165,6 +165,7 @@ function ChartBlock({ chart, queryData }: { chart: ParsedChartBlock; queryData: 
 			chartType: chart.chartType,
 			yAxisMin: chart.yAxisMin,
 			yAxisMax: chart.yAxisMax,
+			hideTotal: chart.hideTotal,
 		});
 		return (
 			<div style={{ margin: '16px 0' }}>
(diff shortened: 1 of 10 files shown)


Bl3f (approved):
LGTM

PR #1247, opened 2026-07-22
Add value formatting to charts ($, %, units)

### LINKED ISSUE #1115: [bug] Improve value formatting in charts ($/%)
## Problem
Numeric values in charts and tables are not formatted with their unit:
- Percentage metrics are not shown with a `%` sign.
- Metrics in USD are not shown with a `$` (currency symbol missing).
- The **cost charts in the admin Usage section have no cost unit** (currency) on the values/axis.

Ex:

Not formatted currencies
Image

Not formatted percentages

Image

## Expected behavior
Chart and dashboard values should render with the right unit/format: `%` for percentages, the currency symbol for monetary metrics (e.g. `$` for USD), and a currency unit on the usage cost charts. Ideally the format follows the metric/column type.

## Customer ID

ad4mou:
## Summary
- Chart values can now show a currency symbol, `%`, or any unit (e.g. V, kg, GB) instead of bare numbers.
- Format is per series: a number format (d3-format) plus a unit placed before or after, so any currency or unit works, not a fixed list.
- Big numbers shorten sensibly: money/counts as K/M/B, scientific units (e.g. bytes) keep k/M/G.
- The admin Usage cost charts now show `$`.
- The chart editor lets you set the number format and a unit (with a Prefix/Suffix choice, clearable fields, and a link to the syntax docs).
- The value axis widens to fit long labels (e.g. `$1,000,000`) so they no longer clip.
- Formatting carries through into stories, including exported HTML (labels, KPI cards, tooltips).

Closes #1115

<a href="https://cubic.dev/pr/getnao/nao/pull/1247?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/frontend/src/components/tool-calls/display-chart-edit-dialog.tsx
```diff
@@ -284,42 +315,115 @@ export function ChartConfigEditDialog({
...
-							))}
+							{draft.series.map((series, index) => {
+								const placement: UnitPlacement = series.value_format?.prefix ? 'prefix' : 'suffix';
+								const unit =
+									placement === 'prefix'
+										? (series.value_format?.prefix ?? '')
+										: (series.value_format?.suffix ?? '');
+
+								return (
+									<div key={index} className='flex flex-col gap-2 rounded-md'>
```
Bl3f: Here let's change the display, hide the row by default and only show it when clicked on a "format" icon that would be on the main row of the serie

Bl3f (changes requested):
small change

apps/frontend/src/components/tool-calls/display-chart-edit-dialog.tsx
ad4mou: Yes that's cleaner, Ok will do!

ad4mou:
LGTM, last commit was just frontend changes to improve the UX as requested and to make sure the new dual-axis charts also support value formatting

Issue #1268, opened 2026-07-24
[bug : tables] column sorting don't conserve max width


socallmebertille:
## Problem

- Columns don't reserve a stable width, so sorting resizes each column to fit only the rows currently in view — column widths jump around, especially with long values
- Because cells reflow on every sort, the conditional-formatting color scale appears to flicker/shift (the computed colors are actually unchanged — it's a visual side effect of the width jump)

## Expected

Column widths stay stable across sorting and paging; the color scale no longer appears to jump

PR #1269, opened 2026-07-24
[fix : tables] column sorting conserve max width

socallmebertille:
## Summary

- `display-table.tsx` : compute a per-column min-width from the longest formatted value across the full dataset, so column widths stay stable regardless of sort/page
- ignore the active sort when its column is no longer present in the data
The color-scale "flicker" was a visual side effect of the width jump, not a color-scale bug — cell colors depend only on value + full-column range, independent of row order

## Related issue

#1268 

<a href="https://cubic.dev/pr/getnao/nao/pull/1269?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/tool-calls/display-table.tsx
+++ b/apps/frontend/src/components/tool-calls/display-table.tsx
@@ -25,6 +25,8 @@ interface TableDisplayProps {
 	humanizeColumnLabels?: boolean;
 }
 
+type Sort = { column: string; direction: SortDirection };
+
 export function TableDisplay({
 	data,
 	columns,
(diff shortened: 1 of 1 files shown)


Bl3f:
LGTM

PR #1272, opened 2026-07-24
[fix : feed] resolve story and feed access from feed panel

### LINKED ISSUE #1263: [bug] Unshared stories's refresh events leak into feed; opening them shows blank error page
## Problem

Users (and admins) see refresh events for stories that were **not shared with them** in their feed. When they click "Open story" on one of these, it loads a blank white page with a "something went wrong" error - because they don't actually have access to that story.

## Expected behavior

- Story refresh events should only appear in a user's feed for stories they have access to (shared with them or owned by them).
- Stories that were never shared with a user must not surface in that user's feed.
- If a user does reach a story they can't access, show a proper "no access" state rather than a blank error page.

## Customer ID
776fa4a9-7636-4659-9f45-56b73021277c

socallmebertille:
## Summary

- resolve activity-feed story links per viewer access — owners open the preview/standalone route while non-owners are routed to the shared-story link, instead of always assuming the preview route
- resolve the presence of feed events for users who have no permissions (eq. issue from user who saw refresh stories that they don't have access to)
- add a StoryAccessError screen :
image

## Closes

#1263 

<a href="https://cubic.dev/pr/getnao/nao/pull/1272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/queries/activity.queries.ts
+++ b/apps/backend/src/queries/activity.queries.ts
@@ -1,4 +1,4 @@
-import { and, desc, eq, inArray, lte, ne, or, sql } from 'drizzle-orm';
+import { aliasedTable, and, desc, eq, inArray, lte, ne, or, sql } from 'drizzle-orm';
 
 import s, {
 	type ActivityTrigger,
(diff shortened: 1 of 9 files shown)


Bl3f (approved):
LGTM

PR #1278, opened 2026-07-25
feat(cli): export Notion databases and embedded views

### LINKED ISSUE #1277: [feature] nao sync: export Notion databases, including the ones embedded in a page
## Problem

`nao sync` exports Notion pages as Markdown but silently drops database content.

When a page contains an embedded database, `notion2md` has no converter for `child_database`. The page still syncs successfully, with only this comment where the database should be:

```markdown
[//]: # (child_database is not supported)
```

The resulting file looks complete, so the agent has no indication that useful context is missing.

Configuring a database URL directly in `notion.pages` does not work either: nao treats its ID as a page ID, and `pages.retrieve` returns a `validation_error`.

## Expected behavior

- A database configured directly in `notion.pages` is exported as a Markdown table.
- If its URL contains `?v=<view_id>`, the export follows that view's filters, sorts, row order, and visible columns.
- A database embedded in a page is rendered in place instead of being silently omitted.
- Large results are truncated with an explicit note rather than silently cut off.

Respecting the view matters for embedded databases. The same data source can be referenced from several pages with different filters; exporting the entire data source would add unrelated rows and hidden columns
to each page.

## Proposed implementation

The Views API, available with Notion API version `2025-09-03` and later, makes this possible. `POST /v1/views/{view_id}/queries` executes a view's saved filters and sorts, while the view configuration describes
its visible columns. These endpoints are exposed by `notion-client` 3.1.

One limitation remains: a `child_database` block exposes its database ID, but Notion does not expose which view is active in that embed. For embedded databases, the PR therefore uses the first view returned for
that database. For directly configured database URLs, an explicit `?v=<view_id>` is honored; without one, the database is exported without view-specific filtering.

If a database cannot be exported safely, the page fails to sync and the previous Markdown file is preserved instead of being overwritten without its table.

This is a need on our side, but it seems generally useful, so I’ve opened a PR alongside this issue. The implementation is intended as a proposal, and I’m happy to discuss the behavior and trade-offs.

tckaros:
Closes #1277

This implements the proposal described in the issue above. I'm happy to adjust the behavior or scope.

`nao sync` now exports Notion databases as Markdown tables, both when configured directly in `notion.pages` and when embedded in a page. Embedded databases were previously replaced with a `child_database is not supported` comment, leaving an apparently complete file without the table content.

- A database configured by URL is exported without view-specific filtering unless its URL contains `?v=<view_id>`, in which case that view's filters, sorts, row order, and visible columns are applied.
- An embedded database uses the first view listed for its database block because Notion does not expose which view is active in the embed. Applying that view prevents a shared data source filtered differently on each page from pulling unrelated rows or hidden columns into every file.
- Tables render at most 500 rows and explicitly report truncation or incomplete results.

Rows are resolved through batched view and data-source queries rather than one request per row. This requires the Views API, so the Notion integration now uses API version `2025-09-03` and pins `notion-client>=3.1.0,<4`.

If a database cannot be exported safely, only the affected item fails. Its previously synced Markdown is preserved, and stale cleanup is skipped after a partial sync rather than overwriting or deleting valid content.

## Testing

- The repository pre-commit checks pass, including TypeScript lint, formatting, migration checks, and CLI lint.
- Tested against a real Notion workspace with a page containing a filtered embedded database. The export contains the 27 rows and visible columns shown by the view, without its hidden columns.

<a href="https://cubic.dev/pr/getnao/nao/pull/1278?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -165,7 +165,7 @@ Syncs configured resources to local files:
 
 - **Databases** - generates markdown docs (`columns.md` with table schema, description, row count, and partition/clustering/index metadata, `query_history.md`, and `preview.md`) for each table into `databases/`
 - **Git repositories** — clones or pulls repos into `repos/`
-- **Notion pages** — exports pages as markdown into `docs/notion/`
+- **Notion pages** — exports pages as markdown into `docs/notion/`. Databases are exported as markdown tables, whether configured directly or embedded inline in a page. A database embedded in a page is exported through one of its views — Notion exposes no way to tell which view a page renders, so the first one listed is used — applying that view's filters, sorts and visible columns rather than dumping the whole data source. A database configured by URL exports every row and column, unless the URL carries `?v=<view_id>`, in which case that view applies. When a database cannot be exported, its page fails to sync and the previously synced markdown is left untouched, rather than being rewritten without its table.
 
 After syncing, any Jinja templates (`*.j2` files) in the project directory are rendered with the nao context.
 
--- a/cli/nao_core/commands/sync/providers/notion/database.py
+++ b/cli/nao_core/commands/sync/providers/notion/database.py
@@ -1,0 +1,702 @@
+"""Exporting Notion databases as markdown tables.
+
+Every page walks its own block tree to find the databases it holds, which duplicates the
+traversal notion2md performs internally and is the dominant request cost on large pages. It
+is deliberate: notion2md silently drops blocks it cannot convert, so its markers alone cannot
+tell an absent database from a lost one.
+"""
+
+import re
+from dataclasses import dataclass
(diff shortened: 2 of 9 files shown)


Bl3f (approved):
ty @tckaros ; LGTM, merging and releasing this today

PR #1315, opened 2026-07-30
[fix : mcp nao] ask_nao clarification questions instead of empty answers

### LINKED ISSUE #875: MCP ask_nao returns empty text when agent asks interactive/clarifying questions
We're integrating NAO as a backend analytics agent via the MCP endpoint (ask_nao tool). The integration works well for direct answers, but when NAO decides to ask a clarifying question (e.g., "What time range should I use for headcount?"), the MCP response returns:

structuredContent.text = "" (empty string)
content[0].text = only [chatId: ...]\n[chatUrl: ...] metadata
The interactive question (with options like "Today", "End of last month", "Last 12 months average") is rendered only in NAO's own chat UI — it's not exposed through the MCP protocol.

Expected behavior: When NAO asks a clarifying question, the MCP response should include the question text (and ideally the options) in structuredContent.text so external integrations can forward it to their own UI.

Current workaround: Added a rule in RULES.md: "NEVER ask the user for clarification. ALWAYS provide an answer using sensible defaults." — but this removes a useful UX feature.

Environment:

NAO self-hosted (Docker, getnao/nao:latest)
MCP Sub-agent mode enabled
Calling via JSON-RPC 2.0 Streamable HTTP
Question: Is there a way to receive interactive questions/options via MCP? Or is this a planned feature?

socallmebertille:
## Summary

`ask_nao` returned an empty answer when nao asked a clarifying question (it lives only in a `tool-clarification` UI part). Harden clarification extraction and fall back to the persisted assistant message so MCP clients reliably receive `status: 'needs_clarification'` with the question and options.

Fix #875 

<a href="https://cubic.dev/pr/getnao/nao/pull/1315?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/mcp/tools/sub-agent.ts
+++ b/apps/backend/src/mcp/tools/sub-agent.ts
@@ -1,5 +1,5 @@
 import type { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js';
-import { type InferUIMessageChunk, readUIMessageStream } from 'ai';
+import { getToolName, type InferUIMessageChunk, isToolUIPart, readUIMessageStream } from 'ai';
 import { z } from 'zod';
 
 import { WEB_CHAT_ONLY_TOOLS } from '../../agents/tools';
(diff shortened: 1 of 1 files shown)


Bl3f (approved):
LGTM

PR #1316, opened 2026-07-30
Fix MCP runtime creation race causing "Unknown MCP server" errors

### LINKED ISSUE #1292: MCP servers randomly fail with "Unknown MCP server" after startup or config reload
Summary

Concurrent MCP server discovery (startup or config reload) races on runtime creation: each server's _ensureRegistered sees `this._runtime == null`, builds its own runtime, and overwrites the previous one. Servers registered on an overwritten instance are lost, and all later refreshes and tool calls for them fail instantly with `Unknown MCP server '<name>'`.

Environment

- getnao/nao:0.3.0 (Docker) — also reproduced after rebuild; the same code is in current main (apps/backend/src/services/mcp.ts)
- 3 stdio servers in agent/mcps/mcp.json (affects any config with ≥2 non-OAuth servers; transport-independent — one of ours is a plain stdio server, two use mcp-remote)

Reproduction

1. Configure 2+ servers in agent/mcps/mcp.json.
2. Start the container with no spec dirs under agent/mcps/ (forces parallel discovery), or touch mcp.json to trigger _reloadAndDiscover.
3. Click refresh on a server in the MCP settings panel, or let the agent call one of its tools.
4. Some subset of servers fails with Unknown MCP server '<name>' in ~30 ms. Which server fails varies per run (whoever's createRuntime() resolves last "wins").

Log signature: `[ERROR] [tool] MCP discovery failed: <name> / MCP tool call failed: <name>/<tool> with the error Unknown MCP server '<name>'.`

Root cause

_ensureRegistered has a check-then-act race:
```
private async _ensureRegistered(name: string): Promise<void> {
    if (!this._runtime) {
        this._runtime = await createRuntime();   // N concurrent callers → N runtimes, last write wins
    }
    if (this._registered.has(name)) {
        return;                                  // sta a discarded runtime
    }
    ...
}
```
All parallel discoveries pass the !this._runtime check before any assignment resolves. Each registers its server on whichever instance is current at its own resolution time; _registered accumulates all names, so subsequent calls skip re-registration and hit the surviving runtime, which never saw the definition.

Related: this code path replaced the per-call runtime from #202 (fixed in #1049) — same symptom, new mechanism.

Suggested fix

Memoize the in-flight createRuntime() so concurrent callers share one runtime
```
private _runtimePromise: Promise<Runtime> | null = null;

private async _ensureRegistered(name: string): Promise<void> {
    if (!this._runtime) {
        if (!this._runtimePromise) {
            this._runtimePromise = createRuntime().catc
                this._runtimePromise = null;
                throw error;
            });
        }
        this._runtime = await this._runtimePromise;
    }
    ...
}
```

Verified in our deployment: with this change, parallel rediscovery of all 3 servers completes cleanly and tool calls succeed; unpatched, at least one server fails every reload. Happy to send this as a PR.

beherap:
Fixes #1292

When several MCP servers are discovered in parallel (startup or config reload), every `_ensureRegistered` call saw `_runtime == null`, built its own runtime, and overwrote the previous one. Servers registered on an overwritten instance were lost, so refreshes and tool calls failed with `Unknown MCP server '<name>'`.

This memoizes the in-flight `createRuntime()` promise so concurrent callers share a single runtime. The promise is cleared in `_resetRuntime()` and on creation failure. ~12 line diff, no behaviour change outside the race.

Tested against our real deployment (3 stdio servers, one plain stdio + two via mcp-remote): before the fix a random subset of servers failed after every startup/reload; with this patch applied all servers register and tool calls succeed consistently.

This PR was written using Claude Fable 5.

<a href="https://cubic.dev/pr/getnao/nao/pull/1316?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/services/mcp.ts
+++ b/apps/backend/src/services/mcp.ts
@@ -24,6 +24,9 @@ const GITIGNORE_CONTENT = '# nao: discovered MCP tool specs (generated at runtim
 
 type DisabledSets = { servers: Set<string>; tools: Set<string> };
 
+/** A runtime paired with the servers registered on it, so the two can never describe different instances. */
+type RuntimeSession = { runtime: Runtime; registered: Set<string> };
+
 /** Thrown when `mcp_call` arguments do not match the target tool's discovered input schema. */
 export class McpArgsValidationError extends Error {
 	constructor(
@@ -68,10 +71,8 @@ export class McpService {
 	private _projectPath = '';
 	private _mcpJsonFilePath = '';
 	private _mcpServers: Record<string, McpServerConfig> = {};
-	private _runtime: Runtime | null = null;
-	/** In-flight createRuntime() shared by concurrent callers so they don't each build a runtime. */
-	private _runtimePromise: Promise<Runtime> | null = null;
-	private _registered = new Set<string>();
+	/** In-flight session shared by concurrent callers so they don't each build a runtime. */
+	private _runtimePromise: Promise<RuntimeSession> | null = null;
 	/** Full tool list per server from the last successful discovery this session. */
 	private _discovered: Record<string, McpToolDefinition[]> = {};
 	/** Cached per-server flag for whether the server is OAuth-protected. */
@@ -242,11 +243,8 @@ export class McpService {
 	}
 
 	private async _callToolMcporter(server: string, tool: string, args: Record<string, unknown>): Promise<unknown> {
-		await this._ensureRegistered(server);
-		if (!this._runtime) {
-			throw new Error('MCP runtime not initialized');
-		}
-		return this._runtime.callTool(server, tool, { args });
+		const runtime = await this._ensureRegistered(server);
+		return runtime.callTool(server, tool, { args });
 	}
 
 	private async _callToolOAuth(opts: {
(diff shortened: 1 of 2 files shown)


Bl3f:
Hey @beherap ty so much for your PR. Just to be sure on what nao version are you?

beherap:
> Hey @beherap ty so much for your PR. Just to be sure on what nao version are you?

Hi @Bl3f, we are on v0.3.0

Bl3f:
Super. @beherap I have changed a bit your PR instead of doing a review because I was a bit stuck on it and wanted to understand the issue.

I wasn't fond of the `while()` structure, i'd prefer avoid this at server startup like this. So I did another thing that also fixes it.

I'll merge now! Congrats for this first PR :)

Bl3f (approved):

beherap:
> Super. @beherap I have changed a bit your PR instead of doing a review because I was a bit stuck on it and wanted to understand the issue.
> 
> I wasn't fond of the `while()` structure, i'd prefer avoid this at server startup like this. So I did another thing that also fixes it.
> 
> I'll merge now! Congrats for this first PR :)

Cool! Thank you for the quick turn around, @Bl3f !

PR #1320, opened 2026-07-31
Exclude sql calls from Langfuse tracing

ad4mou:
## Summary

- Query data never reaches Langfuse anymore. Anything belonging to `execute_sql` and
  `read_query_result` (the SQL text, the returned rows, and error messages) is replaced
  with `[redacted]` before traces are sent.
- The calls stay visible in the trace: you still see that a query ran, how long it took and
  whether it failed. Only the contents are gone.
- Redaction covers both places the data showed up: the tool call itself, and the
  conversation history that gets replayed on every following model call. Redacting only the
  tool call would have left the SQL and rows in every later prompt.
- Everything else is unchanged - questions, answers, other tools, timings, token usage.
- Always on, no setting to turn it off.

<a href="https://cubic.dev/pr/getnao/nao/pull/1320?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/instrumentation.ts
+++ b/apps/backend/src/instrumentation.ts
@@ -2,6 +2,7 @@ import { LangfuseSpanProcessor } from '@langfuse/otel';
 import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
 
 import { env } from './env';
+import { DataToolRedactingSpanProcessor } from './utils/langfuse-redaction';
 
 /**
  * Langfuse tracing is opt-in and requires an explicit destination: it only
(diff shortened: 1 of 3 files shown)


ad4mou:
LGTM, maybe test is too long though

Bl3f (approved):
LGTM

PR #1326, opened 2026-07-31
Keep the conversation feedback prompt hidden after typing and back off to 20s

Bl3f:
## Problem

The "How did this conversation go?" prompt above the chat input only collapsed visually while the user was typing. As soon as the input was cleared (or the draft deleted), the prompt popped straight back, because the 10s inactivity countdown had already fired and stayed fired.

## Change

Typing now counts as activity for the feedback prompt instead of only hiding it:

- While the user has text in the input, the countdown is paused and the prompt is off — not just visually collapsed.
- Clearing the input restarts the countdown from zero, so the prompt does not reappear instantly.
- If the prompt was already showing when the user started typing, it is treated as ignored and the next appearance waits 20s (`FEEDBACK_IGNORED_INACTIVITY_MS`) instead of 10s. The back-off resets when the user switches chat.
- If the user typed before the prompt ever appeared, nothing is treated as ignored and the delay stays at 10s.

`useInactivityTrigger` gains an optional `isPaused` flag and an `onIgnored` callback, so the timing mechanics stay in the hook while the back-off policy stays with the caller. The `ChatInputSuggestions` prop was renamed from `isHidden` to `isUserTyping` to reflect what it actually carries.

## Verification

The behaviour was validated during development with temporary vitest tests (prompt appears after 10s, stays hidden through typing and clearing, returns only at 20s, and keeps the 10s delay when typing happened before it appeared). Those tests also failed against the previous implementation, confirming they captured the bug. They have been removed from the branch at the maintainer's request.

`npm run lint`, `npm run format:check`, and the frontend suite (96 tests) pass on the final state of the branch.

<div><a href="https://cursor.com/agents/bc-d2f95157-abd0-4339-ab30-a6aa4c01ff77?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d2f95157-abd0-4339-ab30-a6aa4c01ff77&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

<a href="https://cubic.dev/pr/getnao/nao/pull/1326?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f:
Will do this differently, I don't like this solution.

PR #1349, opened 2026-08-04
[feat : recommendations improvements] UI + relevance

### LINKED ISSUE #982: [feature] Context recommendations: relevance, root-cause explanation, nav UI, and better fixes
## Problem
Several improvements to the context recommendations feature, gathered from a pilot review. Today recos are noisy, hard to trace, hard to triage, and over-default to one type of fix.

### 1. Relevance of recommendations
- **Deduplicate / group same-type recos.** If the agent hallucinates column names across 3 tables, don't emit 3 recos - emit one, grouped by shared root cause (e.g. "agent didn't check the table schema") and shared recommendation (e.g. "improve rules so it fetches the right schema").

### 2. Explain the origin of the reco
- **Explain the exact sequence that happened**: what the agent read / didn't read, and what mistake followed. Example: "the agent did not read `<table>/columns.md`, then wrote an SQL query with a hallucinated column."
- **For `Tool errors` recos, add a metric**: how many tool calls failed because of this root cause.

### 3. Recommendation navigation UI
- **Categorize recommendations** into buckets - `Tool errors`, `Hallucinations`,`Semantic missing`, `Other` - so users can triage by type.
- **Link to the exact spot in the chat** that triggered the reco. If it's a failed tool call, open directly on that tool call with the error content expanded. If it's a wrong feedback, go to the message who triggered it.  etc.
- **Chat list under each reco as a table with user name + title, not just chat ID.** Observed admin behavior: they want to know *who* ran the chat, because those people already raised the issue with them directly.

### 4. Better fix suggestions
- **Explain different root causes for errors to the agent**. The agent always suppose it's a lack of context. But a root cause for an hallucination could be: context doesn't exist OR context is wrong OR context exists but the agent did not retrieve correctly the context. The agent need to determine which case it is and suggest fixes accordingly.
- The reco agent **over-defaults to "add to rules.md"**. It should propose the appropriate fix type instead:
  - fix a data model
  - create a separate doc
  - create a skill
  - add a metric in the semantic layer
- Give the agent a clear distinction:
  - **docs = content/description** (e.g. "how CRM data is produced")
  - **skills = analysis-process description** (e.g. "how we usually analyze CRM - which filters etc." -> a `crm analysis` skill)

## Customer ID

### LINKED ISSUE #983: [feature] Link user feedbacks to context recommendations
## Problem
Some users observed that they got feedbacks on chats not tackled by context recommendations.

## Expected behavior
- **Every negative feedback should ALWAYS generate at least one recommendation.**
- In the context recommendations panel, it should be **clear that a reco originates from a feedback** (visible link/label a,d what the feedback was by who).
- On the chat replay / feedbacks panel, there should be a **link to the suggested recommendation** for that feedback.

This makes feedbacks and recos two views of the same loop: a feedback points to its reco, and a reco points back to the feedback that motivated it.

## Customer ID

socallmebertille:
## Summary

- [x] UI navigation improvements (4 tabs/window to smooth the navigation, filter/sort system on reco, only 2 actions on each, easier understanding with the new UI)
- [x] 4 new categories for reco 
- [x] Title & description improvements (clearer and shorter)
- [x] Easiest navigation to linked chat/feedback (add the username and title from chat who trigger reco)

## Overview

image

## Related issues

Fix #982 
Fix #983 

<a href="https://cubic.dev/pr/getnao/nao/pull/1349?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/src/db/sqlite-schema.ts
```diff
@@ -769,6 +773,22 @@ export const contextRecommendation = sqliteTable(
 	],
 );
 
+export const recommendationFeedback = sqliteTable(
+	'recommendation_feedback',
```
Bl3f:
```suggestion
	'context_recommendation_linked_feedback',
```

Bl3f (approved):
LGTM small comment

PR #1364, opened 2026-08-06
Fixed chat stream lags and bugs

### LINKED ISSUE #1355: Long streams in chat lags a lot
When the agent is streaming large responses, UI lags a lot, scroll doesn't work, text disappears.
Once stream is done, text looks fine.

ad4mou:
## Summary

- Long agent responses no longer lag or freeze the chat. The page stayed
  responsive while typing, scrolling was fine, but the answer itself would stall
  and sometimes crash the tab.
- Two causes. Streamed markdown was re-parsed from the top on every update, so
  cost grew with the length of the answer. And every update re-rendered every
  chart, table and map already on screen, even finished ones.
- Streamed text is now split into finished blocks and a live tail, and only the
  tail is re-parsed. Finished tool calls are compared by content instead of by
  object identity, so they render once and stay put.
- Result on a long response: about 6ms per update instead of 27ms, worst pause
  16ms instead of 222ms.
- Also caches the repeated scans over the message list (story ids, source
  queries, chart data) so they run once per update rather than once per
  component.

Fix #1355

<a href="https://cubic.dev/pr/getnao/nao/pull/1364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/shared/src/tools/execute-sql.ts
ad4mou, line 1: Finished tool calls were re-rendering on every streaming tick, because the AI SDK
allocates a new `output` object each update even when nothing changed. Deep-comparing
outputs is too expensive (query results can hold thousands of rows), so settled tool
calls now skip the output comparison entirely, assuming a finished result never changes.

That assumption breaks in one case: re-running a query from the SQL editor without
editing the SQL. Same id, same input, same everything we compare so the chart kept
showing stale rows.

`revision` is a counter stamped on the result each time it's replaced in place. One
number to compare instead of thousands of rows.

apps/frontend/src/lib/incremental-markdown-blocks.ts
ad4mou, line 1: Serves to programmatically split the markdown into blocks so that we give only the current block to streamdown and avoid having to re-render everything

apps/frontend/src/hooks/use-agent.ts
```diff
@@ -229,7 +232,12 @@ export const useAgent = ({ disableNavigation = false }: { disableNavigation?: bo
 
 	agentSendRefsStore.set(agentInstance, { adminModeRef, selectedModelRef, mentionsRef });
 
-	const { status, error, clearError, sendMessage, setMessages, messages } = useChat({ chat: agentInstance });
+	const { status, error, clearError, sendMessage, setMessages, messages } = useChat({
+		chat: agentInstance,
+		experimental_throttle: 8,
```
ad4mou, line 237: Key AI SDK parameter, caps UI updates at one per 8ms

apps/frontend/src/components/data-table-card.tsx
ad4mou, line 1: just wrapped everything into memo()

apps/frontend/src/contexts/agent.provider.tsx
ad4mou, line 1: This is the key logic

It kinds of reminds me of the ARP protocol where you broadcast requests to every device and only the one concerned answers.

Here each component keeps track of what states matter for it and gets rendered again only if some state it cares about has changed

ad4mou:
LGTM

apps/frontend/src/components/tool-calls/mcp.tsx
```diff
@@ -9,12 +9,13 @@ import type { ToolCallComponentProps } from '.';
...
-import { useOptionalAgentContext } from '@/contexts/agent.provider';
+import { useAgentMessagesSelector } from '@/contexts/agent.provider';
 import { useToolCallContext } from '@/contexts/tool-call';
 
-const EMPTY_MESSAGES: UIMessage[] = [];
-
 type McpContent = { type: string; text: string };
+type SqlData = Record<string, unknown>[];
+
+const sqlDataByMessages = new WeakMap<UIMessage[], Map<string, SqlData | null>>();
```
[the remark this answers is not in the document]
ad4mou: Not necessary, my tests show that the current cache works great. and building your proposed index would lead to other problems

Bl3f:
There is a git conflict on uv.lock ; i think you should keep the one with 0001 date, but not sure

Bl3f:
LGTM

PR #1368, opened 2026-08-08
feat(cli): Add pass@k and pass^k metrics to nao test KPIs

### LINKED ISSUE #1248: [feature] Add pass@k and pass^k metrics to nao test KPIs
## Problem

`nao test` currently reports accuracy as a single pass/fail rate per run. That hides two things data teams care about when evaluating an analytics agent:

- **Variance across attempts** - an agent that solves a task on the 3rd retry looks the same as one that solves it on the 1st.
- **Consistency** - an agent that passes 75% of the time on a given task is not reliable enough for a customer-facing chat UI, but a single-run score won't surface that.

The standard eval metrics for this are `pass@k` and `pass^k`:

- `pass@k` - probability the agent gets **at least one** correct solution in k attempts. Rises with k (more shots on goal). `pass@1` is the strictest and usually the one that matters most for a single-shot analytics question.
- `pass^k` - probability that **all** k trials succeed. Falls with k. This is the reliability bar for customer-facing agents, where users expect the same correct answer every time (e.g. a 75% per-trial success rate gives ~42% pass^3).

Reference: https://www.anthropic.com/engineering/demystifying-evals-for-ai-agents

## Expected behavior

- `nao test` supports running each test case k times (e.g. `nao test --k 5`).
- Test output reports `pass@k` and `pass^k` alongside the existing accuracy score, both per test case and aggregated across the suite.
- `pass@1` is reported by default so existing single-run behavior stays readable.
- The metrics are available in whatever structured output `nao test` already emits, so teams can track them over time in CI.

## Customer ID

justin212407:
## Description

Closes: #1248

Adds `pass@k` and `pass^k` metrics to `nao test`, alongside the existing per-run pass/fail accuracy score. `nao test` can now run each test case `k` times (`nao test --k 5`), and reports `pass@1` (default), `pass@k`, and `pass^k` both per test case and aggregated across the suite, in the CLI output and in the structured JSON results file.

## Type of change

- [x]  Feature

## Changes made

- Added `TestConfig.k` (default `1`, `ge=1`) to `nao_config.yaml`'s `test:` block, overridable with `k` / `-k`, following the same pattern as the existing `threads` option.
- Each `(test case, model)` pair now runs `k` times instead of once. Added `TestRunResult.attempt` (1-indexed) so a run is self-describing even after the threaded executor reorders results.
- Added `pass_metrics_for_group()` / `summarize_pass_metrics()` in `summary.py`, computing metrics empirically per `(test, model)` group since `-k N` runs exactly N attempts, not an oversampled pool:
    - `pass@1` - mean of `passed` across the group's attempts
    - `pass@k` - 1.0 if any attempt passed, else 0.0
    - `pass^k` - 1.0 if all attempts passed, else 0.0
- Suite/model aggregates average per test case, not per raw attempt, so a test case run with k=5 counts once toward the aggregate.
- Result JSON gains a new top-level `pass_metrics` key (`{k, aggregate, by_test}`). Existing `summary` / `by_model` keys are
unchanged in shape; `ModelSummary` gained additive `k` / `pass_at_1` / `pass_at_k` / `pass_hat_k` fields.
- CLI table output shows `pass@1` always; `pass@k` / `pass^k` columns only appear when `k > 1`, so default (`k=1`) output is unchanged.
- `with_model_summaries()` still loads older result files without `k`/`pass_metrics`, so `nao test server` keeps working on old data.

## Test plan

- [x]  `make lint` passes (`ty check`, `ruff check`, import sort, `ruff format --check`)
- [x]  `uv run pytest -q` passes - 999 passed, 225 skipped (skips are a pre-existing `test_mssql.py` environment gap, unrelated to this change)
- [x]  Added unit tests for `pass_metrics_for_group` / `summarize_pass_metrics` (mixed pass/fail, all-pass, all-fail, k=1 equivalence, aggregation weighted by test case not attempt)
- [x]  Added tests for the runner's k-based fan-out, `attempt` population, and CLI/config override + rejection of invalid `k`
- [x]  Smoke-tested `nao test --k 3` against `example/`: confirmed config loading and correct run-count calculation (`1 test × 1 model × 3 attempts = 3 runs`)
- [ ]  Full live agent round-trip (`nao test --k 3` against a running backend with a real LLM key) - not validated end to end in this environment; plumbing up to the network boundary is verified by the above

**Manual testing steps:**

```bash
cd cli
make lint
uv run pytest -q --ignore=tests/nao_core/commands/sync/integration/test_mssql.py

cd ../example
uv run --project ../cli nao test           # confirm k=1 default output unchanged
uv run --project ../cli nao test --k 3     # confirm pass@1 / pass@3 / pass^3 columns
```

## Review guidance

Changed files: `cli/nao_core/commands/test/runner.py`, `cli/nao_core/commands/test/summary.py`, `cli/nao_core/config/test/__init__.py`, `cli/README.md`, `example/nao_config.yaml`, plus matching test files under
`cli/tests/`. No new dependencies, no changes outside `cli/` (backend/frontend untouched).

The metric-definition choice is worth a close look: since `--k N` runs exactly N attempts (not an oversampled pool), `pass@k`/`pass^k` are computed as direct any/all over the group rather than the combinatorial unbiased-estimator formula from the Codex/HumanEval pass@k paper - that estimator only applies when n > k. `summarize_pass_metrics()` and `pass_metrics_for_group()` in `summary.py` are the two functions to check this logic in.

## Checklist

- [x]  No changes to CSS/frontend (backend-agnostic CLI-only change).
- [x]  No `print`/debug statements left in.
- [x]  Existing tests updated where behavior changed; new tests added for new behavior.

## Screenshots:
Screenshot From 2026-08-08 15-05-04
Screenshot From 2026-08-08 15-05-14
Screenshot From 2026-08-08 15-04-38

<a href="https://cubic.dev/pr/getnao/nao/pull/1368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/nao_core/commands/test/runner.py
+++ b/cli/nao_core/commands/test/runner.py
@@ -21,7 +21,7 @@ from nao_core.ui import UI
 from .case import TESTS_FOLDER, TestCase, discover_tests
 from .client import BACKEND_URL, AgentClientError, VerificationResult, get_client
 from .compare import normalize_dataframe_numbers
-from .summary import ModelSummary, summarize, summarize_by_model
+from .summary import ModelSummary, summarize, summarize_by_model, summarize_pass_metrics
 
 
 @dataclass
@@ -70,6 +70,7 @@ class TestRunResult:
     tool_call_count: int | None = None
     error: str | None = None
     details: TestRunDetails | None = None
+    attempt: int | None = None
 
 
 def check_dataframe(
@@ -295,6 +296,7 @@ def save_results(results: list[TestRunResult], output_dir: Path) -> Path:
         "results": runs,
         "summary": summarize(runs),
         "by_model": [asdict(s) for s in summarize_by_model(runs)],
+        "pass_metrics": summarize_pass_metrics(runs),
     }
 
     output_file.write_text(json.dumps(data, indent=2))
@@ -302,7 +304,7 @@ def save_results(results: list[TestRunResult], output_dir: Path) -> Path:
 
 
 def print_run_table(results: list[TestRunResult]) -> None:
-    """Print one row per run, i.e. per (test, model) pair."""
+    """Print one row per run/attempt."""
     df = pd.DataFrame(
         [
             {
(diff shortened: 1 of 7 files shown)


justin212407:
@Bl3f Can you take a look and review this? Please lemme know if the changes follow the desired implementation. If not happy to follow up accordingly.

socallmebertille (changes requested):
Hi, thanks for your implementation !

Sorry for the lack of specification on this issue. Let me give you a preview of how we'd like the new metrics to be rendered in the terminal.

- First table (the one printed by nao test, one row per test) :
	- The green check in the Status column should only appear when all runs of that test passed (100% success)
	- Add a new column next to it showing the % of runs that passed for that test
- Second table (the summary) :
	- Instead of listing every run with its success %, show only a single totals row — i.e. the aggregate across all tests (like the "total" line of the first table)
	- Add an "Always Pass %" column to this totals row: the share of tests that passed on every run (100% success across all attempts)

Something similar to :

Image

Let me know if anything is unclear!

justin212407:
@socallmebertille I have updated the pr with the intended layout we wanted for this issue in the recent commit, here is the mockup of how the outputs look like as of now:
Screenshot From 2026-08-27 11-20-30

cli/nao_core/commands/test/runner.py
socallmebertille, line 537:
```suggestion
		UI.print(f"[dim] Attempts: {k_count}[/dim]")
```

socallmebertille (changes requested):
Hi @justin212407 , thanks for your commit!

Sorry for the delay, I just reviewed your changes. Everything will be good once you've addressed the last Cubic comment, resolve conflicts and the suggestion I left below! 👍

PR #1373, opened 2026-08-10
Exclude columns also at the SQL Level. Columns hidden in config can't be queried by agent

### LINKED ISSUE #1015: [feature] exclude_columns should block the agent from querying those columns in execute_sql
## Problem

`exclude_columns` today only strips the listed columns from the **context generated by `nao sync`** - the agent just doesn't see them in the synced schema/markdown. It does **not** stop the agent from actually querying them.

That means a column an operator explicitly excluded (typically sensitive / PII) can still be returned if the agent references it directly in `execute_sql` - the exclusion is a context-visibility filter, not an access control. For a governance feature this is a real gap: excluded means excluded, including at query time.

## Expected behavior

`exclude_columns` should be enforced at query execution. When the agent runs `execute_sql`, any query that references an excluded column (for a matching `schema.table.column`, using the existing glob patterns) must be **rejected / blocked** rather than executed.

- Enforcement happens in the `execute_sql` path, not just at sync time.
- The block respects the same `exclude_columns` patterns already defined in `nao_config.yaml`.
- The agent should get a clear error back so it can adjust, rather than silently returning excluded data.

## Technical requirement

This needs **SQL comprehension / parsing**, not a naive string match: to know whether a query touches an excluded column we have to parse the SQL and resolve which columns it actually references (including `SELECT *`, aliases, table qualifiers, subqueries, CTEs, and joins across the relevant tables). A substring check on the raw SQL is not sufficient.

## Open questions (TBD)

- How to handle `SELECT *` against a table that has excluded columns - reject, or auto-expand and strip the excluded ones?
- Behavior when column references can't be fully resolved (ambiguous / dynamic SQL) - fail closed (block) or warn?

ad4mou:
## Summary
- Make `exclude_columns` actually apply when the agent runs SQL, not only when syncing context.
- Explicit use of an excluded column is blocked with a clear error.
- `SELECT *` still runs, but excluded columns are removed first and a yellow warning explains what was stripped.
- If we can’t safely tell whether a query touches excluded columns, we block it.

Closes #1015

<a href="https://cubic.dev/pr/getnao/nao/pull/1373?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/fastapi/main.py
```diff
@@ -264,21 +280,39 @@ async def execute_sql(request: ExecuteSQLRequest):
...
+                        status_code=400,
+                        detail=(
+                            "azure_access_token is required when the database auth_mode is "
+                            "'azure_entra_id'. Runtime queries must use the end user's access "
+                            "token; any configured user/password is only used by nao sync."
+                        ),
+                    )
+                validated_sql = request.sql
+                if db_config.exclude_columns:
+                    if not getattr(db_config, "user", None) or not getattr(
```
Bl3f: I don't think it's the best Python structure we can avoid this check + raise here and move the exclude_columns check before the whole thing so we validate the SQL and then pass a validated SQL below once for all.

apps/backend/fastapi/test_main.py
```diff
@@ -58,6 +85,43 @@ def test_execute_sql_simple_duckdb(duckdb_project_folder):
...
+        "/execute_sql",
+        json={
+            "sql": "SELECT * FROM users",
+            "nao_project_folder": duckdb_project_with_excluded_columns,
+        },
+    )
+
+    assert response.status_code == 400
+    assert response.json()["detail"] == (
+        "Query blocked because SELECT * would include excluded column(s): main.users.email. "
```
Bl3f, line 123: When dialect is BigQuery, Snowflake, Databricks, DuckDB and Clickhouse we can recommend the LLM to use EXCEPT ou EXCLUDE (cf. https://claude.ai/share/09c2d2d5-9369-4e7d-a370-6a04e7abd6f0) instead of asking to write every column

cli/nao_core/config/databases/exclude_columns_guard.py
```diff
@@ -0,0 +1,490 @@
...
+        raise _blocked(f"the database dialect '{db_config.type}' is not supported")
+
+    expression = _parse_query(sql, dialect)
+    if not any(expression.find_all(exp.Table)):
+        return sql
+
+    owns_connection = conn is None
+    try:
+        if conn is None:
+            conn = db_config.connect()
```
Bl3f: I don't think the column guard should connect to the database to get info about tables ; it's weird that it needs it.

To me you should parse the SQL query and from the selected columns we should check if a column is wrongly selected and that's all. Also I think 480 rows to do this operation is a bit too much.

cli/nao_core/config/databases/base.py
```diff
@@ -190,11 +190,13 @@ def connect(self) -> BaseBackend:
         """Create an Ibis connection for this database."""
         ...
 
-    def execute_sql(self, sql: str) -> pd.DataFrame:
+    def execute_sql(self, sql: str, conn: BaseBackend | None = None) -> pd.DataFrame:
         """Execute arbitrary SQL and return results as a DataFrame."""
         import pandas as pd  # noqa: F811
 
-        conn = self.connect()
+        owns_connection = conn is None
```
Bl3f: As said in another file I don't think columns guards need a connection so not sure we need to expose or re-use a connection from elsewhere.

Bl3f (changes requested):
Need more work I think ; esp. on the column guard, which is not really the best name I'd say.

We should have a parser and from this parser raise usage errors on columns/tables

cli/nao_core/config/databases/column_access.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def _load_table_infos(
+    expression: exp.Query,
+    project_path: Path,
+    db_config: _DatabaseConfigLike,
+) -> TableInfos:
+    database_folder = get_database_folder_names([db_config])[0]
+    path = column_catalog_path(project_path, db_config.type, database_folder)
+    catalog = load_column_catalog(path)
```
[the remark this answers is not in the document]
ad4mou: This is the intended behavior.
The reason is, assuming there was no error in the column catalog generation, then our column exclusion will work only as of the latest nao sync. If new tables are added, we have no way to know if they should be blocked or not, and it's unlikely that they should be blocked since the config was done before those were added, so I think they shouldn't be blocked by default.

cli/nao_core/config/databases/column_catalog.py
```diff
@@ -0,0 +1,83 @@
...
+
+COLUMN_CATALOG_VERSION = 1
+
+Column = dict[str, str]
+Tables = dict[str, list[Column]]
+Schemas = dict[str, Tables]
+
+
+def column_catalog_path(project_path: Path, database_type: str, database_folder: str) -> Path:
+    return project_path / ".meta" / "databases" / f"type={database_type}" / database_folder / "columns.json"
```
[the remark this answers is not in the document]
ad4mou: true but has nothing to do with this PR

ad4mou:
I tested extensively with a DuckDB and a BigQuery DB, the feature works great.
Although the agent can kind of go crazy is the choices of columns to exclude are not judicious

ClaireGz:
@ad4mou we will put this under the enterprise license - can you implement the change?

ad4mou:
Yes absolutely, this is what it looks like for now.
I'll wait until the new settings are merged to add the little Entreprise lock and to add the feature to the list in the new Entreprise page.

https://github.com/user-attachments/assets/792d3714-74cf-44af-9391-d2ca21bcd52d

apps/backend/fastapi/main.py
```diff
@@ -97,6 +101,7 @@ class ExecuteSQLRequest(BaseModel):
     database_id: str | None = None
     env_vars: dict[str, str] | None = None
     azure_access_token: str | None = None
+    enforce_excluded_columns: bool = False
```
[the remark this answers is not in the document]
ad4mou: Resolved in 7db991a. Live-story and story-filter SQL now use the same licence-and-project-setting check as agent SQL and explicitly send `enforce_excluded_columns`. The FastAPI default remains `False` so callers cannot enable this enterprise feature without a valid licence.

ad4mou:
LGTM

apps/backend/src/services/excluded-columns.service.ts
```diff
@@ -0,0 +1,6 @@
+import type { AgentSettings } from '../types/agent-settings';
```
Bl3f, line 4: Add the Enterprise licence header for this file.

apps/backend/src/services/live-story.ts
```diff
@@ -38,8 +39,17 @@ export async function executeLiveQuery(
 		throw new Error('Project path not configured');
 	}
 
-	const envVars = await projectQueries.getEnvVars(projectId);
-	return executeRawSql(stripSqlFilterBlocks(query.sqlQuery), project.path, query.databaseId, envVars);
+	const [envVars, agentSettings] = await Promise.all([
+		projectQueries.getEnvVars(projectId),
+		projectQueries.getAgentSettings(projectId),
+	]);
+	const enforceExcludedColumns = await resolveExcludedColumnEnforcement(agentSettings);
```
Bl3f: Shouldn't we factorise this pattern elsewhere to avoid changing at multiple places everytime we want to run safe SQL?

apps/backend/src/services/live-story.ts
```diff
@@ -67,17 +77,21 @@ export async function refreshStoryData(chatId: string, slug: string): Promise<Re
 		throw new Error('Project path not configured');
 	}
 
+	const [projectEnvVars, agentSettings] = await Promise.all([
+		projectQueries.getEnvVars(projectId),
+		projectQueries.getAgentSettings(projectId),
+	]);
+	const enforceExcludedColumns = await resolveExcludedColumnEnforcement(agentSettings);
```
Bl3f: Related to the factorisation.

apps/backend/src/services/live-story.ts
Bl3f: Like moving this to executeRawSQL? so we never forget to pass it and make a unsecure call by mistake

Bl3f (approved):
A few comments, except from this LGTM

PR #1403, opened 2026-08-14
Fix File Explorer failing to load when Git is unreachable

ad4mou:
## Summary

- The File Explorer file list now comes straight from the context folder. It no longer waits on the connected repo, so a repo problem can't stop your files from showing.
- Opening a file falls back to the folder's copy, read-only, when the repo can't be reached instead of erroring.
- Saving still fails with the real reason, so genuine problems stay visible.
- Removed the unused "tracked" flag from the file list.

## Why

Listing files and resolving the connected repo were one step, and that step can clone over the network. Any hiccup gave "Failed to load files" even though the files were readable the whole time. Now they're independent: files first, repo after, repo only decides what's editable.

<a href="https://cubic.dev/pr/getnao/nao/pull/1403?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/trpc/context-explorer.routes.ts
+++ b/apps/backend/src/trpc/context-explorer.routes.ts
@@ -5,7 +5,7 @@ import { z } from 'zod';
 import * as userQueries from '../queries/user.queries';
 import type { ContextExplorerFileAccess } from '../services/context-explorer.service';
 import {
-	getFileTreeResponse,
+	getFileTree,
 	MAX_CONTEXT_FILE_SIZE,
 	readFileContent,
 	searchFileContents,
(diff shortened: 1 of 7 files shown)


apps/backend/src/trpc/context-explorer.routes.ts
```diff
@@ -77,12 +78,12 @@ export const contextExplorerRoutes = {
 	}),
 
 	getFileTree: contextAdminProtectedProcedure.query(async ({ ctx }) => {
-		const access = await createFileAccess(ctx.project.id, ctx.project.path, ctx.user);
-		return getFileTreeResponse(access);
+		const entries = await getFileTree(requireProjectPath(ctx.project.path));
```
ad4mou, line 81: key change here, this function does nothing related to git now so that git failures don't affect our source context files

PR #1408, opened 2026-08-17
feat:nao test: assert the agent performed a specific action

justin212407:
## Description

Today, `nao test` can only verify final agent output by comparing the generated SQL's result against a reference `sql` dataframe. There's no way to assert on *intermediate* agent behavior, so a test case can't check things like "did the agent ask a clarifying question before running SQL" or "did the agent call a specific tool."

This adds an extensible `assertions` mechanism to `nao test` that checks the agent's recorded tool-call trace independently of (and combinable with) the existing dataframe-based SQL verification.

## Changes made

- New `Assertion` framework (`cli/nao_core/commands/test/assertions.py`):
  introduces a `ToolCallAssertion` type that checks whether a named tool was invoked, with optional `args` subset-matching (recursive for nested dicts) and `min_count`. `parse_assertions()` validates the `assertions` list from a test YAML and raises `AssertionConfigError` on malformed entries (missing `type`, unknown fields, invalid `min_count`, etc.). `evaluate_assertions()` runs all assertions against the tool-call trace and returns a combined pass/fail with a human-readable message.
- `TestCase` (`cli/nao_core/commands/test/case.py`) now carries an `assertions: list[Assertion]` field, parsed from the test YAML alongside the existing `sql` field, which is now optional (`sql: str | None = None`) so assertion-only tests don't need reference SQL.
- `runner.py`: assertions are evaluated alongside SQL verification when both are present (`passed = assertions_passed and data_passed`), and a new code path handles assertion-only test cases (no `sql`) by reporting pass/fail on the assertions alone.
- Backend support: `apps/backend/src/routes/test.ts` makes the `sql` field in the test request body optional (defaults to `''`), and `apps/backend/src/agents/tools/index.ts` keeps the `clarification` tool available in `testMode` (previously stripped) so `nao test` can assert on clarifying follow-up questions and other intermediate actions in the recorded tool-call trace.
- Docs: `cli/README.md` documents the new `assertions` YAML syntax with a `tool_call` example (asserting a `clarification` follow-up), and explains that `tool_call` assertions support `tool`, `args`, and `min_count`, and that SQL verification and assertions can be combined.

## Screenshots
Screenshot From 2026-08-17 11-58-42

## Example test case

```yaml
name: ambiguous_revenue_period
prompt: What was the revenue?
assertions:
  - type: tool_call
    tool: clarification
```

## Validation

- Added unit tests: `tests/nao_core/commands/test_assertions.py`, `test_case.py`, `test_runner.py` covering assertion parsing/validation, `TestCase` YAML loading with assertions, and runner behavior for combined SQL + assertion checks and assertion-only test cases.
- `apps/backend/src/routes/test.ts` — verified optional `sql` field doesn't break existing SQL-based test payloads.
- Manual check: an assertion-only test YAML (no `sql`) runs end-to-end through `nao test` and reports pass/fail based solely on the tool-call trace.

<a href="https://cubic.dev/pr/getnao/nao/pull/1408?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


justin212407:
@Bl3f ptal and lmk if there are any changes that you might want me to make

apps/backend/src/agents/tools/index.ts
```diff
@@ -89,14 +94,9 @@ export const getTools = (
...
-		execute_python,
-		execute_sandboxed_code,
-		clarification: clarificationTool,
-		suggest_follow_ups,
-		write: writeTool,
-		...rest
-	} = tools;
+	const { execute_python, execute_sandboxed_code, suggest_follow_ups, write: writeTool, ...rest } = tools;
+	// Keep clarification available in testMode so `nao test` can assert follow-up
+	// questions and other intermediate actions on the recorded tool-call trace.
```
socallmebertille, line 99: Maybe we should avoid cluttering this file with this comment, this is basically information you've already included in the PR description, there is no need to repeat it here again

```suggestion
```

apps/backend/src/agents/tools/index.ts
```diff
@@ -55,6 +55,11 @@ export const getTools = (
 	agentSettings: AgentSettings | null,
 	extraTools?: Record<string, unknown>,
 	options: {
+		/**
+		 * @deprecated No longer strips tools. Kept so existing callers that pass
+		 * `testMode` continue to typecheck. Clarification stays available so
+		 * `nao test` can assert intermediate actions (e.g. follow-up questions).
+		 */
 		testMode?: boolean;
```
socallmebertille, line 63: You can delete this part as long as you remove the three call-site props in the same edit.


```suggestion
```

cli/tests/nao_core/commands/test_assertions.py
socallmebertille, line 1: I think this file is unnecessary

socallmebertille (changes requested):
Hi, thanks for your contribution!

You should remove the `!testMode` condition into `apps/backend/src/components/ai/system-prompt.tsx` (l. 109) if you want to make available the `clarification` tool in test mode.

justin212407:
I have removed the `!testMode` condition so clarification guidance is available during tests, removed the obsolete testMode option and its call sites, the redundant comments, and deleted the standalone test_assertions.py file.n Also updated test-mode stopping so runs terminate when clarification is called.

cli/nao_core/commands/test/assertions.py
socallmebertille: I still think the test file is unnecessary

socallmebertille: I still think the test file is unnecessary

cli/tests/nao_core/commands/test_case.py
socallmebertille, line 1: Could you drop this one ? It’s the same assertion-internals coverage we already removed with `@test_assertions.py`. The runner tests are enough.

cli/nao_core/config/databases/duckdb.py
socallmebertille, line 1: Where does this changes came from ? Are you sure it is PR related ?

socallmebertille (changes requested):
Hi again @justin212407, sorry for the delay !

Few final points :
- one cubic-driven unit test that recreates the `@test_assertions.py` coverage we already dropped
- `@cli/nao_core/config/databases/duckdb.py` doesn’t look related to this PR — can you confirm or remove it if so ?

cli/tests/nao_core/commands/test_case.py
justin212407: Yep dropped the file as you suggested!!

cli/nao_core/config/databases/duckdb.py
justin212407: I am sorry it got mixed up with one of the other prs i was working on. fixed in the recent commit.

socallmebertille (approved):
Look really good, please resolve latest conflicts and it will be good :)

justin212407:
> Look really good, please resolve latest conflicts and it will be good :)

Hey I have resolved the conflicts, ready to merge!!

PR #1412, opened 2026-08-18
Org members inherit access to all projects

### LINKED ISSUE #1173: [feature] Org members should automatically get access to the org default project (SSO or manual invite)
## Current issue

On self-hosted instances there is no way to set a default project to all organization members. A new user who signs in for the first time with SSO gets stuck at onboarding with the message "No project configured. Set NAO_DEFAULT_PROJECT_PATH environment variable."

This happens even when a default project already exists and works for the admin. Setting or re-setting the `NAO_DEFAULT_PROJECT_PATH` environment variable and restarting does not help, so the only workaround today is for an admin to add each user to the project by hand.

The same gap applies to users added manually by an admin, not only to SSO sign-ups: joining the org does not grant access to any project.

## What we want

Let an organization define a default project that all of its members automatically have access to. Every user in the org (existing members and anyone who joins later) should get access to that default project without an admin adding them one by one.

This should hold regardless of how the user joined the org:
- signing up through SSO
- being added manually by an admin

When the org has only one project, that project should be used as the default - no explicit configuration required.

## Customer ID

1c5a6e32-5b65-4ebb-823d-7c55f86bcf50
44e63512-73b5-4ec6-bed6-2a6df27ea43b

ad4mou:
## Summary

- Members of an organization can now access all the org's project without an admin adding them to
  it one by one. This fixes new users getting stuck on "No project configured" bug.
- Being in the org is what grants access; people still don't show up in the project's team
  list unless they were added there explicitly.
- Their permission level comes from their org role, unless they've been given a specific
  role on the project, which wins.

Closes #1173

<a href="https://cubic.dev/pr/getnao/nao/pull/1412?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -149,7 +149,6 @@ export const listProjectMembersWithRoles = async (projectId: string): Promise<Us
 			name: s.user.name,
 			email: s.user.email,
 			role: s.projectMember.role,
-			messagingProviderCode: s.user.messagingProviderCode,
 		})
 		.from(s.user)
 		.innerJoin(s.projectMember, eq(s.projectMember.userId, s.user.id))
@@ -167,7 +166,6 @@ export const listUsersWithProjectAccess = async (projectId: string): Promise<Use
 			name: s.user.name,
 			email: s.user.email,
 			role: sql<UserRole>`coalesce(${s.projectMember.role}, ${s.orgMember.role})`,
-			messagingProviderCode: s.user.messagingProviderCode,
 		})
 		.from(s.user)
 		.leftJoin(s.projectMember, and(eq(s.projectMember.userId, s.user.id), eq(s.projectMember.projectId, projectId)))
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -105,7 +105,7 @@ export const listUserProjectsWithRoles = async (userId: string): Promise<UserPro
 	const results = await db
 		.select({
 			project: s.project,
-			userRole: sql<UserRole>`coalesce(${s.projectMember.role}, 'viewer')`,
+			userRole: sql<UserRole>`coalesce(${s.projectMember.role}, ${s.orgMember.role}, 'viewer')`,
 		})
 		.from(s.project)
 		.leftJoin(s.projectMember, and(eq(s.projectMember.projectId, s.project.id), eq(s.projectMember.userId, userId)))
--- a/apps/backend/src/queries/project.queries.ts
+++ b/apps/backend/src/queries/project.queries.ts
@@ -208,8 +208,8 @@ export const getProjectByUserId = async (
 		return null;
 	}
 
-	const membership = await getProjectMember(project.id, userId);
-	return membership ? project : null;
+	const role = await getUserRoleInProject(project.id, userId);
+	return role ? project : null;
 };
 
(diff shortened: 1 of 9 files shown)


apps/backend/src/queries/project.queries.ts
```diff
@@ -208,8 +208,8 @@ export const getProjectByUserId = async (
 		return null;
 	}
 
-	const membership = await getProjectMember(project.id, userId);
-	return membership ? project : null;
+	const role = await getUserRoleInProject(project.id, userId);
```
ad4mou, line 211: key change is here, this was the only place where the wrong function was called: `getProjectMember`

ad4mou:
Since org members have access to all projects. They should not get added to all project teams.
So I cleaned up a few functions who did that and hence were now useless

apps/backend/src/trpc/project.routes.ts
```diff
@@ -703,6 +703,10 @@ export const projectRoutes = {
 		return projectQueries.listProjectMembersWithRoles(ctx.project.id);
 	}),
 
+	listUsersWithAccess: projectProtectedProcedure.query(async ({ ctx }) => {
+		return projectQueries.listUsersWithProjectAccess(ctx.project.id);
```
Bl3f, line 707: `listUsersWithProjectAccess` returns the `messagingProviderCode` it means every user in a project can get the providerCode of someone else and try to login in their place I think.

Bl3f: looks like it was also the case before but i think it can be fixed

Bl3f: and why do we need this route?

apps/frontend/src/routes/_sidebar-layout.settings.project.budgets.tsx
```diff
@@ -138,7 +138,7 @@ function RouteComponent() {
 		enabled: perUserSpendEnabled,
 	});
 	const projectMembers = useQuery({
-		...trpc.project.listAllUsersWithRoles.queryOptions(),
+		...trpc.project.listUsersWithAccess.queryOptions(),
```
Bl3f, line 141: why? we also added the coalesced to listAllUsersWithRoles it seems

Bl3f (commented):
A few comments.

apps/backend/src/trpc/project.routes.ts
ad4mou: Agreed on `messagingProviderCode`. I’ll remove it from all user-list responses.

We do need this route for per-user budgets. Org members can generate spend without being explicit project members, so `listAllUsersWithRoles` omits users who need to appear there.

apps/frontend/src/routes/_sidebar-layout.settings.project.budgets.tsx
ad4mou: Ok It's a bit a confusing but I think:

`listAllUsersWithRoles` calls `listProjectMembersWithRoles`, which only returns explicit project members. 

`listUsersWithAccess` calls `listUsersWithProjectAccess`, whose coalesce also handles org members. Since org members can generate spend, the budget page must use `listUsersWithAccess`.

ad4mou:
LGTM

PR #1413, opened 2026-08-18
Tell the agent to use database id if multiple dbs are configured

### LINKED ISSUE #1004: [bug] execute_sql with several databases connected always fails at first shot
## Problem
One user has several databases connected, one in the US, one in EU

SQL execution fails recurrently with:
`Error executing SQL query: {"message":"Multiple databases configured. Please specify database_id.","available_databases":["DB-1","DB-2"]}`

This happens even after the context rules were edited to tell the agent there are two databases. The agent does not pass a `database_id` when multiple databases are configured, so the query fails. Reported as a recurring error affecting multiple users.

## Expected behavior
When multiple databases are configured, the agent should resolve and pass the correct `database_id` in the SQL execution call (from the context rules, or by asking the user which one), instead of failing. Rules that declare the available databases should be honored.

## Customer ID

ad4mou:
## Summary
- When a project has several databases connected, the agent's first SQL query used to fail: nothing told it which database names it was allowed to use, so it either sent none or copied a folder name from the synced metadata. It retried and usually recovered, but the user saw a red error first.
- The agent is now handed the actual database names from `nao_config.yaml`, each with its type and a couple of identifying details, so it picks the right one on the first try.
- Databases that are configured but not yet synced are now visible to it too, previously they were entirely absent from what it could see, which is likely how most of these failures started.
- Nothing changes for single-database projects: the list is only added to the prompt when two or more are configured.

Closes #1004

<a href="https://cubic.dev/pr/getnao/nao/pull/1413?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/components/ai/system-prompt.tsx
+++ b/apps/backend/src/components/ai/system-prompt.tsx
@@ -9,6 +9,7 @@ import { tokenCounter } from '../../services/token-counter';
 import type { UserMemory } from '../../types/memory';
 import { MEMORY_CATEGORIES, MemoryCategory } from '../../types/memory';
 import { formatCurrentDate } from '../../utils/date';
+import type { ConfiguredDatabase } from '../../utils/nao-config';
 import { groupBy } from '../../utils/utils';
 import { getDialectSqlQueryRules, getDialectToolCallRules } from './dialect-rules';
 import { NaoContextStructure } from './nao-context-structure';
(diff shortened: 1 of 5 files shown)


Bl3f (approved):
LGTM

PR #1416, opened 2026-08-18
[fix : story] data unavailable due to wrong `query_id`

socallmebertille:
## Issue

Some LLM models take the liberty of assigning common English names to `query_ids` when querying data, and then passing them to stories to create blocks of charts, tables, or maps.

## Solution

To prevent this issue from recurring, I added a validator that checks the format of the query ID and verifies it corresponds to an `execute_sql` result produced in the same chat. Mismatches are surfaced as non-blocking template_warnings so the agent fixes the reference before finishing.

<a href="https://cubic.dev/pr/getnao/nao/pull/1416?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f (approved):
LGTM

PR #1417, opened 2026-08-18
Add an adaptive prompt section describing the available context and what each template contains

### LINKED ISSUE #980: [feature] Improve agent system prompt to navigate file system
## Problem
The agent doesn't have enough clarity on how to navigate the context file system, which leads it to read the wrong file (or no file).
This is currently patched by the skill to create RULES.md - which creates a personnalized overview of context structure.
But we can't rely on users to use the skill to create and maintain those rules.

1. **No explanation of the templates of tables context.** The agent doesn't understand what each template is for (schema, how_to_use, etc.) or when to use each one.
2. **No guidance on the repos / docs / semantics folders.** The system prompt doesn't tell the agent these folders exist or how to use them, so relevant context goes unread.

## Expected behavior
- Overall better orchestration of the agent across the whole context folder
- - The system prompt includes guidance on the repos / docs / semantics folders so the agent knows they exist and when to read them.
- The system prompt explains each table-context template and when the agent should consult it, ideally surfaced **dynamically** based on which files actually exist in the context file system for a given table.

## Example
As an example, these are the rules I added in our own Rules.md to help agent:
`Context Map
Where context lives beyond this file. Read the pointer, then open the referenced file.

Per-table context — databases/type=bigquery/database=nao-production/schema=prod_silver/table=<table>/:

columns.md — full column list + types.
annotations.md — dbt model + column descriptions (the authoritative business definition of the table/columns, copied from repos/dbt/models/*.yml). Read this first for what a column means. A short stub means the table has no dbt model doc.
ai_summary.md — what the table represents + data-quality/ambiguity risks (LLM-generated).
profiling.md — distinct counts, min/max, and top_values for categorical columns. Read before filtering on any column value.
preview.md — sample rows.
dbt repo — repos/dbt/:

models/silver.yml — silver model + column documentation.
models/silver_semantic_layer.yml — semantic layer: canonical measures + top-level metrics: (MRR, churn, active users, tokens, …). Metric definitions are owned here.

Docs — docs/:
crm.md — CRM funnel statuses, company tiering, opportunity-stage taxonomy. Read before any sales/pipeline question.`

## Open question
How do we match what each user actually builds in their context? One option: regenerate this navigation guidance automatically on every context push, so the prompt reflects the real folder structure rather than a static assumption.

## Customer ID

ad4mou:
## Summary

- The agent now gets a short guide to the project context in its system prompt: what each folder is for, and what each per-table file contains and when to read it. Before, it was only told the files existed, so it often read the wrong one or none.
- The guide only mentions what the project actually has. The per-table file list comes from `nao_config.yaml`, and folders like `semantics/` or `docs/` appear only when they exist and have something in them so the agent never goes looking for something that isn't there.
- `annotations.md` is now called out as the place for human notes, and the agent is told to trust it over the generated files when they disagree.
- Removed leftover mentions of `description.md` and `how_to_use.md` in the recommendations prompt and the two context skills. Those files were deleted a while back and no longer exist.
- The checks run once per user message.

Closes #980

## Two decisions regarding the issue

- **Per project, not per table.** The issue asked for the file list to adapt to each table. The config already decides which files sync writes, and checking every table folder would mean scanning thousands of directories for the same answer. So the list adapts per project instead.
- **No regeneration on context push.** : The genuinely useful version where the prompt adapts to the user's DB ("read `annotations.md` first for what a column means", "check `crm.md` before sales questions") needs a model to read the files and understand them which is already what the `write-context-rules` skill already does, with a human reviewing the result. Doing it automatically on every sync would pay for that pass every time and ship it unreviewed.

<a href="https://cubic.dev/pr/getnao/nao/pull/1417?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/components/ai/context-recommendations-system-prompt.tsx
+++ b/apps/backend/src/components/ai/context-recommendations-system-prompt.tsx
@@ -3,12 +3,15 @@ import dbConfig, { Dialect } from '../../db/dbConfig';
 import { Block, Bold, Br, Code, CodeBlock, List, ListItem, renderToMarkdown, Span, Title } from '../../lib/markdown';
 import type { LinkedContextRepo } from '../../types/context-recommendation';
 import { ALLOWED_APP_DB_VIEWS } from '../../utils/app-db-allowlist';
+import type { ContextPresence } from '../../utils/nao-config';
 import { AppDbTimestamps } from './app-db-timestamps';
 import { NaoContextStructure } from './nao-context-structure';
 
 export function renderContextRecommendationsSystemPrompt(options?: {
 	proposeFixes?: boolean;
 	linkedRepos?: LinkedContextRepo[];
+	templates?: string[];
+	contextPresence?: ContextPresence;
 	contextRepoConnected?: boolean;
 	customInstructions?: string;
 }): string {
@@ -18,6 +21,8 @@ export function renderContextRecommendationsSystemPrompt(options?: {
 		<ContextRecommendationsSystemPrompt
 			proposeFixes={options?.proposeFixes ?? false}
 			linkedRepos={options?.linkedRepos ?? []}
+			templates={options?.templates}
+			contextPresence={options?.contextPresence}
 			contextRepoConnected={options?.contextRepoConnected ?? false}
 			customInstructions={customInstructions}
 		/>,
@@ -27,11 +32,15 @@ export function renderContextRecommendationsSystemPrompt(options?: {
 function ContextRecommendationsSystemPrompt({
 	proposeFixes,
 	linkedRepos,
+	templates,
+	contextPresence,
 	contextRepoConnected,
 	customInstructions,
 }: {
 	proposeFixes: boolean;
 	linkedRepos: LinkedContextRepo[];
+	templates?: string[];
+	contextPresence?: ContextPresence;
 	contextRepoConnected: boolean;
 	customInstructions?: string;
 }) {
@@ -50,7 +59,11 @@ function ContextRecommendationsSystemPrompt({
 				context that you may recommend improving, correcting, or extending.
 			</Span>
 
-			<NaoContextStructure />
+			<NaoContextStructure
+				templates={templates}
+				repoNames={linkedRepos.map((repo) => repo.name)}
+				contextPresence={contextPresence}
+			/>
 			<Span>
 				<Code>RULES.md</Code> and <Code>semantics/*.md</Code> hold the project-wide rules and metric definitions
 				the agent relies on — the most common place a fix belongs.
--- a/apps/backend/src/components/ai/nao-context-structure.tsx
+++ b/apps/backend/src/components/ai/nao-context-structure.tsx
@@ -1,28 +1,116 @@
-import { Block, Italic, List, ListItem, Title } from '../../lib/markdown';
+import { Block, Br, Code, List, ListItem, Title } from '../../lib/markdown';
+import type { ContextPresence } from '../../utils/nao-config';
 
 /** Explains how the nao project context is laid out on disk. Shared by every system prompt. */
-export function NaoContextStructure() {
+export function NaoContextStructure({
+	templates,
+	repoNames = [],
+	contextPresence,
+}: {
+	templates?: string[];
+	repoNames?: string[];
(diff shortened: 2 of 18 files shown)


apps/backend/src/components/ai/context-recommendations-system-prompt.tsx
```diff
@@ -50,7 +59,11 @@ function ContextRecommendationsSystemPrompt({
 				context that you may recommend improving, correcting, or extending.
 			</Span>
 
-			<NaoContextStructure />
+			<NaoContextStructure
+				templates={templates}
+				repoNames={linkedRepos.map((repo) => repo.name)}
+				contextPresence={contextPresence}
```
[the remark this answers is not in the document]
ad4mou: Not applying this one. The paths are unconditional here on purpose.

This is the context-audit agent, whose job is to find context that's *missing* and
propose creating it. "`RULES.md` and `semantics/*.md` are the most common place a fix
belongs" is precisely what it needs to hear when those files don't exist yet 
-> a missing `RULES.md` is a finding, not a stale path.

`contextPresence` gates the `NaoContextStructure` bullets because those describe what's
available to *read*. The surrounding prose describes where a fix should *land*, which is
a different question and shouldn't shrink to what happens to exist today.

apps/backend/src/utils/nao-config.ts
```diff
@@ -62,6 +99,69 @@ export function extractConfiguredRepos(projectFolder: string): LinkedContextRepo
...
+	for (const database of config.databases) {
+		if (!isRecord(database)) {
+			continue;
+		}
+
+		const templates = getDatabaseTemplates(database);
+		const migratedTemplates = [
+			...new Set(
+				templates
+					.filter((template) => template !== 'description')
```
ad4mou, line 117: this feels stale but it's actually because we still accept `desription` in nao_config without labelling it a bad config so that users upgrading from older versions don't get their sync refused. So we need to filter it out here since our new prompt doesn't support `description`

ad4mou:
LGTM

apps/backend/src/components/ai/nao-context-structure.tsx
```diff
@@ -1,28 +1,143 @@
...
-				<ListItem>
-					Databases folders are named following this pattern: type={`<database_type>`}/database=
-					{`<database_name>`}/schema={`<schema_name>`}/table={`<table_name>`}.
-				</ListItem>
-				<ListItem>
-					Each table has files describing the table schema and the data in the table (like columns.md,
-					preview.md, etc.)
-				</ListItem>
+				{[
+					isPresent('rules') && (
```
Bl3f: Not necessary, rules is always concatenated to the system prompt.

apps/backend/src/utils/nao-config.ts
```diff
@@ -9,6 +9,16 @@ import type { LinkedContextRepo } from '../types/context-recommendation';
 import { logger } from './logger';
 
 const ENV_PATTERN = /\$?\{\{\s*env\(['"]([^'"]+)['"]\)\s*\}\}/g;
+const DATABASE_TEMPLATES = ['columns', 'preview', 'profiling', 'query_history', 'ai_summary'] as const;
```
Bl3f, line 12: i dont like the direction this nao-config.ts is going because here it means we need to maintain 2 truth for the schema of the file (in python and in ts) we should find a way to wrap and import the yaml parser from the python in ts so we dont rewrite it twice. I think it's for another PR, but looks ugly and it's growing

Bl3f (approved):
small comment and change LGTM

apps/backend/src/utils/nao-config.ts
ad4mou: ok yeah, see #1685

apps/backend/src/components/ai/nao-context-structure.tsx
ad4mou: yeah right, fixed

PR #1419, opened 2026-08-19
fix(stories): persist MCP query templates

btalx:
## Summary

Fix story filters for charts created through the MCP context-layer flow.

MCP execute_sql previously persisted only result rows. When create_story or update_story attached those rows to a chat, it pinned an empty SQL string, so the story could render cached data but could not validate or re-execute filter templates.

## Changes

- persist sql_query and database_id with MCP query data
- backfill retained query definitions from MCP call logs in both PostgreSQL and SQLite migrations
- pin the persisted definition with cached results when a story is attached to a chat
- preserve the existing data-only behavior for chat-created queries
- add regression coverage for both paths

## Verification

- targeted story/filter tests: 6 passed
- full backend/frontend/shared lint: passed
- formatting check: passed
- PostgreSQL and SQLite migration consistency checks: passed
- fresh SQLite migration: passed
- complete pre-commit hook, including Python CLI checks: passed

Linear: https://linear.app/lucis-life/issue/ENG-9200/nao-story-filters-render-but-do-not-re-execute-persisted-chart-queries

<a href="https://cubic.dev/pr/getnao/nao/pull/1419?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/migrations-postgres/0064_persist_mcp_query_definitions.sql
+++ b/apps/backend/migrations-postgres/0064_persist_mcp_query_definitions.sql
@@ -1,0 +1,9 @@
+ALTER TABLE "mcp_query_data" ADD COLUMN "sql_query" text;--> statement-breakpoint
+ALTER TABLE "mcp_query_data" ADD COLUMN "database_id" text;--> statement-breakpoint
+UPDATE "mcp_query_data"
+SET
+	"sql_query" = "mcp_call_log"."tool_input"->>'sql_query',
+	"database_id" = "mcp_call_log"."tool_input"->>'database_id'
+FROM "mcp_call_log"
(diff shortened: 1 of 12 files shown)


Bl3f:
Hey @btalx, thank you for this PR / raising this issue. We are currently working on something to help users migrate from external tools to nao stories, this issue is a deeply related to it. For the moment I keep it open but we might solve it doing some structural changes tho. So let's see, we will keep you posted.

PR #1432, opened 2026-08-20
feat(charts) add horizontal bar built-in chart type

### LINKED ISSUE #1431: Add progress bar chart to default charts
Lots of people like it so let's add it!

Image

ad4mou:
## Summary

- Adds a **progress bar** chart to the built-in chart types. Each row gets a horizontal bar in a grey track, with the category on the left and the value in a column on the right.
- Bars are sized against the largest value in the data. Takes one column of values; the agent is told to sort and limit rows in SQL.
- Rebuilt on our own chart library instead of copying the example custom chart, so it works everywhere the other charts do: image export, stories, and the chart editor. The example file stays as the demo for project-defined charts.
- Values show by default and can be switched off in the chart editor.
- Small cleanup: the muted text colour, label size, and label gap were duplicated across the chart code and are now defined once.

Closes #1431

CleanShot 2026-08-21 at 18 33 43@2x
CleanShot 2026-08-21 at 18 33 57@2x
CleanShot 2026-08-21 at 18 32 28@2x
CleanShot 2026-08-21 at 18 34 18@2x
CleanShot 2026-08-21 at 18 34 38@2x

<a href="https://cubic.dev/pr/getnao/nao/pull/1432?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/agents/tools/display-chart.ts
+++ b/apps/backend/src/agents/tools/display-chart.ts
@@ -58,11 +58,11 @@ export default createTool<displayChart.Input, displayChart.Output>({
 			};
 		}
 
-		if (chartType === 'progress_bar' && series.length !== 1) {
+		if (chartType === 'horizontal_bar' && series.length !== 1) {
 			return {
 				_version: '1',
 				success: false,
-				error: 'Progress bar charts require exactly one series.',
+				error: 'Horizontal bar charts require exactly one series.',
 			};
 		}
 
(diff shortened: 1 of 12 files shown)


Bl3f:
Not sure we should call it "progress_bar" maybe it's horizontal bar chart that is most common, even if there is a nothing of progress compared to 100% it's still an horizontal bar chart but 100% or not

ad4mou:
Oh ok I see, fair enough

ad4mou:
LGTM

PR #1466, opened 2026-08-25
feat(mattermost): add Mattermost connector

### LINKED ISSUE #1090: [feature] Mattermost connector / bot
## Problem
nao supports Slack, Teams, Telegram, and WhatsApp as messaging channels, but not Mattermost. Some teams run Mattermost as their internal chat and cannot use the Slack bot.

A user asked for a Mattermost bot so their colleagues can query nao directly from Mattermost, the same way the Slack integration works.

## Expected behavior
Add a Mattermost connector / bot so users can chat with the nao agent from Mattermost (message in, agent answers), consistent with the existing Slack / Teams integrations.

## Customer ID

ad4mou:
## Summary
- Adds Mattermost as a messaging connector, alongside Slack, Teams, Telegram and WhatsApp. Configure a server URL and bot token per project in Settings → Project → Mattermost; the bot holds an outbound connection per project, so no inbound webhook is needed.
- Replies in the channel rather than a thread, matching Slack: one message in the channel, follow-up housekeeping in a thread. DMs stay unthreaded.
- Answers mentions in channels, and everything in DMs. Follow-ups in an existing conversation don't need a repeat mention.
- People are linked to their nao account automatically by their Mattermost email
- Feedback is given by reacting 👍/👎 on an answer. Reactions arrive over the existing connection, so no network access back to nao is needed.
- A Stop button appears while nao is answering. This needs Mattermost to reach nao over HTTP, so it's off by default with its own optional Callback URL setting, since the address Mattermost uses can differ from the one people use in a browser. Callbacks are verified with a signed token.

## Out of scope

- Slash commands, file uploads and modals.

Closes #1090

<a href="https://cubic.dev/pr/getnao/nao/pull/1466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/src/services/mattermost.ts
```diff
@@ -0,0 +1,962 @@
...
+		if (!attached) {
+			logger.warn('Could not associate Mattermost answer post with assistant message', {
+				source: 'system',
+				projectId: this._config.projectId,
+			});
+			return;
+		}
+		const results = await Promise.allSettled([
+			this._adapter.addReaction(ctx.thread.id, ctx.convMessage.id, '+1'),
+			this._adapter.addReaction(ctx.thread.id, ctx.convMessage.id, '-1'),
```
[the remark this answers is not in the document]
ad4mou: not true, reactions do work

apps/backend/src/db/sqlite-schema.ts
```diff
@@ -319,6 +328,7 @@ export const chatMessage = sqliteTable(
 		supersededAt: integer('superseded_at', { mode: 'timestamp_ms' }),
 		versionGroupId: text('version_group_id'),
 		source: text('source', { enum: MESSAGE_SOURCES }),
+		mattermostPostId: text('mattermost_post_id'),
```
Bl3f, line 331: why this here?

ad4mou: This links a Mattermost answer thumbs up/down to its nao message so reaction feedback is saved against the correct message.

Basically, buttons look bad on mattermost, so I wanted to avoid having 2 for thumbs up/down which is why I found a way to store feedback via reactions on the message

apps/backend/src/services/mattermost-helpers.ts
```diff
@@ -0,0 +1,450 @@
...
+
+export function cacheMattermostEmail(
+	emailCache: MattermostEmailCache,
+	userId: string,
+	email: string | null,
+	now = Date.now(),
+): void {
+	const normalizedEmail = email ? email.toLowerCase() : null;
+	emailCache.set(userId, {
+		email: normalizedEmail,
```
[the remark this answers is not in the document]
ad4mou: That is such an edge case, I'll ignore it

apps/backend/src/db/sqlite-schema.ts
Bl3f: I don't think we should do an exception in the data model to put mattermost in the chatMessage table

ad4mou: oh yeah fair enough, but so should we just not have reactions in Mattermost or use the very bad looking buttons?

Bl3f: i did not see the bad looking buttons this morning :/

apps/backend/src/app.ts
```diff
@@ -400,6 +406,7 @@ export const startServer = async (opts: { port: number; host: string }) => {
 
 	void pingLicensesServer();
 	void slackService.startSocketModeForAllProjects();
+	void mattermostService.startForAllProjects();
```
Bl3f, line 409: Is it really necessary? shouldn't it be a reactive route instead?

apps/backend/src/queries/feedback.queries.ts
```diff
@@ -21,6 +21,13 @@ export const upsertFeedback = async (feedback: NewMessageFeedback): Promise<Mess
 	return result;
 };
 
+export const deleteFeedbackVote = async (messageId: string, vote: 'up' | 'down'): Promise<void> => {
```
Bl3f, line 24: How is it handle elsewhere? Why do we need a specific function for this here but not elsewhere in the code?

Bl3f (commented):
a few comments

apps/backend/src/app.ts
ad4mou: Mattermost’s outgoing webhooks do not provide the full event flow we need, especially DMs, mentions, threads, and reactions.

So we use its WebSocket API. Unlike an HTTP route, that connection must be restored after every backend restart, which is why `startForAllProjects()` is necessary. Slack does the same for projects using Socket Mode.

apps/backend/src/queries/feedback.queries.ts
ad4mou: yeah this is mainly because we used reactions instead of buttons for thumbs up/down.
I'll try to put it back as buttons

ad4mou:
So I kept `👍/👎` as native Mattermost reactions.
the fix is: 
Mattermost sends `👍/👎` reactions to nao through the existing WebSocket. Because each Mattermost post includes its corresponding nao message ID, nao can save the vote directly in the existing feedback table. 
This removes the need for a `new chat_message` column or migration, while keeping the Stop button unchanged.

ad4mou:
LGTM

Issue #1008, opened 2026-06-30
[feature] add "deselect all" button in Chats Replay when filtering by user


florian-drouet:
Current's behavior in Chats Replay user filtering is to select all users by default. However, when you want to filter on a single user, you have to click on one user and then deselect all users except the one you want filter on one by one.

A quick fix would be to add a "Deselect All" button when clicking on the Users filter and then select one (or more) users that you want to focus on. Ideally there should be an "apply" button to actually apply the filter (if not, the list could update automatically and this behavior can be frustrating). There should also be a text field with fuzzy completion to make the search more user-friendly.

I'd be happy to take the issue and open a PR on it !

Quick video to show you the current behavior and why it's painful ⬇️ 

https://github.com/user-attachments/assets/7b88dc85-d517-4998-8487-c443963100e4

Bl3f:
Ah yes, we can do this obv!

Bl3f:
lgtm

PR #1510, opened 2026-08-28
feat: improve user filter on Usage costs and replay section

florian-drouet:
<a href="https://cubic.dev/pr/getnao/nao/pull/1510?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/settings/usage-filters.tsx
+++ b/apps/frontend/src/components/settings/usage-filters.tsx
@@ -1,3 +1,4 @@
+import { useState } from 'react';
 import { Radio, ThumbsUp, Users, Wrench } from 'lucide-react';
 import { CHAT_REPLAY_FEEDBACK_STATES, CHAT_REPLAY_TOOL_STATES, providerLabel } from '@nao/shared/types';
 import { USAGE_SOURCES } from '@nao/backend/usage';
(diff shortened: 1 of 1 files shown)


florian-drouet:
The issue this PR aims to solve is https://github.com/getnao/nao/issues/1008

Here is a quick video of the user filter now :
https://github.com/user-attachments/assets/b3acb963-97db-4656-aee0-4c1f82324d72

Bl3f:
Thanks for the PR @florian-drouet ; I thought with the current behaviour we would have fixed the issue as when you select an item it only selects it when everything is selected. 

Nonetheless what you did is clearer, the only thing that I would change is the checkbox, id make it a bit more a checkbox with a tick inside so people understand better. 

Thanks!

florian-drouet:
@Bl3f the current behaviour was much better indeed but I liked the one-click possibility to select/deselect all users and the "Apply" button to filter only when needed (rather than on the go). Is it okay for you ?

I agree the checkbox was a bit weird, I've made it back to a single tick as you can see in the screenshot so that it fits the rest of nao's design.

image

Bl3f:
Yeah work well for me this way! Thank you

Bl3f (approved):
LGTM

PR #1525, opened 2026-09-01
Fix Chat not found issue when browser loses network.

### LINKED ISSUE #1491: [bug] Network/auth failure in chat surfaces as "Chat not found" instead of a connectivity error
## Problem

When the browser loses network connectivity (in this case a VPN disconnect), opening an existing chat from the left nav renders the "Chat not found" empty state:

> Chat not found
> This chat may have been deleted, moved, or you may not have access to it.
> [Start a new chat]

The chat still exists and comes back after reconnecting, but the message tells the user their chat was deleted or that they lost access to it. Reported twice by the same user, who assumed data loss both times. A full page reload correctly showed the browser's "no network" page, which is what made the real cause obvious.

Root cause is most likely that the chat fetch does not distinguish a failed request (network error, 401/403 from an expired session, 5xx) from a successful response saying the chat does not exist. Everything that is not a successful fetch falls through to the not-found state.

## Expected behavior

- Distinguish transport/auth failures from a genuine 404 in the chat fetch.
- On a network failure, show a connectivity error with a retry action, for example "Can't reach nao. Check your connection." plus a Retry button, and do not offer "Start a new chat".
- On a 401/403, show a session/auth error and prompt to re-authenticate.
- Only show "Chat not found" when the server actually returns 404 for that chat id.
- Ideally recover automatically once connectivity returns, instead of leaving the user on a dead-end state.

## Customer ID

2b737d12-b3f3-47af-bb91-07863bb2ee26

thenry42:
Creates a better distinction between different chat errors when the network is suddenly lost. It used to display "Chat not found. This chat may have been deleted".

- UNAUTHORIZED Error (401)
- INTERNAL SERVER ERROR (500)

The point is to prevent the user from assuming data loss when the issue is only front-end related. With those error codes, the user can click a retry button to reload the page.

Closes #1491 

<a href="https://cubic.dev/pr/getnao/nao/pull/1525?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/chat-access-error.tsx
+++ b/apps/frontend/src/components/chat-access-error.tsx
@@ -4,12 +4,7 @@ import { Lock, SearchX, TriangleAlert, WifiOff } from 'lucide-react';
 import { MobileHeader } from '@/components/mobile-header';
 import { Button } from '@/components/ui/button';
 import { getSafeRedirectPath } from '@/lib/safe-redirect';
-import {
-	isForbiddenError,
-	isInternalServerError,
-	isNotFoundError,
-	isUnauthorizedError,
-} from '@/lib/trpc-error';
+import { isForbiddenError, isInternalServerError, isNotFoundError, isUnauthorizedError } from '@/lib/trpc-error';
 
 export function ChatAccessError({
 	error,
--- a/apps/frontend/src/components/chat-access-error.tsx
+++ b/apps/frontend/src/components/chat-access-error.tsx
@@ -1,0 +1,118 @@
+import { Link } from '@tanstack/react-router';
+import { Lock, SearchX, TriangleAlert, WifiOff } from 'lucide-react';
+
+import { MobileHeader } from '@/components/mobile-header';
+import { Button } from '@/components/ui/button';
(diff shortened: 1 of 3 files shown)


apps/frontend/src/components/chat-access-error.tsx
```diff
@@ -0,0 +1,113 @@
...
+			>
+				<Button asChild variant='secondary'>
+					<Link to='/'>Go home</Link>
+				</Button>
+			</ChatErrorLayout>
+		);
+	}
+
+	const isServerError = isInternalServerError(error);
+	return (
```
Bl3f, line 75: Could have been also a if like the previous components so it's easier to read and spot the ternary below.

Bl3f (approved):
LGTM, just a small comment / change.

PR #1527, opened 2026-09-01
fix/view sql in story even if env filter var is false 1505 + readme update

### LINKED ISSUE #1505: [bug] SQL-behind-chart button is off if the beta_story_filter_enabled is false
## Problem

The "see the SQL behind a chart" button in story **view** mode (shipped in #395, next to the edit/pen button) does not appear if the var beta_story_filter_enabled = false

## Expected behavior

This behaviour should always be on, and not depend on another var env

## Customer ID
2de47978-b331-41be-8e75-55a9b9a7334b

lshiina-figon:
- "see SQL behind chart" button in story view mode is now visible and displays the executed SQL even if the beta_story_filter_enabled var is false in the .env
- SQL in stories no longer depend on the variable and display either way
- readme includes a section that shows how to run nao locally

fixes #1505 

<a href="https://cubic.dev/pr/getnao/nao/pull/1527?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/side-panel/story-preview.tsx
+++ b/apps/frontend/src/components/side-panel/story-preview.tsx
@@ -67,7 +67,10 @@ export const StoryPreview = memo(function StoryPreview({
 
 	const useLiveUnfiltered = isViewingLatest && isNoCacheMode && !storyFilters.hasActiveFilters;
 	const querySqlSource = useMemo(
-		() => (storyFilters.filtersEnabled ? { api: filterApi, selections: storyFilters.debouncedSelections } : null),
+		() => ({ 
+			api: filterApi,
+			selections: storyFilters.filtersEnabled ? storyFilters.debouncedSelections : {},
+		}),
 		[filterApi, storyFilters.filtersEnabled, storyFilters.debouncedSelections],
 	);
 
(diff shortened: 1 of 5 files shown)


apps/frontend/src/components/side-panel/story-preview.tsx
```diff
@@ -67,8 +67,14 @@ export const StoryPreview = memo(function StoryPreview({
 
 	const useLiveUnfiltered = isViewingLatest && isNoCacheMode && !storyFilters.hasActiveFilters;
 	const querySqlSource = useMemo(
-		() => (storyFilters.filtersEnabled ? { api: filterApi, selections: storyFilters.debouncedSelections } : null),
-		[filterApi, storyFilters.filtersEnabled, storyFilters.debouncedSelections],
+		() =>
+			filtersEnabled
+				? {
+						api: filterApi,
+						selections: storyFilters.filtersEnabled ? storyFilters.debouncedSelections : {},
```
Bl3f: not sure why we need a ternary here if we are in the first part of the higher ternary that is already filtersEnabled?

Bl3f (approved):
Small comment, apart from this 👍

apps/frontend/src/components/side-panel/story-preview.tsx
lshiina-figon: you're right it is redundant, i probably forgot to optimize it during my latest commit, its removed!

PR #1532, opened 2026-09-01
feat(stories): offer live mode after a story is created

### LINKED ISSUE #1471: [feature] Prompt to turn a new story into a live story right after creation
## Problem

Live stories are not discoverable. Nothing in the story creation flow tells a user that a story can refresh on its own, so stories default to being one-off static snapshots and the live capability goes largely unused.

## Expected behavior

- Once a story is created, a popup appears offering to turn it into a live story.
- The popup is the moment of discovery: it should make clear what a live story is, not just present a toggle.
- Accepting converts the story to live directly from the popup, without sending the user to settings.
- Declining is non blocking, and the story can still be made live later from the story itself.

omjeem:
Closes #1471

Live stories only surfaced through the small toggle in the story header, so unless you already knew the feature existed you'd never find it. This adds the offer at the point where it makes sense — right after the agent builds you a story.

I put the card in the existing `ChatInputSuggestions` panel rather than adding a modal, since that panel already does exactly this job for MCP auth, the "create a story" prompt, and conversation feedback. It's non-modal and already collapses when you start typing, which fits the "declining is non blocking" requirement.

**What changed**

- `chat-input-suggestions.tsx` — a `useLiveStorySuggestion` hook plus a fourth suggestion card. Visible when the chat contains a story created by the agent (`action: 'create'`, `output.version === 1`) that isn't live yet and hasn't been dismissed. Dismissal persists in localStorage, keyed `chatId:storySlug`.
- `live-story-settings-dialog.tsx` — a new optional `proposeLive` prop that only affects the dialog's initial state. All copy, controls and save behaviour are unchanged, and the header toggle doesn't pass it, so that path is untouched.
- `use-story-viewer-live-settings.ts` — an optional `enabled` flag (defaults true, so the four existing call sites are unaffected) so the suggestion can reuse the hook's mutation and invalidation set without firing a query in chats with no story.

No backend, schema or migration changes — everything from `updateLiveSettings` down already existed and was simply unreachable unless you found the toggle.

**On the schedule default**

`proposeLive` preselects a daily refresh rather than letting the dialog fall back to "Manual refresh only". With `cacheSchedule: null`, `syncStoryRefreshJob` creates no recurring job and `isCacheExpired` never returns true, so the story fetches once and never refreshes again — live in name only, which is the exact thing the issue is about.

**One deliberate deviation**

The issue says accepting should convert "without sending the user to settings". The discovery and the offer both happen in the card, so the user never navigates anywhere, but accepting does open the live settings dialog rather than converting
silently. I chose that because each scheduled refresh runs real SQL against the user's warehouse and makes an LLM call when `isLiveTextDynamic` is on, and enrolling someone in that without showing the cadence felt wrong. Happy to make
it one-click if you'd rather.

**Testing**

Tested locally against a Postgres project: created a story from chat, accepted the offer, confirmed the story row updates, a recurring `scheduled_job` is created, and the header pill flips on. `npm run lint` passes.

https://github.com/user-attachments/assets/aeb52781-8a75-4655-8517-8fd20a5ab126

<a href="https://cubic.dev/pr/getnao/nao/pull/1532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/chat-input-suggestions.tsx
+++ b/apps/frontend/src/components/chat-input-suggestions.tsx
@@ -1,10 +1,12 @@
 import { useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react';
 import { useMutation } from '@tanstack/react-query';
-import { MessageSquare, X, ThumbsDown, ThumbsUp, Check, Plug } from 'lucide-react';
+import { Activity, MessageSquare, X, ThumbsDown, ThumbsUp, Check, Plug } from 'lucide-react';
 import { NegativeFeedbackDialog } from './chat-messages/assistant-message-actions';
 import { Button } from './ui/button';
 import StoryIcon from './ui/story-icon';
 import type { UIMessage, UIToolPart } from '@nao/backend/chat';
+import { useStoryViewerLiveSettings } from '@/components/side-panel/hooks/use-story-viewer-live-settings';
+import { LiveStorySettingsDialog } from '@/components/side-panel/live-story-settings-dialog';
 import { useAgentContext, useAgentMessages } from '@/contexts/agent.provider';
 import { useChatId } from '@/hooks/use-chat-id';
 import { useInactivityTrigger } from '@/hooks/use-inactivity-trigger';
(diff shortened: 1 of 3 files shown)


Bl3f:
Well it's nice, ty for this contribution! LGTM

Bl3f:
Would you be able pls to fix the conflicts?

omjeem:
Hey @Bl3f, Conflicts resolved and pushed. Should be good to merge now.

omjeem:
Hey @Bl3f , Conflicts resolved again after #1673. Should be good to merge.

PR #1550, opened 2026-09-02
fix(chat): show a recoverable state when a tool output never arrives

### LINKED ISSUE #1546: Tool call keeps showing the running placeholder after a dropped stream
## Problem

When the stream to the tab drops mid-run, the message stops updating but the tool part is left with no output. `execute-sql.tsx`, `execute-python.tsx` and `query-app-db.tsx` render their in-flight placeholder whenever `output` is missing, so an expanded tool block sits on "Executing query..." forever. The tab has already set `isRunning=false`, so nothing will update it, and the only way out is a manual reload — even though the server usually finished the run and persisted the result.

## Root cause

Those three bodies branch on `output ? ... : <placeholder>` and ignore `isSettled` from `ToolCallContext`. `read.tsx`, `grep.tsx` and `read-query-result.tsx` already gate on it.

## Fix

A shared `ToolOutputFallback` that keeps the running label while `!isSettled` and otherwise shows a "no result reached this tab" state with a reload button, used in the three bodies. I have the patch and would like to open the PR.

VaitaR:
Fixes #1546.

The three tool bodies that rendered their in-flight placeholder whenever `output` was missing now use a shared `ToolOutputFallback`, which keeps the running label while the message is unsettled and otherwise says the result never reached this tab and offers a reload. `read.tsx` and `grep.tsx` already worked this way; this just gives the rest the same rule.

Verified with `npm run test -w @nao/frontend -- src/components/tool-calls/tool-output-fallback.test.tsx` (8 passed), and the new tests fail against the unpatched components. `tsc --noEmit`, eslint and prettier are clean on the touched files.

<a href="https://cubic.dev/pr/getnao/nao/pull/1550?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/tool-calls/execute-python.tsx
+++ b/apps/frontend/src/components/tool-calls/execute-python.tsx
@@ -2,6 +2,7 @@ import { useState } from 'react';
 import { Streamdown } from 'streamdown';
 import { Code, Copy, Terminal } from 'lucide-react';
 import { ToolCallWrapper } from './tool-call-wrapper';
+import { ToolOutputFallback } from './tool-output-fallback';
 import type { ToolCallComponentProps } from '.';
 import { useToolCallContext } from '@/contexts/tool-call';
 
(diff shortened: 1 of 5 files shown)


Bl3f (approved):
LGTM

PR #1552, opened 2026-09-02
fix(security): resolve symlinks before accepting a tool path

### LINKED ISSUE #1548: toRealPath accepts symlinks that resolve outside the project folder
## Problem

`toRealPath` (`apps/backend/src/utils/tools.ts`) is the boundary for the agent's `read` and `grep` tools, but it checks containment lexically: `path.resolve` followed by `startsWith(projectFolder + sep)`. A symlink inside the project folder can point anywhere on the host, and it passes — fs follows the link when the file is opened. The `.git`, environment-file and `.naoignore` guards below the check compare the same unresolved path, so they are bypassed the same way.

The reverse case also misbehaves: when the project folder itself is reached through a symlink, a legitimate file inside it can be rejected.

## Fix

Resolve the candidate with `fs.realpathSync` (or its parent, for a file that does not exist yet), resolve the project folder the same way, and re-check containment against the resolved folder before the remaining guards run.

I have the patch plus tests that create real symlinks in a temp dir, and would like to open the PR.

VaitaR:
Fixes #1548.

`toRealPath` re-checks containment against the realpath of the candidate — or of its parent, for a file that does not exist yet — and against the realpath of the project folder, so the `.git`, env-file and `.naoignore` guards below it see the file that will actually be opened.

Verified with `npm run test -w @nao/backend -- tests/tools-symlink-paths.test.ts tests/tools-paths.test.ts` (70 passed) and `tests/sql-file-paths.test.ts` (16 passed). Three of the five new cases fail against the unpatched function. `tsc --noEmit`, eslint and prettier clean.

<a href="https://cubic.dev/pr/getnao/nao/pull/1552?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/utils/tools.ts
+++ b/apps/backend/src/utils/tools.ts
@@ -293,15 +293,35 @@ export const toRealPath = (virtualPath: string, projectFolder: string): string =
 	const relativePath = virtualPath.startsWith('/') ? virtualPath.slice(1) : virtualPath;
 
 	// Resolve and normalize (this handles .. and .)
-	const resolvedPath = path.resolve(normalizedFolder, relativePath);
+	const candidatePath = path.resolve(normalizedFolder, relativePath);
 
-	// Check if path is outside project folder
-	const withinFolder = resolvedPath === normalizedFolder || resolvedPath.startsWith(normalizedFolder + path.sep);
+	const lexicallyWithinFolder =
+		candidatePath === normalizedFolder || candidatePath.startsWith(normalizedFolder + path.sep);
+	if (!lexicallyWithinFolder) {
+		throw new Error(`Access denied: path '${virtualPath}' is outside the project folder`);
+	}
+
+	// The lexical check above is not enough on its own: a symlink inside the project folder can
+	// point anywhere, and fs follows it. Re-check containment against the symlink-resolved path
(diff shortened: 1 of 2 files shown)


Bl3f:
Thanks @VaitaR for finding this and for the patch and tests — the issue is real and the direction was right. This PR got auto-closed by the contributor gate, so I've picked it up in #1682, which keeps your test cases and adds a couple more.

Two things I changed, for the record: returning the realpath from `toRealPath` regressed project folders that are themselves behind a symlink (`toVirtualPath` in `list`/`grep` started throwing, and a not-yet-existing nested path was rejected), and the fallback only handled one missing level, so `/linked-dir/new/file.txt` still resolved lexically. #1682 keeps the lexical return value, runs the guards on both the lexical and the fully resolved path, and lets the context explorer (which has its own symlink policy) opt out. Details in the PR description.

Appreciate the report.

PR #1567, opened 2026-09-02
feat(backend): MCP_ENDPOINT_ENABLED env default for the MCP endpoint

### LINKED ISSUE #1566: MCP endpoint can only be enabled through the settings UI — no deployment default
The MCP endpoint (Settings > MCP Endpoint) ships disabled and the only way to enable it is the admin toggle in the UI. For declaratively-managed self-hosted deployments that's a gap: after a fresh install or a DB reset, every external MCP integration (we connect Dust to nao's endpoint) is broken until someone remembers to click the toggle — everything else about our deployment is reproducible from config.

Proposal: an `MCP_ENDPOINT_ENABLED` env var that supplies the deployment's default for the toggle while a project has no stored MCP settings. Once an admin saves the settings in the UI, the stored value wins and the env var is ignored — so it's a first-boot default, not an override, and the UI keeps full control.

PR is ready, I'd like to submit it.

afranzi:
Closes #1566.

Adds an `MCP_ENDPOINT_ENABLED` env var (`true`/`false`, validated) that acts as the deployment's default for the Settings > MCP Endpoint toggle:

- `getMcpEndpointSettings` falls back to the env-derived default only when the project has **no stored settings** — once an admin saves anything in the UI, the stored value wins and the env var is ignored. First-boot default, not an override.
- Applies everywhere the settings are read, including the `/mcp` route guard, so a fresh deployment comes up with external MCP clients able to connect.
- The mode toggles (sub-agent / context-layer) keep their built-in defaults (both on).
- Documented in `.env.example`.

Tested:
- new `tests/mcp-endpoint-env.test.ts` (5 tests): built-in default preserved when unset, enable/disable defaults for settings-less projects, admin-saved settings never overridden, invalid values rejected at parse time
- backend lint (`tsc --noEmit && eslint`) and prettier clean
- our production use case is exactly this: we connect dust.tt to nao's MCP endpoint on EKS, and today a reinstall breaks that integration until the toggle is clicked again

This PR was written using Claude Fable 5 (claude-fable-5).

<a href="https://cubic.dev/pr/getnao/nao/pull/1567?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -125,6 +125,12 @@ NAO_DEFAULT_PROJECT_PATH=/Users/blef/Work/naolabs/chat/example
 # every backend; a bigger file is rejected before anything is stored.
 # NAO_STORAGE_MAX_FILE_SIZE_MB=10
 
+# Default for the MCP endpoint toggle (Settings > MCP Endpoint) while a project
+# has no stored settings — lets a fresh deployment come up with external MCP
+# clients already able to connect. Once an admin saves the settings in the UI,
+# the stored value wins and this is ignored.
+# MCP_ENDPOINT_ENABLED=true
+
 # Enterprise license — signed token issued by getnao.
 # Accepts either the raw token or a path to a file containing it.
 # When unset, the app runs in OSS mode and all EE features (SSO, …) stay off.
--- a/apps/backend/src/env.ts
+++ b/apps/backend/src/env.ts
@@ -166,6 +166,16 @@ const envSchema = z.object({
 		.optional()
 		.transform((val) => val?.trim() || undefined),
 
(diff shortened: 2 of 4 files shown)


Bl3f (approved):
LGTM

PR #1576, opened 2026-09-03
fix(mcp): evict stuck ask_nao runs from in-memory registry

### LINKED ISSUE #1575: Memory leak: ask_nao runs stuck in 'running' state are never evicted from in-memory registry
## Problem

Production deployments of nao (2 replicas, `deploy/overlays/production/rke2-data-deliv`, image `getnao/nao:0.3.6`) are hitting `OOMKilled` (exit code 137) repeatedly under normal traffic. Both pods restarted 4 times in a ~1 hour window.

Datadog `container.memory.working_set` for both pods shows a clear sawtooth: memory climbs steadily for hours (e.g. ~625MB baseline up to ~1GB), then drops sharply back down exactly when `kubernetes.containers.restarts` increments — i.e. the process is restarted, not that usage plateaus. This is a memory-leak signature, not a static under-provisioning issue. It also correlates with load: after one pod died, the survivor absorbed its traffic and OOM'd roughly 6x faster (~3 minutes vs ~20 minutes to hit the same ceiling).

## Suspected root cause

`apps/backend/src/mcp/tools/ask-nao-runs.ts` keeps a module-level, per-process `Map` (`runs`) tracking every `ask_nao` background run:

```ts
const runs = new Map<string, AskNaoRunState>();
const FINISHED_RUN_TTL_MS = 30 * 60 * 1000;

setInterval(() => {
  const now = Date.now();
  for (const [chatId, state] of runs) {
    if (state.status !== 'running' && now - state.finishedAt > FINISHED_RUN_TTL_MS) {
      runs.delete(chatId);
    }
  }
}, 5 * 60 * 1000).unref();
```

The sweep only expires entries that have already left the `'running'` state (i.e. `complete` or `error` with `finishedAt` set). Any run that gets stuck in `'running'` forever — for example if `agent.stream()` in `runAskNaoInBackground` (`apps/backend/src/mcp/tools/sub-agent.ts`) hangs indefinitely, or an error occurs in a path that isn't caught by its `try/catch` — is **never removed** from the map. Because the map is in-process memory only, it resets to empty on every restart, which matches the observed sawtooth exactly.

There's a secondary contributor: completed/errored runs are retained for a full 30 minutes holding their entire `AskNaoResult` (`text`, `queries[].preview` row data, `story_ids`). `resolveAnswerPayload` already has a DB-backed fallback (`reconstructAnswerFromDb`) for when an entry is missing from the map, so this 30-minute in-memory retention isn't load-bearing for correctness and mainly adds to peak memory under concurrent load.

## Suggested fix

1. Expire stuck `'running'` entries past a max duration, not just finished ones:
```ts
const MAX_RUN_MS = 15 * 60 * 1000; // well above ASK_NAO_SYNC_BUDGET_MS

setInterval(() => {
  const now = Date.now();
  for (const [chatId, state] of runs) {
    const isStale =
      (state.status !== 'running' && now - state.finishedAt > FINISHED_RUN_TTL_MS) ||
      (state.status === 'running' && now - state.startedAt > MAX_RUN_MS);
    if (isStale) runs.delete(chatId);
  }
}, 5 * 60 * 1000).unref();
```
2. Reduce `FINISHED_RUN_TTL_MS` since `reconstructAnswerFromDb` already covers the miss case.
3. Add a hard cap on `runs.size` (evict oldest) as a backstop against future regressions in the sweep logic.

I'll open a PR with this fix.

## Environment
- nao version: v0.3.6 (also present in v0.3.7–v0.3.10 — this code path hasn't changed since)
- Deployment: Kubernetes, 2 replicas, memory limit 2Gi
- Evidence: Datadog `container.memory.working_set` + `kubernetes.containers.restarts` for both pods over 24h, showing the climb-then-reset pattern aligned with restart counts.

---
This issue was investigated and written with the help of Claude Sonnet 5 (claude-sonnet-5).

meetmehta2601:
## Summary

`askNaoRuns`'s cleanup sweep only expired runs already in `complete`/`error` state (it checks `finishedAt`, which only those states have). A run stuck in `running` forever — a hung `agent.stream()`, or an error thrown outside `runAskNaoInBackground`'s `try/catch` — was invisible to the sweep and stayed in the map for the life of the process.

We saw this in production: `container.memory.working_set` for the `nao` backend climbed for hours, then dropped sharply right when the pod got `OOMKilled` and restarted — repeating on both replicas. That's consistent with an in-memory map that only clears via process restart, not steady-state usage.

This PR:
- Expires `running` entries older than `MAX_RUN_AGE_MS` (15 min, well above the 45s sync budget), in addition to the existing finished-run TTL.
- Shortens the finished-run TTL from 30 to 5 minutes — safe because `resolveAnswerPayload` already falls back to `reconstructAnswerFromDb` when an entry is missing.
- Caps the map at `MAX_TRACKED_RUNS` (5,000) as a backstop against a future regression in the sweep.

Fixes #1575

This PR was written using Claude Sonnet 5 (claude-sonnet-5).

## Test plan
- `npm run lint:backend` (tsc + eslint) passes clean on this change.
- No test coverage existed for `ask-nao-runs.ts` before this change; didn't add any since I couldn't validate the surrounding agent/stream behavior against real services per the contributing guide — happy to add unit tests for the sweep logic itself if that's wanted.
- Manually traced the sweep against all three `AskNaoRunState` variants (`running`, `complete`, `error`) to confirm the new branch is correct.
- No behavior change for the happy path (run completes within the old/new TTL); only affects runs that were previously stuck forever or long-lived past 30 minutes.

### DIFF
--- a/apps/backend/src/mcp/tools/ask-nao-runs.ts
+++ b/apps/backend/src/mcp/tools/ask-nao-runs.ts
@@ -26,7 +26,19 @@ export type AskNaoRunState =
 
 const runs = new Map<string, AskNaoRunState>();
 
-const FINISHED_RUN_TTL_MS = 30 * 60 * 1000;
+// Short: `get_nao_answer` only needs this while a client is actively polling right after
+// `ask_nao` returned `running`. Once a run is done, `resolveAnswerPayload` falls back to
+// `reconstructAnswerFromDb`, so we don't need to hold the full result in memory for long.
+const FINISHED_RUN_TTL_MS = 5 * 60 * 1000;
+
+// Backstop for runs that never leave the `running` state (e.g. a hung stream, or an
+// error thrown outside `runAskNaoInBackground`'s try/catch). Without this, such entries
+// are invisible to the sweep below (it only looks at `finishedAt`, which `running`
+// entries don't have) and accumulate for the lifetime of the process.
+const MAX_RUN_AGE_MS = 15 * 60 * 1000;
+
+// Hard cap so a future bug in the sweep/TTL logic can't reintroduce unbounded growth.
+const MAX_TRACKED_RUNS = 5_000;
 
 /**
  * Tracks `ask_nao` agent runs that outlive their originating MCP request.
@@ -37,6 +49,12 @@ const FINISHED_RUN_TTL_MS = 30 * 60 * 1000;
  */
 export const askNaoRuns = {
 	start(chatId: string): void {
+		if (runs.size >= MAX_TRACKED_RUNS) {
+			const oldestChatId = runs.keys().next().value;
+			if (oldestChatId !== undefined) {
+				runs.delete(oldestChatId);
+			}
+		}
 		runs.set(chatId, { status: 'running', startedAt: Date.now() });
 	},
 	complete(chatId: string, result: AskNaoResult): void {
@@ -54,7 +72,11 @@ setInterval(
 	() => {
 		const now = Date.now();
 		for (const [chatId, state] of runs) {
-			if (state.status !== 'running' && now - state.finishedAt > FINISHED_RUN_TTL_MS) {
+			const isStale =
+				state.status === 'running'
+					? now - state.startedAt > MAX_RUN_AGE_MS
+					: now - state.finishedAt > FINISHED_RUN_TTL_MS;
+			if (isStale) {
 				runs.delete(chatId);
 			}
 		}


Bl3f:
hey @meetmehta2601 I can't reopen your PR because you've recreated it / forced-pushed. 

I'd like to eventually re-open it, just to be sure before, are you using nao in production and did you really encountered this issue?

meetmehta2601:
Hello @Bl3f 
Thanks for quick response sure i will create new clean PR.
Yes I am using it on production and yesterday only we faced this issue.

meetmehta2601:
Opened a clean replacement here since I can't reopen this one: #1577

meetmehta2601:
@Bl3f 
I guess the issue is i dont have any LGTM comment so i cant open a PR :   
This is a pure account-level gate, not something tied to the PR's content or branch history:
  - @Bl3f only asked a clarifying question on #1576 — no lgtm or lgtmi reply yet.
  - Until a maintainer replies with one of those, every new PR from this account will auto-close instantly, regardless of how clean it is.

Bl3f:
I've reopened other PR. Will add you to contributors later (as lgtm works only on issues).

PR #1586, opened 2026-09-03
Feat flexible usage filter 1482

thenry42:
- Add reusable usage period filters with configurable days and granularity
- Persist filters and the selected period per project
- Support creating, editing, selecting, and removing filters
- Add SQLite and PostgreSQL migrations for project preferences
- Limit requests to 2,000 chart buckets and saved filters to 16
- Improve usage filter labels, dialogs, validation, and control sizing
- Default chat replay pagination to 20 rows per page (to avoid odd behavior on the button)
- Add backend and frontend tests

<a href="https://cubic.dev/pr/getnao/nao/pull/1586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/migrations-postgres/meta/0065_snapshot.json
```diff
@@ -0,0 +1,7166 @@
+{
```
Bl3f: As a rule of thumb the best is for a single PR to have only one migration. As it's a PR that you create yourself you can eventually remove the 2 migrations and regenerate a single one so we dont get all the back and forth around migrations.

apps/backend/src/db/pg-schema.ts
```diff
@@ -404,6 +405,28 @@ export const messageFeedback = pgTable('message_feedback', {
 		.notNull(),
 });
 
+export const userProjectPreference = pgTable(
```
Bl3f: Should it be a user preference or a project preference, I mean when a user create a filter should every other admin have access to the shortcut or not? @ClaireGz wdyt

apps/backend/src/trpc/usage.routes.ts
```diff
@@ -14,4 +42,157 @@ export const usageRoutes = {
...
+				},
+			);
+			return preferences.usagePeriod;
+		}),
+
+	createPeriodEntry: adminProtectedProcedure.input(createPeriodEntryInputSchema).mutation(async ({ ctx, input }) => {
+		assertPreferenceProject(input.projectId, ctx.project.id);
+		const entry = { id: crypto.randomUUID(), ...input.entry };
+		await userProjectPreferenceQueries.mutateUserProjectPreferences(ctx.user.id, ctx.project.id, (current) => {
+			const sanitizedCurrent = sanitizePeriodPreferences(current).preferences;
```
Bl3f: A lot of the code in each route looks repeated with small changes, maybe there is some factorisation possible? (ask Fable eventually)

apps/backend/src/trpc/usage.routes.ts
```diff
@@ -14,4 +42,157 @@ export const usageRoutes = {
...
+	getPeriodSettings: adminProtectedProcedure.input(projectPreferenceInputSchema).query(async ({ ctx, input }) => {
+		assertPreferenceProject(input.projectId, ctx.project.id);
+		const preferences = await getSanitizedPeriodPreferences(ctx.user.id, ctx.project.id);
+		return {
+			preference: preferences.usagePeriod ?? null,
+			entries: parsePeriodEntries(preferences),
+		};
+	}),
+
+	updatePeriodPreference: adminProtectedProcedure
```
Bl3f: Not sure I one look I see what's the difference between PeriodPreferences and PeriodEntry

apps/backend/src/trpc/usage.routes.ts
```diff
@@ -14,4 +42,157 @@ export const usageRoutes = {
...
+	}),
 };
+
+function parsePeriodEntries(preferences: UserProjectPreferences) {
+	if (!Array.isArray(preferences.usagePeriodEntries)) {
+		return [];
+	}
+	return preferences.usagePeriodEntries
+		.flatMap((entry) => {
+			const parsed = usagePeriodEntrySchema.safeParse(entry);
```
Bl3f: Not sure also why we need to parse stuff, can't it be already objects?

Bl3f (commented):
A few visual feedbacks and comments already so you can start working on it (didn't look the frontend yet)

* Would increase the width so we can read the whole filter
Image

* In the dropdown, i'd like to have cursor:pointer on the items of the dropdown and not the default cursor
*

apps/backend/src/trpc/usage.routes.ts
thenry42: They are indeed already objects but they need to be validated and cleaned, it is a naming issue. Fixed in next commit.

apps/backend/src/trpc/usage.routes.ts
thenry42: naming issue as well. 
PeriodPreference -> PeriodSelection
PeriodEntry -> Saved Period

apps/backend/src/db/pg-schema.ts
thenry42: Should be a user preference according to @ClaireGz.

apps/backend/src/db/sqlite-schema.ts
```diff
@@ -426,6 +427,30 @@ export const messageFeedback = sqliteTable('message_feedback', {
 		.notNull(),
 });
 
+export const userProjectPreference = sqliteTable(
+	'user_project_preference',
```
Bl3f: could it be possible to put this in userPreferences table instead?

thenry42: Done !

PR #1596, opened 2026-09-03
fix(sync): strip .git/ from local_path repos to fix gitlink bug

### LINKED ISSUE #1595: nao sync: local_path repos still get gitlinked — #720 only fixed the url path
## Problem

`nao sync` supports two ways to bring a repo into `repos/<name>/`: a `url` (git clone) or a `local_path` (plain copy). #716 documented that the `url` path left a `.git/` folder inside `repos/<name>/`, which turns it into a git **gitlink** instead of tracked files — so anyone else who clones the context repo (a teammate, CI, the deployed agent) gets an empty folder instead of the synced content. #720 fixed that for the `url` path by stripping `.git/` after cloning.

`sync_local_repo()` (same file, `local_path` branch) has the identical bug and was never touched by #720. It does a plain `shutil.copytree(source_path, repo_path, dirs_exist_ok=True)` with no `.git` filtering. If `local_path` points at a real git checkout — which it usually does, since that's the whole point of pointing at a sibling repo on disk — its `.git/` gets copied wholesale into `repos/<name>/.git`, and the parent repo silently records `repos/<name>` as a gitlink again.

## How I hit it

`nao_config.yaml`:
```yaml
repos:
- name: data-dbt
  local_path: ../data-dbt
```
After `nao sync` + `git add`, `git show` on that commit prints:
```
-Subproject commit 00774a7...
+Subproject commit bad0501...
```
instead of a real file diff. `git ls-files -s repos/data-dbt` confirms mode `160000` (gitlink), and there's no `.gitmodules` — it's not a real submodule, just git's implicit nested-repo detection kicking in because `.git/` is physically there.

## Why it matters

Same blast radius as #716: whoever clones the context repo after the sync — teammate, CI, deployed agent — gets an empty `repos/<name>/` folder, silently, no error. It just took longer to notice because `local_path` is the less-used path (most setups use `url`), but the bug is the exact same class, just in the sibling code path #720 didn't cover.

## Scope

Mirror #720 in `sync_local_repo()`: exclude `.git/` when copying, in both the filtered and unfiltered branches.

I'd like to implement this — PR incoming.

tclavelloux:
Fixes #1595

## Problem

`sync_local_repo()` had the same gitlink bug #716 reported and #720 fixed — just in the `local_path` branch instead of the `url`/clone branch. It copied `source_path` with `shutil.copytree(..., dirs_exist_ok=True)` and no filtering, so a `.git/` folder in the source got copied straight into `repos/<name>/.git`. Git then treats `repos/<name>/` as a nested repo (gitlink, mode `160000`) instead of tracked files, and anyone who clones the parent repo afterward gets an empty folder.

## Fix

Same idea as #720, applied to the copy path instead of the clone path:

- Unfiltered copy: pass `ignore=shutil.ignore_patterns(".git")` to `shutil.copytree` so `.git/` is never copied in the first place.
- Filtered copy (the manual `rglob` loop): skip any path under `.git/` before the include/exclude check.

Added two tests mirroring the existing `TestSyncLocalRepo` cases — one for the unfiltered branch, one for the filtered branch — both asserting `.git/` doesn't exist in the synced output while regular files do.

`pytest tests/nao_core/commands/sync/test_repository_provider.py` — 48 passed. `ruff check` / `ruff format --check` on both changed files pass (no new warnings; the 5 pre-existing `BLE001`/`SIM102` hits are untouched code).

This PR was written using Claude Sonnet 5 (claude-sonnet-5).

<a href="https://cubic.dev/pr/getnao/nao/pull/1596?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/cli/nao_core/commands/sync/providers/repositories/provider.py
+++ b/cli/nao_core/commands/sync/providers/repositories/provider.py
@@ -151,7 +151,14 @@ def sync_local_repo(repo: RepoConfig, base_path: Path) -> bool:
         has_filters = bool(repo.include or repo.exclude)
 
         if not has_filters:
-            shutil.copytree(source_path, repo_path, dirs_exist_ok=True)
+            # Skip .git/ so a source that happens to be a git checkout doesn't turn
+            # repo_path into a nested repo (which git records as a gitlink, not files).
+            shutil.copytree(
+                source_path,
+                repo_path,
+                dirs_exist_ok=True,
+                ignore=shutil.ignore_patterns(".git"),
+            )
         else:
             repo_path.mkdir(parents=True, exist_ok=True)
             for file_path in source_path.rglob("*"):
@@ -159,6 +166,8 @@ def sync_local_repo(repo: RepoConfig, base_path: Path) -> bool:
                     continue
 
                 relative = PurePosixPath(file_path.relative_to(source_path)).as_posix()
+                if relative == ".git" or relative.startswith(".git/"):
+                    continue
                 if not _matches_patterns(relative, repo.include, repo.exclude):
                     continue
 
--- a/cli/tests/nao_core/commands/sync/test_repository_provider.py
+++ b/cli/tests/nao_core/commands/sync/test_repository_provider.py
@@ -385,6 +385,42 @@ class TestSyncLocalRepo:
         assert (output / "local-repo" / "new.txt").exists()
         assert not (output / "local-repo" / "old.txt").exists()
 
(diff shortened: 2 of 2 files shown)


Bl3f (approved):
LGTM

PR #1604, opened 2026-09-04
Fix autofill search bar bug in new settings page

### LINKED ISSUE #1603: bug: search bar gets autofilled in new settings page

ad4mou:
Just making sure that the search bar cannot be autofilled by a password manager..
And avoiding the env fields from being autofilled, by suggesting they're keys, not passwords

**Note:** I found just setting `AutoComplete=off` wasn't enough, Password manager would still try to fill the search bar, so I made it so that the search bar field can be written to only when focused, i.e. when user is clicked on it, and is typing.

Closes  #1603

<a href="https://cubic.dev/pr/getnao/nao/pull/1604?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/settings/env-vars-section.tsx
+++ b/apps/frontend/src/components/settings/env-vars-section.tsx
@@ -104,6 +104,7 @@ export function EnvVarsSection({ isAdmin }: { isAdmin: boolean }) {
 								<Input
 									id={inputId}
 									type='password'
+									autoComplete='new-password'
 									value={value}
 									onChange={(e) => handleChange(key, e.target.value)}
 									placeholder='Enter value...'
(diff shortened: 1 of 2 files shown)


apps/frontend/src/components/settings/env-vars-section.tsx
```diff
@@ -104,6 +104,7 @@ export function EnvVarsSection({ isAdmin }: { isAdmin: boolean }) {
 								<Input
 									id={inputId}
 									type='password'
+									autoComplete='new-password'
```
Bl3f, line 107: Should it be a unique id so it does not autocomplete clash for each field?

Bl3f (commented):
Small comment

apps/frontend/src/components/settings/env-vars-section.tsx
ad4mou: no it's actually not an id, just a way to avoid google forcing its autofill

PR #1605, opened 2026-09-04
Fix gemini test error 1258

### LINKED ISSUE #1258: nao test run: intermittent, model-dependent Gemini 500 "function call turn comes immediately after a user turn or after a function response turn"
## Summary

Running `nao test run` against Gemini, a random subset of tests fails with a raw 500 from the Gemini `generateContent` API:

```
Request failed: 500 {"error":"Please ensure that function call turn comes immediately after a user turn or after a function response turn."}
```

The failing test comes back `status=error`, `tokens=0`, `tool_calls=0` — the agent never runs, the request is rejected upfront by Gemini. There is no retry, so a single transient 500 fails the whole test. Same family as #1228 (a `nao test` 500), but here it's a turn-ordering error, not a model/base_url one.

## Environment
- nao: 0.2.11
- Provider: `google` (tested `gemini-2.5-pro` and `gemini-2.5-flash`)

## Steps to reproduce
1. A `tests/` suite + Gemini configured in `nao_config.yaml`.
2. `nao test run -m google:gemini-2.5-pro -t 1`
3. A random handful of tests come back `error` with the 500 above.

## Actual behavior

Random tests fail upfront. Evidence it is transient and infra-side, not test-content:

- Two consecutive runs of the same 28-test suite at `-t 1`: **9 distinct tests** hit the 500, only **1 in common**.
- Re-running the errored tests via `--select` passes some of them **unchanged**.
- Persists identically at `-t 1` → not concurrency.

Strongly **model-dependent** (same suite, same day):

| Model | passing | tests hitting the 500 |
|---|---|---|
| `gemini-2.5-pro` | 14/28 | ~5/28 (random) |
| `gemini-2.5-flash` | 0/28 | **28/28** |

`gemini-2.5-flash` (and `gemini-3.5-flash`, anecdotally) trip the error on essentially every test; `gemini-2.5-pro` only occasionally.

## Cause (hypothesis)

Gemini `contents` validation error: a `functionCall` part appears where it must be preceded by a user turn or a `functionResponse`. Likely tied to how the test agent serializes its message history into the Gemini request (leading function call, consecutive model turns, or reasoning/function-call parts split across turns). Flash models seem to produce turn structures that trip this far more often than pro.

## Suggested fix

1. **Retry on 5xx** (with backoff) in the test runner / LLM client — absorbs the transient case.
2. **Audit the Gemini request builder** for turn ordering: every `functionCall` turn preceded by a user turn or `functionResponse`, never consecutive model turns.

Happy to provide anonymized request payloads if useful.

thenry42:
- Fix Gemini turn-order errors during nao test SQL verification by restoring the original user prompt.
- Apply the workaround only to the Google provider, preserving all other providers’ existing message sequences.
- Add regression coverage for Google and every other supported provider.
- This affects test verification only; normal chat agents are unchanged.

fixes #1258

<a href="https://cubic.dev/pr/getnao/nao/pull/1605?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/agents/provider-meta.ts
+++ b/apps/backend/src/agents/provider-meta.ts
@@ -872,8 +872,8 @@ export function isAnthropicApiModel(provider: LlmProvider, modelId: string): boo
 	}
 }
 
-/** Whether a model uses Google's native Generative AI message transport. */
-export function usesGoogleGenerativeAiApi(provider: LlmProvider, modelId: string): boolean {
+/** Whether the selected provider/model adapter requires Gemini's user-first alternating turn sequence. */
+export function requiresGeminiTurnSequence(provider: LlmProvider, modelId: string): boolean {
 	switch (providerKind(provider)) {
 		case 'google':
 			return true;
--- a/apps/backend/src/agents/provider-meta.ts
+++ b/apps/backend/src/agents/provider-meta.ts
@@ -872,6 +872,18 @@ export function isAnthropicApiModel(provider: LlmProvider, modelId: string): boo
 	}
 }
 
+/** Whether a model uses Google's native Generative AI message transport. */
+export function usesGoogleGenerativeAiApi(provider: LlmProvider, modelId: string): boolean {
+	switch (providerKind(provider)) {
+		case 'google':
+			return true;
+		case 'vertex':
+			return !modelId.startsWith('claude-');
+		default:
+			return false;
+	}
+}
+
 /**
  * Resolve the tunable-parameter capabilities for a model. Falls back to sensible per-provider
  * defaults for custom (unlisted) models so tuning still works for user-added model ids.
--- a/apps/backend/src/agents/provider-meta.ts
+++ b/apps/backend/src/agents/provider-meta.ts
@@ -872,18 +872,6 @@ export function isAnthropicApiModel(provider: LlmProvider, modelId: string): boo
 	}
 }
 
-/** Whether the selected provider/model adapter requires Gemini's user-first alternating turn sequence. */
-export function requiresGeminiTurnSequence(provider: LlmProvider, modelId: string): boolean {
-	switch (providerKind(provider)) {
-		case 'google':
-			return true;
-		case 'vertex':
-			return !modelId.startsWith('claude-');
-		default:
-			return false;
-	}
-}
-
 /**
  * Resolve the tunable-parameter capabilities for a model. Falls back to sensible per-provider
  * defaults for custom (unlisted) models so tuning still works for user-added model ids.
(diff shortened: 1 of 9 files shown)


apps/backend/src/agents/provider-meta.ts
```diff
@@ -872,6 +872,18 @@ export function isAnthropicApiModel(provider: LlmProvider, modelId: string): boo
 	}
 }
 
+/** Whether the selected provider/model adapter requires Gemini's user-first alternating turn sequence. */
+export function requiresGeminiTurnSequence(provider: LlmProvider, modelId: string): boolean {
+	switch (providerKind(provider)) {
+		case 'google':
+			return true;
+		case 'vertex':
+			return !modelId.startsWith('claude-');
```
Bl3f: in vertex we can include other kind of model than anthropic ones

Bl3f (commented):
small comment, except from this lgtm

apps/backend/src/agents/providers.ts
```diff
@@ -225,6 +228,10 @@ function createModel(provider: LlmProvider, settings: ProviderSettings, modelId:
 		: LLM_PROVIDERS[providerKind(provider)].create(settings, modelId);
 }
 
+function vertexAdapter(modelId: string): 'anthropic' | 'google' {
+	return modelId.startsWith('claude-') ? 'anthropic' : 'google';
```
Bl3f, line 232: Not sure about this because actually on vertex you can have models other than google or anthropic, so i feel it's not the right switch to do.

Bl3f (changes requested):

apps/backend/src/agents/providers.ts
thenry42: Completely abandonned the google adapter approach and went for a different message sequence for all providers. Small fixe, works everywhere but tiny increase of token usage for other providers, only in nao test.

apps/backend/src/services/test-agent-verification.ts
```diff
@@ -0,0 +1,34 @@
...
+		{ role: 'user', content: buildVerificationPrompt(expectedColumns, queryResults) },
+	];
+}
+
+function buildVerificationPrompt(columns: string[], queryResults: Map<string, QueryResult>): string {
+	const tables = [...queryResults.entries()]
+		.map(([queryId, result]) => `- ${queryId} (${result.data.length} rows): ${result.columns.join(', ')}`)
+		.join('\n');
+
+	return `Based on your previous analysis, write a DuckDB query returning the final answer to the original question.
```
Bl3f, line 23: Add in this system prompt the fact that the user prompt has been added so it adds a bit more context to the AI.

thenry42: done

PR #1606, opened 2026-09-04
Fix/slack split bug 1536

### LINKED ISSUE #1536: Slack section chunking splits inside fenced code blocks, breaking rendering
# Slack section chunking splits inside fenced code blocks, breaking rendering

## Summary

`createTextBlocks` in `apps/backend/src/utils/messaging-provider.ts` now chunks an
oversized text segment into multiple Block Kit sections (`chunkSlackText` at
`SLACK_SECTION_TEXT_MAX_CHARS = 2900`). The chunker is not aware of fenced code blocks:
when a single ```` ``` ```` block is longer than the section limit, the split lands inside
the fence. The first section is left with an unterminated opening fence, the middle
sections carry code with no fence at all, and the final section holds a stray closing
fence. In Slack the code renders as plain prose rather than a code block.

## Environment

- Image: `getnao/nao@sha256:312141eae21e1412d411b8d71a6a3d801e786da400ea67edf6d3991f0639aa28`
- File: `apps/backend/src/utils/messaging-provider.ts`
- Functions: `createTextBlocks` -> `chunkSlackText` -> `findSlackChunkBreak`

## Reproduction

Call `createTextBlocks` with a single fenced code block whose body exceeds 2900 characters:

```ts
import { createTextBlocks } from './messaging-provider.ts';

const body = Array.from({ length: 220 }, (_, i) =>
  `select col_${i}, sum(amount) as total_${i} from very_long_table_name_${i} group by 1`
).join('\n');
const input = `Here is the query:\n\n\`\`\`sql\n${body}\n\`\`\`\n\nThat is the answer.`;

const blocks = createTextBlocks(input); // run under the image's bun runtime
```

Counting fence markers (```` ``` ````) per resulting text block:

| Block | fence markers | state |
|------:|--------------:|-------|
| 0 | 1 | opens ```` ```sql ````, never closed |
| 1-5 | 0 | raw code, no fence -> Slack renders as plain text |
| 6 | 1 | stray closing ```` ``` ```` |

Every middle section is emitted outside any code block.

## Root cause

`findSlackChunkBreak` chooses a break position by scanning backwards for `\n\n`, then
`\n`, then a space. It has no notion of whether that position sits inside a fenced code
block, so a boundary inside a ```` ``` ```` block cuts the fence open.

## Expected behaviour

When a chunk boundary falls inside a fenced code block, the chunker should close the fence
at the end of the current section and reopen it at the start of the next, so each section
is independently well formed. The reopening marker should be the bare fence (```` ``` ````)
rather than the original info string, because Slack mrkdwn has no language support and would
render the info string as a literal line.

With that behaviour the same input produces seven sections that are each fence balanced and
render as continuous code.

## Impact

This is the "long answer freezes / renders as prose" failure for any response containing a
code block over roughly 2900 characters, which is routine for generated SQL. It is the
reason we currently overlay a forked `messaging-provider.ts` in our deployment; the fork's
only divergence from upstream is fence-aware section chunking. A fix here lets us drop the
fork.

## Suggested fix

Make the section chunker fence-aware: track open-fence state while walking the text, and
when a boundary lands inside a fence, append the closing marker to the section being emitted
and prepend the reopening bare marker to the next. Inline code spans should likewise not be
split mid-span.

## Related edges (lower priority, same area)

- The per-call section cap bounds one `createTextBlocks` call, not the whole message. A
  heavy multi-tool answer accumulates tool-call, image, and multiple text regions and can
  still exceed Slack's 50-block message limit.
- A second table in one message, or an oversized table, can still exceed the 3000-character
  section limit if it falls back to an unbounded text section rather than a native table.

ad4mou:
## Summary
- Keep long code blocks formatted when Slack splits an answer.
- Keep unfinished code formatted while generating and after stopping. (Add ``` at the end and add code before it so it streams formatted)
- Preserve indentation and surrounding text.
- Add a closing ``` if agent is stopped mid-streaming code

Closes #1536

<a href="https://cubic.dev/pr/getnao/nao/pull/1606?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/services/slack.ts
+++ b/apps/backend/src/services/slack.ts
@@ -1432,6 +1432,7 @@ class ProjectSlackBot {
 				blockIndex: run.blockIndex,
 				blockCount: run.blockCount,
 				blocks: createTextBlocks(streamState.latestSourceText.slice(run.sourceStart, run.sourceEnd), {
+					balanceIncompleteCodeFence: activeStream.stopRequested,
 					truncation: { kind: 'link', url: chatUrl },
 					tableState,
 				}),
@@ -1447,6 +1448,7 @@ class ProjectSlackBot {
 				blockIndex: ctx.textBlockIndex === -1 ? emptyRunBlockIndex : ctx.textBlockIndex,
 				blockCount: ctx.textBlockIndex === -1 ? 0 : ctx.textBlockCount,
 				blocks: createTextBlocks(openRunText, {
+					balanceIncompleteCodeFence: activeStream.stopRequested,
 					truncation: { kind: 'link', url: chatUrl },
 					tableState,
 				}),
--- a/apps/backend/src/services/slack.ts
+++ b/apps/backend/src/services/slack.ts
@@ -1479,7 +1479,11 @@ class ProjectSlackBot {
 		streamState.latestSourceText = stripAssistantTags(text);
 		const visibleText = streamState.latestSourceText.slice(streamState.textRunStart);
 		const tableState = { ...streamState.tableStateAtRunStart };
-		const blocks = createTextBlocks(visibleText, { ...options, tableState });
+		const blocks = createTextBlocks(visibleText, {
+			...options,
+			balanceIncompleteCodeFence: true,
+			tableState,
+		});
 		streamState.tableStateAtRunEnd = tableState;
 		if (blocks.length === 0) {
 			if (visibleText.trim() && streamState.emptyRunBlockIndex === -1) {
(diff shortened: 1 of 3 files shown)


ad4mou:
LGTM, 
it's just logic to split message efficiently and format it whilst streaming

Bl3f (approved):
LGTM

Issue #1554, opened 2026-09-02
Donut / pie chart labels of legend overflows


ClaireGz:
Issue

<img width="554" height="519" alt="Image" src="https://github.com/user-attachments/assets/41e6aafd-3cc1-4a66-bde5-89e17d28090e" />

PR #1612, opened 2026-09-05
fix(frontend): prevent donut chart legend overflow

Khizarshah01:
Fixe #1554 

The Donut/Pie chart legend can move downward after switching between **chat tabs or between view data and view chart etc.** each time the chart becomes visible again, the legend may be positioned lower than before.

The issue was caused by Recharts recalculating the `verticalAlign="middle"` position while the chart was temporarily hidden and had no usable height. 

### Changes: 
Changed `verticalAlign` to `top` to avoid that calculation and keep the legend position stable.

### Before

https://github.com/user-attachments/assets/d7e96bb1-21a6-4867-9f73-ce3ba51d2b4c

### After (fixed)

https://github.com/user-attachments/assets/2faeb4f9-de1b-4dbf-812b-8b006b0f6351

Tested across different screen resolutions. no visual or layout issues.

<a href="https://cubic.dev/pr/getnao/nao/pull/1612?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/frontend/src/components/tool-calls/display-chart.tsx
+++ b/apps/frontend/src/components/tool-calls/display-chart.tsx
@@ -686,7 +686,7 @@ export const ChartDisplay = memo(function ChartDisplay({
 							payload={legendPayload}
 							layout={isPie && !compactPieLegend ? 'vertical' : 'horizontal'}
 							align={isPie && !compactPieLegend ? 'right' : 'center'}
-							verticalAlign={isPie && !compactPieLegend ? 'middle' : 'bottom'}
+							verticalAlign={isPie && !compactPieLegend ? 'top' : 'bottom'}
 							content={
 								<ChartLegendContent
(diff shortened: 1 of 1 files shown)


Bl3f:
Not sure it will really fix the issue long term. I'd say weave to think how to position better legend.

Khizarshah01:
I agree that, and i tested a few different positioning options and found this one looks better:

**Horizontal + centered + bottom**
Screenshot 2026-09-07 at 12 21 55 AM

PR #1621, opened 2026-09-07
fix(backend): verify MCP access tokens against the local JWKS

### LINKED ISSUE #1620: MCP token verification fetches JWKS from the external issuer URL, breaking self-hosted split-horizon deployments
### Summary
`verifyOAuthAccessToken` (apps/backend/src/auth.ts) verifies MCP bearer tokens via better-auth's `verifyAccessToken`, passing `jwksUrl = ${BETTER_AUTH_URL}/api/auth/jwks`. That makes the server fetch its **own** JWKS over its **external** issuer URL.

In a self-hosted split-horizon deployment the issuer host isn't resolvable from the server's own network — e.g. the UI/`BETTER_AUTH_URL` host is VPN-only or lives only in a public DNS zone that the app's VPC/private DNS shadows. The JWKS fetch then fails, better-auth swallows the error (`TypeError`/`JWSInvalid` are caught and dropped in `@better-auth/core` `verifyAccessToken`), `payload` stays undefined, and **every** MCP token is rejected with `no token payload` — even a freshly minted, structurally valid one.

### Symptom
- `POST /api/auth/oauth2/token` → `200` with a valid EdDSA JWT (correct `aud`, `iss`).
- Client sends it to `POST /mcp` → `[WARN] MCP JWT verification failed (...): APIError: no token payload`, then `MCP bearer token rejected`, so `/mcp` 401s.
- Reproduced by a real MCP client (dust.tt) against a self-hosted 0.3.11; confirmed from inside the pod: fetching the external `/api/auth/jwks` fails to resolve, while `http://127.0.0.1:<port>/api/auth/jwks` returns the exact `kid` in the token.

### Proposed fix
nao is the token issuer, so it already holds its signing keys — it shouldn't re-fetch them over its external URL. Verify against the in-process JWK set (`auth.api.getJwks()`) with `jose`'s `createLocalJWKSet` + `jwtVerify`, keeping the same `audience`/`issuer` checks. No network round-trip, works regardless of DNS topology.

PR incoming.

_Disclosure: investigated and drafted with Claude (Opus 4.8)._

afranzi:
Fixes #1620.

### Problem
`verifyOAuthAccessToken` verified MCP tokens via better-auth's `verifyAccessToken` with `jwksUrl = ${BETTER_AUTH_URL}/api/auth/jwks` — the server fetching its **own** JWKS over its **external** issuer URL. In self-hosted split-horizon deployments that host isn't resolvable from the server's own network, so the fetch throws, better-auth swallows it, and every MCP token is rejected with `no token payload` (even valid ones). Full repro in #1620.

### Fix
Verify against the in-process JWK set (`auth.api.getJwks()`) with `jose` (`createLocalJWKSet` + `jwtVerify`), keeping the same `audience`/`issuer` checks — no self-referential network fetch. Extracted the crypto path into `mcp/verify-jwt.ts` so it's unit-testable without a live auth instance.

### Testing
- New `tests/mcp-local-jwks-verify.test.ts` (5 cases): valid token, multi-audience match, wrong audience, wrong issuer, key-not-in-set. All pass.
- `tsc --noEmit`, eslint, prettier clean on the changed files.
- Verified against a live self-hosted deployment: after this change `/mcp` accepts the same token the external-fetch path was rejecting.

_Disclosure: implemented with Claude (Opus 4.8)._

<a href="https://cubic.dev/pr/getnao/nao/pull/1621?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/auth.ts
+++ b/apps/backend/src/auth.ts
@@ -8,7 +8,6 @@ import type { BetterAuthPlugin, Session } from 'better-auth';
 import { APIError, betterAuth } from 'better-auth';
 import { drizzleAdapter } from 'better-auth/adapters/drizzle';
 import { createAuthMiddleware } from 'better-auth/api';
-import { verifyAccessToken } from 'better-auth/oauth2';
 import { jwt } from 'better-auth/plugins';
 import { bearer } from 'better-auth/plugins/bearer';
 import type { JWTPayload } from 'jose';
@@ -16,6 +15,7 @@ import type { JWTPayload } from 'jose';
 import { db } from './db/db';
 import dbConfig, { Dialect } from './db/dbConfig';
 import { env, isCloud, MCP_VALID_AUDIENCES } from './env';
+import { verifyJwtWithLocalJwks } from './mcp/verify-jwt';
 import * as orgQueries from './queries/organization.queries';
 import * as projectQueries from './queries/project.queries';
 import * as userQueries from './queries/user.queries';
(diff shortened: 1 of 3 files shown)


Bl3f (approved):
LGTM

PR #1659, opened 2026-09-13
fix: bound execute_sql results and recover the sidecar

### LINKED ISSUE #1654: [bug] Unbounded fetch in execute_sql OOM-kills the worker, which never respawns - all SQL fails until manual container replacement
## Problem

`execute_sql` fetches result sets unbounded, so a single large `SELECT` can OOM-kill the multiprocessing worker - and the worker does not come back. Every later `execute_sql` fails until someone manually replaces the container.

Reported by a customer running self-hosted nao against Snowflake, who has hit it more than once and traced it themselves.

**Three problems stacked on top of each other:**

1. **The fetch is unbounded.** A user asks for a big select, the whole result set is pulled at once, and worker memory spikes fast enough to get killed.

2. **The result is held twice.** The conversion materializes the full list of rows and the DataFrame at the same time:

   ```python
   return pd.DataFrame([tuple(row) for row in cursor.fetchall()], columns=columns)
   ```

   `cursor.fetchall()` builds the complete list of tuples before `pd.DataFrame` copies it, so peak memory is roughly double the result set at exactly the moment it is already largest.

3. **There is no recovery.** This is the part that turns a bad query into an outage: the killed worker does not respawn, so SQL execution stays broken for everyone on that deployment until a human intervenes. One oversized query taken by one user takes down the tool for the whole instance.

**Their current workaround does not hold.** They added a rule in `RULES.md` capping results at 10,000 rows. It is arbitrary, it depends on the agent honoring it, and row count is the wrong unit anyway - a very wide result set blows the same budget in far fewer rows.

## Expected behavior

- **Batched/streamed fetching** for large result sets, so memory does not scale with the size of the result.
- **Avoid holding two full copies** during the DataFrame conversion - build it from an iterator or in chunks rather than from a fully materialized list.
- **The worker recovers on its own** after a crash. A killed worker respawns and `execute_sql` works again without manual container replacement. This matters independently of the memory fixes: whatever else exhausts memory later should degrade to one failed query, not a dead tool.
- **A bounded result contract enforced in code, not in prompt rules.** A server-side cap - by bytes or rows, configurable - so the guardrail does not depend on the agent following `RULES.md`. When the cap is hit, the user gets a clear "result too large, narrow your query" message rather than a crash.

## Priority note

Filed as P0 because of point 3 rather than the memory use itself: the failure is not a slow query or a failed request, it is complete and persistent loss of SQL execution for every user of the deployment, recoverable only by manual ops.

## Customer ID

50bf7f81-d5ba-493e-8463-551b848c76cc

mansiverma897993:
## Summary
- Fetch SQL cursor results in batches instead of `fetchall()`, avoiding the duplicate full-result list during DataFrame conversion.
- Enforce configurable row and decoded-payload limits (`NAO_SQL_MAX_RESULT_ROWS`, `NAO_SQL_MAX_RESULT_BYTES`) and return HTTP 413 with an actionable message when exceeded.
- Apply the bounded path to BigQuery, Redshift token queries, and materialized ClickHouse result rows; retain a safe rejection for unsupported non-streaming driver results.
- Increase FastAPI supervisor restart retries and restore the worker directory after each request.

Fixes #1654.

## Testing
- `pytest apps/backend/fastapi/test_main.py -k "not bigquery" -q` — 12 passed, 2 deselected (live BigQuery SSO tests require authenticated credentials).
- `pytest cli/tests/nao_core/config/test_execute_sql_limits.py -q` — 2 passed.
- CLI `ruff check`, `ruff format --check`, and `ty check` on changed database code — passed.
- Root JavaScript lint could not start locally because the installed npm is 10.9.3 while this checkout pins npm 11.10.1.

The commit is DCO signed and authored by Mansi Verma.

<a href="https://cubic.dev/pr/getnao/nao/pull/1659?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f:
I'm sorry, we will work on this internally and it touches core component of nao.

PR #1662, opened 2026-09-14
feat: enable iam authentication for Redshift

### LINKED ISSUE #1522: [feature] Add IAM authentication support for Redshift
### Problem
Redshift supports IAM-based temporary credentials via get_cluster_credentials. Right now nao only supports username/password, which forces teams to store long-lived credentials. Some AWS setups don't allow that and IAM roles or named profiles are the only option.

### Proposed solution
Introduce auth_mode 'iam', call boto3.redshift.get_cluster_credentials() to exchange IAM identity for short-lived db credentials, then pass those to the existing ibis connection. Three credential sources supported: named profile, explicit access keys, or default chain (env vars / instance role).

I've implemented this and tested it against a real Redshift cluster. Happy to submit a PR if this gets an approval!

synthpieter:
Closes #1522 

### Problem
Redshift supports IAM-based temporary credentials via get_cluster_credentials. Right now nao only supports username/password, which forces teams to store long-lived credentials. Some AWS setups don't allow that and IAM roles or named profiles are the only option.

### Solution
Adds IAM authentication for Redshift, alongside the existing username/password and Azure Entra ID flows. When auth_mode is set to iam, credentials are exchanged for short-lived DB creds via boto3's get_cluster_credentials() instead of a stored password. Supports three credential sources: a named AWS profile, explicit access keys, or the default AWS credential chain (env vars / instance role).

### Open questions
I noticed there are actually zero tests for any Redshift connectivity. Do we want to add just tests for iam authentication in this PR/branch, or does that need to happen in a separate issue/pr and add it for all the connection options?

This PR was written using Claude Sonnet 4.6 & 5 (claude-sonnet-4-6, claude-sonnet-5).

<a href="https://cubic.dev/pr/getnao/nao/pull/1662?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


cli/nao_core/config/databases/redshift.py
```diff
@@ -208,28 +209,56 @@ class RedshiftConfig(DatabaseConfig):
     user: str | None = Field(
         default=None,
         description=(
-            "Username. Required for password auth. In azure_entra_id mode it is optional "
-            "but recommended: nao sync uses it to read metadata (columns, previews, query history). "
-            "It is never used at runtime from /execute_sql."
+            "Username. Required for password auth. In iam mode it is the db user requested via "
+            "get_cluster_credentials, not the IAM-prefixed username used to connect. In azure_entra_id "
```
Bl3f: Don't know to update this description.

cli/nao_core/config/databases/redshift.py
```diff
@@ -208,28 +209,56 @@ class RedshiftConfig(DatabaseConfig):
...
         ),
     )
     password: str | None = Field(
         default=None,
         description=(
-            "Password. Required for password auth. In azure_entra_id mode it is optional "
-            "but recommended: nao sync uses it to read metadata (columns, previews, query history). "
-            "It is never used at runtime from /execute_sql."
+            "Password. Required for password auth; ignored in iam mode, where a temporary password is "
+            "fetched via get_cluster_credentials instead. In azure_entra_id mode it is optional but "
```
Bl3f: Same here.

cli/nao_core/config/databases/redshift.py
```diff
@@ -244,10 +273,47 @@ def promptConfig(cls) -> "RedshiftConfig":
...
+                    "Default credential chain (env vars / instance role)",
+                ],
+            )
+            if iam_source == "AWS Profile":
+                profile_name = ask_text("AWS Profile Name:", default="default")
+            elif iam_source == "Access Keys":
+                aws_access_key_id = ask_text("AWS Access Key ID:", required_field=True)
+                aws_secret_access_key = ask_text("AWS Secret Access Key:", password=True, required_field=True)
+                aws_session_token = ask_text("AWS Session Token (optional):", password=True) or None
+        else:
```
Bl3f: should not be the default else, the default should be password

cli/nao_core/config/databases/redshift.py
```diff
@@ -244,10 +273,47 @@ def promptConfig(cls) -> "RedshiftConfig":
...
+        password: str | None = None
+        auth_mode = RedshiftAuthMode.PASSWORD
+        cluster_id: str | None = None
+        region_name: str | None = None
+        aws_access_key_id: str | None = None
+        aws_secret_access_key: str | None = None
+        aws_session_token: str | None = None
+        profile_name: str | None = None
+
+        if auth_method == "Username / Password":
```
Bl3f: Can we rely on something else than a user facing string here?

Bl3f (changes requested):
A few small comments, a part from this LGTM

cli/nao_core/config/databases/redshift.py
synthpieter: Thanks for your feedback! This is one of my first open source contributions, so it's nice to learn.

Probably, yes, but I looked at other comparable setups in for example
https://github.com/getnao/nao/blob/36bb21122592fc289773ae66d2c810092bafa36e/cli/nao_core/config/databases/athena.py#L76-L91
https://github.com/getnao/nao/blob/36bb21122592fc289773ae66d2c810092bafa36e/cli/nao_core/config/databases/bigquery.py#L461-L479

I figured it was best to follow the same pattern?

cli/nao_core/config/databases/redshift.py
synthpieter: Do you mean to not update this description? Or do it better? :)

Bl3f: Sorry, my English was weird, I wrote it while speaking to someone else... I mean "do not update the description"

cli/nao_core/config/databases/redshift.py
Bl3f: Cant we use a id, display_string mapping for choices instead?

synthpieter: More like this? Peeked a bit at the init setup process

cli/nao_core/config/databases/redshift.py
synthpieter: Check, reverted.

cli/nao_core/config/databases/redshift.py
synthpieter: I think this is solved by going for the questionary approach.

cli/nao_core/config/databases/redshift.py
synthpieter: Reverted!

synthpieter:
@Bl3f Would love to know if there is more work needed! Please let me know.

PR #1667, opened 2026-09-14
fix - Chat cleanly displays provider 500 error #1633

### LINKED ISSUE #1633: [bug] Provider 500 renders as a raw Zod invalid_union dump in chat instead of an error message
## Problem

When the upstream LLM provider returns a 500, the chat shows `No response` for the user's message and then renders a multi-thousand-character Zod validation dump in red, directly in the conversation.

What actually happened: LiteLLM returned a provider error payload instead of a stream event.

```json
{"error":{"message":"litellm.APIError: The server had an error processing your request. Sorry about that! You can retry your request, or contact us through an Azure support request at: https://go.microsoft.com/fwlink/?linkid=2213926 if you keep seeing this error. (Please include the request ID <redacted> in your email.)","type":null,"param":null,"code":"500"}}
```

That error object is then fed to the stream-event parser, which tries it against every variant of the OpenAI Responses API event union and fails on all of them. The user sees the resulting `invalid_union` report - every rejected variant with its own per-field `invalid_type` errors, covering `response.output_text.delta`, `response.completed`, `response.function_call_arguments.delta`, `response.reasoning_summary_text.delta`, `response.apply_patch_call_operation_diff.done` and roughly a dozen more - as the answer to their question. Full text truncated here; it runs to several screens.

Two distinct defects:

1. **A provider error payload is treated as a stream part.** A non-event object arriving on the stream should be detected and unwrapped as an error before it reaches schema validation. The `code: "500"` and `message` are right there in the payload.
2. **Validation failures render raw to the end user.** Even when parsing genuinely fails, the chat should not print the validator's internal output. This is the same failure mode as #1037, which was fixed for a different stream part type (`data-compactionSummaryStarted`) - the underlying "unparseable part becomes UI text" path is still open.

Related: #1491 (transport failure surfaced as a misleading UI state rather than a real error message).

## Expected behavior

- A provider error on the stream is recognized as an error and surfaced as a short human-readable message, e.g. "The model provider returned an error (500). Retry." with the provider request ID available for support but not dumped inline.
- Transient upstream 5xx errors are retried before the turn is failed.
- No schema-validation output ever reaches the chat transcript; validation failures are logged server-side and shown to the user as a generic failed-response state with a retry action.
- The user's message does not end up in a `No response` state with no way forward - a failed turn is retryable from the UI.

lshiina-figon:
When the LLM provider returns 500, the chat displays a clean error message followed by a copy-able request ID and a retry button.
Raw Zod dump is no longer displayed to the user.
Capture d’écran 2026-09-14 à 15 09 56

Fixes #1633 

<a href="https://cubic.dev/pr/getnao/nao/pull/1667?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/apps/backend/src/services/agent.ts
+++ b/apps/backend/src/services/agent.ts
@@ -481,6 +481,15 @@ class AgentManager {
 	): ReadableStream<InferUIMessageChunk<UIMessage>> {
 		let error: unknown = undefined;
 		let result: StreamTextResult<AgentTools, never> | undefined;
+		const handleError = (err: unknown): string => {
+			error = err;
+			logger.error(`Agent stream error: ${String(err)}`, {
+				source: 'agent',
+				projectId: this.chat.projectId,
+				context: { chatId: this.chat.id, modelId: this._modelSelection.modelId },
+			});
+			return formatErrorMessageForUI(err);
+		};
 
 		return createUIMessageStream<UIMessage>({
 			generateId: () => crypto.randomUUID(),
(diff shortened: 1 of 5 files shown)


Bl3f:
Would be awesome if we can add to this a button "Send to nao" ; when clicking on "Send to nao" it would package the conversation (without output of the SQL or any PII) and send to us the error so we can debug and fix later on. 

If you can work on this would be awesome, you can send the errors to our licenses server i think.

Bl3f (approved):
LGTM, let's do what i said in another PR.

PR #1671, opened 2026-09-15
fix(charts): bundle a font so chart PNGs render text outside Docker

### LINKED ISSUE #1669: Chart PNGs render with no text on hosts without usable fonts
## Description

On some deployments, every chart exported as a PNG comes out with no text at all: no title, no axis tick labels, no axis titles, no data labels, no legend text. The vector shapes render correctly (bars, donut arcs, gridlines, legend colour swatches), so the image looks half-drawn.

This affects every PNG surface: downloading a chart from the nao UI, charts posted to Slack/Teams, and story exports. The interactive chart in the browser is unaffected.

**Important for reproduction:** this does not reproduce on a typical dev machine. Whether text renders depends on the fonts available on the host doing the rasterisation — see "Root cause" below. A maintainer testing locally will most likely see correct text.

## Steps to reproduce

1. Deploy on a host with no fonts. On the affected host, `fontconfig` is not even installed:
   ```
   $ fc-list | wc -l
   Command 'fc-list' not found, but can be installed with: sudo apt install fontconfig
   0
   ```
2. Ask the agent a question that produces a `display_chart` result.
3. Download the chart as a PNG, or have it posted to Slack.

**Expected:** the PNG matches what the UI shows, text included.
**Actual:** shapes only, no glyphs anywhere.

## Evidence

Downloading the same chart from the nao UI and from Slack produces byte-identical files (same MD5, same size), which rules out Slack's upload path: the PNG is already textless when generated.

Decoding an affected PNG (1600x1048, i.e. the default 800x500 plus a 24px legend band, at `zoom = 2`) shows 0 non-white pixels in the title band, and the ink in the bottom band accounted for by axis tick marks rather than glyphs.

Measured on a legend fragment taken from `buildLegend`, rasterised through resvg:

| Host fonts | Ink pixels | What renders |
|---|---|---|
| system fonts available | 1195 | swatch + text |
| no usable fonts (`loadSystemFonts: false`) | 100 | swatch only, no glyphs |

100 pixels is the colour swatch alone. That is the broken case.

## Root cause

Charts are rasterised by `svgToPng` in `apps/backend/src/utils/generate-chart.ts` using `@resvg/resvg-js`. resvg is a Rust renderer (usvg/fontdb), not a browser, and needs a **concrete** font family to shape glyphs.

The SVG asks for `system-ui`, which is not a font name — it is a CSS keyword that only a browser's style engine resolves to the platform UI font. resvg looks it up in fontconfig, never finds it, and falls back to whatever else the host offers:

- On a host with fonts, the fallback finds something and text renders. The bug is invisible.
- On a host with no usable fonts, there is nothing to fall back to. resvg draws the shapes (which need no font) and silently skips every glyph.

Three things make this fragile:

1. **`system-ui` gives resvg no concrete target.** Used in `utils/generate-chart.ts` (legend text), `shared/chart-builder.tsx` (title) and `shared/chart-data-labels.tsx` (data labels).
2. **Axis ticks declare no family at all.** `AXIS_TICK` in `shared/chart-builder.tsx` only sets `fontSize`, and Recharts adds no default.
3. **`svgToPng` gives resvg no fallback.** It passes only `font: { loadSystemFonts: true }` — no `fontFiles`, no `defaultFontFamily` — so the outcome is entirely at the mercy of the host.

The result is non-deterministic rendering: the same chart produces different output, or no text at all, depending on the machine.

### Not the cause

Two plausible-looking suspects were measured and ruled out:

- **CSS custom properties.** `fill: 'var(--foreground, #111827)'` (`chart-data-labels.tsx`, chart title, and `AXIS_LABEL_STYLE` via `MUTED_FOREGROUND`) is not resolved by resvg, which falls back to the default black fill. Ink measurements are identical with and without `var()`, so this does not remove text — it only makes colours wrong. Worth fixing, but a separate defect.
- **`system-ui` alone.** On a host with fonts, `system-ui` renders text fine (1195 ink pixels). The keyword is only fatal in combination with a host that has nothing to fall back to.

## Impact

On affected deployments, every chart shared outside the nao UI is unreadable: bars with no categories, no scale, no legend. Charts posted to Slack are the main way non-technical users consume these answers.

## Environment

- The `Dockerfile` runtime stage installs `fontconfig` and `fonts-dejavu-core`, so the standard container image should not be affected. Deployments that run the compiled binary outside that image, or where fontconfig sees no fonts, are.
- Because the failure depends on the host rather than on the code path, it can appear in production while every dev machine looks fine.

briacbln:
Fixes #1669

## Problem

Outside the official Docker image, chart PNGs (downloads, Slack/Teams posts, story exports) render with **no text at all** — no title, no axis labels, no data labels, no legend text — while the shapes draw correctly.

`1da0eecd` ("Hotfix: add fonts to docker image", March 2026) added `fontconfig` and `fonts-dejavu-core` to the runtime stage, which fixes the official image. It does nothing for the pip wheel, which is the documented install path for self-hosting, and that is what this PR addresses.

Reproduced on a real deployment: a `python:3.12-slim` image that installs `nao-core[postgres,anthropic,notion]==0.3.7` from pip and nothing else.

```
$ docker exec <container> fc-list | grep -i dejavu
$ docker exec <container> ls /usr/share/fonts/truetype/dejavu
ls: cannot access '...': No such file or directory
```

Three independent checks, each run rather than assumed:

- `python:3.12-slim` ships **no** `/usr/share/fonts` at all and no `fontconfig` binary.
- The published wheel `nao_core-0.3.7-py3-none-manylinux2014_x86_64.whl` contains **no** sans-serif font. Its only `.ttf` files are KaTeX faces under `nao_core/bin/public/assets/`, served to the browser for formula rendering. No DejaVu, Noto, Liberation or Roboto.
- The wheel lays the binary out as `nao_core/bin/nao-chat-server` with **no** `assets/` directory beside it.

So resvg has nothing to load, on either the host or the package.

## Cause

`svgToPng` (`apps/backend/src/utils/generate-chart.ts`) rasterises with `@resvg/resvg-js`. resvg is a Rust renderer, not a browser, and needs a **concrete** font family to shape glyphs.

The SVG asks for `system-ui`, which is not a font name but a CSS keyword only a browser resolves. resvg looks it up, finds nothing, and falls back to whatever the host offers. With fonts installed the fallback succeeds, which is why this never reproduces on a dev machine. With none there is nothing to fall back to: resvg draws the shapes and skips every glyph.

Compounding it, axis ticks (`AXIS_TICK`) declared no family at all, and `svgToPng` passed resvg no `fontFiles` and no `defaultFontFamily`, leaving the outcome entirely up to the host.

Ink pixels on a legend label, rasterised through resvg with `loadSystemFonts: false` to reproduce the affected deployment:

| | before | after |
|---|---|---|
| legend label | **0** | **1781** |
| title band | 1412 | 6053 |

Zero is the bug: the label is not faint, it is absent. The title band keeps some ink before the fix because the chart frame draws there too.

## Fix

Ship a concrete font and name it explicitly, so rendering no longer depends on the host:

- Bundle DejaVu Sans (Regular + Bold) in `apps/backend/assets/fonts/`, with its licence. Bold is needed for the `600`-weight chart title. 1.4 MB total.
- Add `CHART_FONT_STACK` (`'DejaVu Sans', system-ui, sans-serif`) and `CHART_EMBEDDED_FONT_FAMILY` in `apps/shared/src/chart-fonts.ts`, replacing four duplicated literals. The stack suits both paths: browsers lacking DejaVu fall through to `system-ui`, while resvg gets a concrete family it has been handed the file for.
- Pass `fontFiles`, `defaultFontFamily` and `sansSerifFamily` to resvg, keeping `loadSystemFonts: true` for extra glyph coverage. `defaultFontFamily` is the safety net: even a `<text>` with no declared family now renders.
- Give `AXIS_TICK` and `AXIS_LABEL_STYLE` a family, covering all 13 axis call sites at once.
- Resolve `var(--x, fallback)` before rasterising. resvg ignores custom properties and falls back to black, so axis labels and data labels lost their intended grey. This is a colour fix, not the text fix — measurement confirmed `var()` never removed text. It runs inside `svgToPng` only: `renderChartToSvg` also feeds story HTML, where the variables must survive for theming.
- Apply the same family in `generate-map.tsx`: `generateMapImage` falls back to `svgToPng` when Chromium is unavailable, hitting the identical bug on that path.
- Copy `apps/backend/assets/` in `cli/build.py` so the wheel carries the fonts next to the binary.

Font resolution (`apps/backend/src/utils/chart-fonts.ts`) follows the existing `email-logo.ts` pattern, and its first candidate is `dirname(process.execPath)/assets/fonts` — the layout `cli/build.py` produces. That is the same mechanism `app.ts:316` already uses to find `public/` in the compiled binary, and `cli.ts:32-43` documents why it is correct: under `bun build --compile`, `Bun.main` points into the embedded `/$bunfs/` filesystem while `process.execPath` points at the real binary on disk. Resolution degrades to `[]` rather than throwing, so a missing directory falls back to system fonts instead of failing the render.

No Dockerfile change: `1da0eecd` already covers that path, and this fix no longer depends on it.

## Why bundle a font instead of installing one

Installing a font package only ever fixes the environment you control. `1da0eecd` did that for the official image and the bug survived in every other install — this PR exists because of that gap. A wheel user cannot be asked to `apt-get install` anything.

It also leaves rendering non-deterministic: the same chart gets different text metrics per host. Label widths are hard-coded (`charWidth = 7` in `generate-chart.ts`, `DATA_LABEL_CHAR_WIDTH_RATIO = 0.6` in `chart-data-labels.tsx`) and only make sense against a known font.

## Testing

Run on Node 22.14.0 with npm 11.10.1, as `engines` requires.

- `npm run lint` (`tsc --noEmit && eslint` across backend, frontend, shared): passes.
- `prettier --check .`: passes.
- `apps/backend/tests/generate-chart-fonts.test.ts`: 6/6.
- Full backend suite: 62 failures, of which **61 already fail on the parent commit** `daa698d9`. Compared per-test via vitest JSON reports; the delta is one pre-existing flake (`sandbox-runtime`, gated on `/dev/kvm`, which passes 3/3 in isolation on both commits). No chart, font or map test fails. The unrelated failures come from an unmigrated local SQLite file and from `git for-each-ref --exclude=` requiring git ≥ 2.42.
- End-to-end PNG check: rasterised through the production `svgToPng` path and counted ink pixels, decoding the PNG with `node:zlib` alone so no dependency was added. Numbers in the table above.

The unit test asserts on the generated SVG without rasterising, so it stays deterministic and independent of the native binding and of CI fonts:

- every `<text>` node declares a resolvable family, reading `font-family` **or** `style` — Recharts sets the attribute on ticks but folds an axis label's props into `style`, and resvg honours both;
- the markup never falls back to bare `system-ui, sans-serif` (regression guard);
- title, axis label and category labels are present, compared against decoded text since `renderToString` escapes `€` to `&#x20ac;`;
- `resolveCssVariables` leaves no `var(--` behind and does not disturb plain colours.

## Not verified

- A wheel built from this branch, installed clean, generating a PNG. `cli/build.py` needs `bun`, unavailable on the machine used. The resolution mechanism is argued from `cli.ts`'s own documented contract and from the layout of the published 0.3.7 wheel, but the built artefact was not exercised. **This is the path the PR is for, so it deserves a reviewer's check.**
- Pie/donut specifically: the only charts using `buildVerticalLegend` + `truncateLabel`, whose `charWidth = 7` now matches a known font. Worth confirming truncation is neither too aggressive nor too loose.
- No CI job runs vitest (`js-lint.yml` runs `npm ci && npm run lint` only), so the new test will not execute in CI as things stand.

## Not included

Two unrelated defects found while investigating, left for separate PRs:

- `_uploadChartImageFile` (`services/slack.ts`) passes no `title`/`initial_comment` to `files.uploadV2`, so Slack shows only the filename under an image.
- `mdToMrkdwn` (`utils/messaging-provider.ts`) only strips an orphan `**` when it is not followed by a non-blank character, so `**73` survives into Slack as broken bold.

<a href="https://cubic.dev/pr/getnao/nao/pull/1671?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Bl3f:
Hey @briacbln thanks for this. Before moving forward, I'm curious to know why would you want to not use the official Docker image or even inherit from it?

briacbln:
@Bl3f 
Fair point, and you're right that my setup isn't the argument. I could inherit from the official image and my problem goes away, so let me drop that part.

I install nao as a library rather than as the app: pip install nao-core, which is what the README's own 1-minute Quickstart recommends. But that's a preference, and preferences don't justify a patch.

The reason I think this is still worth fixing is that the pip wheel is a published artifact that renders chart PNGs with no text on any host without fonts. 1da0eecd fixed the image back in March but nothing fixed the wheel.

There's also a reason unrelated to packaging: label widths are hard-coded (charWidth = 7 in generate-chart.ts, DATA_LABEL_CHAR_WIDTH_RATIO = 0.6 in chart-data-labels.tsx). Those constants only mean something against a known font, so today truncation and label collision depend on whichever font the host happens to resolve, inside the official image too.

If the 1.4 MB of fonts is the sticking point, the smaller version is to set only defaultFontFamily/sansSerifFamily to a concrete family without bundling files: that stops the silent blank-text failure wherever a font exists, but leaves a bare host broken. Happy to cut it down to that, or to close this if you'd rather treat the wheel as a Docker-first install.

One correction to my own description: it originally framed this as "chart PNGs have no text", which stopped being true in March. The scope is the wheel.

PR #1692, opened 2026-09-17
fix(mcp): expose valid database IDs to execute_sql

### LINKED ISSUE #1652: [bug] execute_sql over MCP: agent picks the wrong database_id when a project has several databases
## Problem

When a project has more than one database configured, the client agent calling `execute_sql` over MCP does not reliably know which `database_id` to use, and the call errors.

Reported by a customer whose project exposes two databases, a raw layer and a mart layer. The agent got confused about which one to target and returned an error rather than picking one or asking.

Two things narrow this down:

- **It is model-dependent.** It happened once, and stronger models get it right consistently - the reporter noted that with a more capable model he never saw it, "or maybe it did and it just figured out what the right call was." A failure that disappears with model strength is a prompt/description gap, not broken logic.
- **The equivalent problem was already fixed in the UI.** The in-app path knows to pin the database; the tool definition handed to external MCP clients does not carry the same instruction, so the model is left to infer it.

That points at how the tool is described to the client rather than at query execution, which should make it a small fix.

## Expected behavior

- The `execute_sql` tool description states explicitly that `database_id` must be specified, and that a project may expose several databases.
- The tool surfaces the available database ids to the calling agent, so choosing one does not depend on the model having inferred them from earlier context.
- When `database_id` is missing or unknown, the error names the valid ids rather than leaving the agent to guess on retry.
- Where a project has exactly one database, it is used without requiring the agent to name it, so the stricter contract does not add friction to the single-database case.

## Related

- #1428 - `Invalid input for tool execute_sql`, a malformed `database_id` payload. Different failure, same argument.

## Customer ID

ad4mou:
## Summary

- Gives MCP clients the valid database IDs for `execute_sql`.
- Requires `database_id` when multiple databases are configured.
- Preserves automatic selection when only one database exists.

Closes #1652

<a href="https://cubic.dev/pr/getnao/nao/pull/1692?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


apps/backend/src/mcp/tools/context-layer.ts
```diff
@@ -145,6 +151,60 @@ function registerExecuteSql(server: McpServer, ctx: McpContext): void {
...
+	return [
+		...new Set(
+			configuredDatabases.map((database) => database.id).filter((databaseId) => databaseId !== LOCAL_DATABASE_ID),
+		),
+	];
+}
+
+function buildExecuteSqlDescription(warehouseDatabaseIds: string[]): string {
+	if (warehouseDatabaseIds.length < 2) {
+		return EXECUTE_SQL_BASE_DESCRIPTION;
```
ad4mou, line 164: just made sure that if there only one DB we don't add the id

ad4mou:
LGTM

PR #1698, opened 2026-09-18
fix: fallback to make sure saved files show as file chips

### LINKED ISSUE #1650: [bug] Generated file is unreachable: saved-file tag renders as raw text instead of a preview/download control
## Problem

When the agent writes an analysis to a file, the file is unreachable from the chat. The `saved-file` tag is rendered as literal text instead of being interpreted into a preview/download control.

The user asked where the file with their analysis was, and the response rendered as:

```
Le fichier est ici : <saved-file path="/home/<project>/<name>.md"><name>.md</saved-file>

Tu peux cliquer dessus pour le prévisualiser ou le télécharger. [...]
```

So the message tells the user to click the file to preview or download it, and there is nothing clickable - just the raw markup. The analysis the agent produced cannot be opened, downloaded, or reached any other way from the UI.

This is a frontend tag-interpretation failure, not an agent-side one: the agent emitted the tag it is supposed to emit, and the frontend should have turned it into a button.

The prose makes it worse than a missing control. Because the agent describes an affordance in its own text, a user who sees no button assumes their session is broken rather than that the feature failed - the reporter's read was "there's nothing to click on."

## Expected behavior

- The frontend parses the `saved-file` tag and renders a file control offering preview and download.
- A tag that cannot be parsed, or one pointing at a file that no longer exists, never renders as raw markup. It degrades to something intelligible and says the file is unavailable.
- The clickable control, not the prose, is what tells the user the file is there - so a rendering failure cannot leave the message promising something the UI does not show.

## Customer ID

14349e6f-129f-4c25-ada3-1c93cdb32b82

ad4mou:
## Summary

- Render backticked saved-file tags as preview and download controls.
- Hide partial tags while streaming to prevent visual flicker.
- Keep normal and fenced code unchanged.

Closes #1650

<a href="https://cubic.dev/pr/getnao/nao/pull/1698?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ad4mou:
LGTM

Bl3f:
LGTM

PR #1703, opened 2026-09-18
feat: let users select their organization

### LINKED ISSUE #1693: let users select their org
The way we decide the org a user should be on is problematic, we just take the first org he was added to.
So for instance if a user signs up using an SSO, which means he wasn't first added to the org by invitation, he gets into a solo org instead of the org linked to the SSO.

Only when the user gets invited first does that cause no issues.

even just if a user is invited into any other org, he can never really access it since he already has one.

Especially problematic in nao cloud

So we should, in organisation settings, have a small little organisation selector.

ad4mou:
## Summary
- Add an organization dropdown for users who belong to multiple organizations.
- Remember the selection and use it across organization settings and actions.
- Keep the existing default when no valid selection exists.

Closes #1693

<a href="https://cubic.dev/pr/getnao/nao/pull/1703?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ad4mou:
LGTM

Bl3f:
LGTM

